PR #8, opened 2017-07-27
Create first acceptance test

cherewaty:
Sets up:
- mirage server in development and test environments
- mirage factory for vendors
- a basic search UI
- acceptance test with simple page object

hltwxkg4yo


mirage/start.js
```diff
@@ -0,0 +1,38 @@
...
+req.keys().forEach((modulePath) => {
+  const moduleParts = modulePath.split('/');
+  const moduleType = moduleParts[1];
+  const moduleName = moduleParts[2];
+
+  if (moduleName) {
+    const moduleKey = camelCase(moduleName.replace('.js', ''));
+    modules[moduleType][moduleKey] = req(modulePath).default;
+  }
+});
```
cowboyd: What exactly we're doing here needs some documentation.

mirage/start.js
```diff
@@ -0,0 +1,38 @@
...
+
+export function startDevMirage() {
+  let options = Object.assign(modules, { environment, baseConfig });
+  return new Mirage(options);
+}
+
+export function startTestMirage() {
+  let testModules = { ...modules };
+  delete modules.scenarios;
```
cowboyd: Why do we need to mutate the list of modules here?

src/index.js
```diff
@@ -17,4 +18,8 @@ if (module.hot) {
   module.hot.accept('./components/app', render);
 }
 
+if (process.env.NODE_ENV === 'development') {
```
cowboyd: Is this code run in the browser? Not sure what the mechanics are here, although definitely requires a little documentation.

tests/app-test.js
```diff
@@ -1,10 +1,77 @@
-/* global describe, beforeEach, afterEach, it */
+/* global describe, beforeEach, afterEach */
 import { expect } from 'chai';
+import $ from 'jquery';
+import sinon from 'sinon';
+import it from './it-will';
```
sadatay: maybe we should be explicit here about this being a convergent `it` and name it something else.  it might not be immediately clear to someone from the outside, since calling this `it` makes this look like a run-of-the-mill mocha test suite

.babelrc
```diff
@@ -1,4 +1,4 @@
 {
-  "presets": [["env", { "modules": false }], "react"],
+  "presets": [["env", { "modules": false }], "stage-3", "react"],
```
wwilsman: [stage-3](https://babeljs.io/docs/plugins/preset-stage-3/) includes rest, spread, and async generators. 

I was thinking maybe using [stage-2](https://babeljs.io/docs/plugins/preset-stage-2/) for class property transforms.

Looks like Folio's current modules use stage-0

mirage/start.js
wwilsman: In testing, we don't want to auto-load our scenarios. But since the require.context stuff has to happen at root, we just remove scenarios from the object passed in as mirage options.

We could also probably just move this down to `delete options.scenarios` beneath the `Object.assign`

src/index.js
wwilsman: `process.env` will be replaced with strings thanks to `webpack.EnvironmentPlugin`

tests/app-test.js
wwilsman: I was actually thinking maybe we want to move this to `it.will` so we could still use `it` if we absolutely need to

.babelrc
cowboyd: Let's roll stage-3. I don't think the object spread syntax is going to change, and dang it's handy.

tests/app-test.js
wwilsman: We also need a `.only` variant as that currently doesn't work with our convergent-it

cowboyd: What about `it.immediately` for the original behavior. The reason I say is that we want "not thinking about asynchrony" to be the default.

cowboyd: or `it.synchronously`

sadatay: i like the`it.immediately` route, but i think we should make a note about it in the readme.  people see `it` and think `it` - even though we're accepting this as our default it's not most people's default

mirage/start.js
sadatay: it'd probably be slightly better (albeit a bit uglier) to push the module list in from the outside.  what if we pulled all the code above these methods into a generic `start` that takes a module list and then `startTest...` and `startDev...` pass in what they want.  if you wanted to DRY it you could change `moduleTypes` to be those modules common to both and then pass the default concatenated with whatever extra moduleTypes you want for the given use case

wwilsman: Even better, we have `process.env.NODE_ENV` set to `"test"` when run in testing mode and `"development"` otherwise.

We could probably just add an if statement that excludes scenarios from the `moduleTypes` array.

cowboyd: If we have object spread, we can do without mutation

```js
{scenarios, ...options} = {options};
```

tests/app-test.js
cowboyd: @wwilsman agree, about `.only`, but let's bring that in with a separate PR

PR #11, opened 2017-08-03
Enable mirage in production.

cowboyd:
In order to demo and also to build out funcitonality with the server being incomplete and not fully defined, it's best to have mirage going everywhere.

### DIFF
--- a/mirage/index.js
+++ b/mirage/index.js
@@ -2,8 +2,16 @@
 
 let start = () => {};
 
-if (process.env.NODE_ENV !== 'production') {
-  start = require('./start').default;
-}
+// Currently we run mirage in production because the server side is
+// incomplete.
+//
+// However, when we switch over to the actual API, then we'll want to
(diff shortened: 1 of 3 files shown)


mirage/index.js
```diff
@@ -2,8 +2,14 @@
 
 let start = () => {};
 
-if (process.env.NODE_ENV !== 'production') {
-  start = require('./start').default;
-}
+// currently we run mirage in production
+// when we switch over to the initial api protect this
+// require with an if statement so that webpack won't
+// attempt to bundle mirage.
```
sadatay: i'd add this punctuation:
// currently we run mirage in production`.`
// when we switch over to the initial api`,` protect this
// require with an if statement so that webpack won't
// attempt to bundle mirage.

cherewaty: and this sentence goes on and on and on and ooooooon; don't stop; punctuatin'; dun dun dun; hold on to this feeling

PR #12, opened 2017-08-06
Stripify application

cowboyd:
The app before was a very simple react application. It lacked the container specific chrome that comes with a stripe application.

This adds that layer. The primary challenge was setting it up for development and testing since there needs to be a webpack configuration that can be consumed by Karma, and also by the default platform.

In order to run our platform with mirage, we had to add an entry to the production and dev webpack config to boot mirage.

In testing, we created a `TestHarness` component which re-supplies the boot logic normally contained in `@folio/stripes-core/src/index`


cowboyd:
@wwilsman I had to fix a bug in our custom `stripes-core` where  a new browser history object was getting created on each render, instead of once. https://github.com/cowboyd/stripes-core/commit/24d06dc27cda7e99ec1c13459387dcc4b19c1734 so it requires a review again.

PR #15, opened 2017-08-08
Get rid of separate `platform` package

cowboyd:
The only thing preventing us from running our tests and a demo application was that you need to be able to require the package under development `@folio/eholdings` from within the test harnes and the demo code. We were accomplishing this by having a separate `platform/` directory with its own `package.json` and `node_modules/` so that it could load the eholdings module.

The developer ergonomics of this are terrible though because you need to re-yarn install any time you want to make a set of changes to the eholdings app.

This sidesteps the issue entirely by adding a `postinstall` script (`bin/link-self`) which symlinks the eholding package root directory into its own `node_modules` directory. Now any node process that runs from the root of the eholdings package directory will be able to load the eholdings module itself.

In other words, your checkout will look like:

```
/path/to/ui-eholdings/
  - node_modules/
  - @folio/
    - stripes-connect/
    - stripes-core/
    - ....
    - eholdings -> /path/to/ui-eholdings
```

By only having a single project, a single package.json, and a single root `node_modules/` directory, it drastically simplifies the development workflow and reduces the need for re-installs.

### Implementation notes

- `yarn`, `yarn start`, `yarn test`, `yarn build` all from the main directory.
- the `demo/` directory now only holds the stripes configuration and scripts to build the demo platforrm


package.json
```diff
@@ -28,8 +29,8 @@
     "karma-mocha": "^1.3.0",
     "karma-mocha-reporter": "^2.2.3",
     "karma-webpack": "^2.0.3",
-    "lodash": "^4.17.4",
     "mirage-server": "cowboyd/mirage-server",
+    "lodash": "^4.17.4",
```
cherewaty: How did `lodash` get moved out of alpha order? 🤔

cowboyd: I think that was me. Worth putting back?

cherewaty: Nah, it'll get fixed at some point, not worth reopening this.

PR #17, opened 2017-08-09
Create vendor details page

cherewaty:
screen shot 2017-08-09 at 10 39 31 am

### DIFF
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -46,7 +46,7 @@ export default class App extends Component {
             No results found for <strong>{`"${searchQuery}"`}</strong>.
           </p>
         ) : (
-          <ul data-test-search-results-list>  
+          <ul data-test-search-results-list>
             {hasSearchResults && searchResults.vendors.map((vendor) => (
               <li data-test-search-results-item key={vendor.vendorId}>
                 {vendor.vendorName}
--- a/src/components/app.js
+++ b/src/components/app.js
@@ -36,7 +36,7 @@ export default class App extends Component {
             onChange={this.handleChange} />
           <button data-test-search-submit type="submit" disabled={!search}>Search</button>
         </form>
-        {!!errors && errors.map((err, i) => (
+        {!!errors && Array(errors).map((err, i) => (
           <p key={i} data-test-search-error-message>
             {err.message}. {err.code}
           </p>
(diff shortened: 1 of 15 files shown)


src/components/app.js
```diff
@@ -46,7 +46,7 @@ export default class App extends Component {
             No results found for <strong>{`"${searchQuery}"`}</strong>.
           </p>
         ) : (
-          <ul data-test-search-results-list>
+          <ul data-test-search-results-list>
```
sadatay: certainly not a blocker, but we've got two trailing spaces here.  we saw a similarly minor typo nail us yesterday in that one test.  gonna look into whether or not ESLint can be configured to scrub this in JSX

src/components/vendor-show.js
```diff
@@ -0,0 +1,35 @@
...
+          </p>
+        ))
+      ) : (
+        <p>Loading...</p>
+      )}
+    </div>
+  );
+}
+
+VendorShow.displayName = 'VendorShow';
```
cowboyd: This isn't necessary since we're using `function` syntax.

src/routes/vendor/vendor-show.js
```diff
@@ -0,0 +1,87 @@
...
+import fetch from 'isomorphic-fetch';
+
+import View from '../../components/vendor-show';
+
+class Vendor {
+  constructor(prev = {}, props = {}) {
+    Object.assign(this, prev, props);
+  }
+  get isRequested() { return false; }
+  get isLoading() { return false; }
```
sadatay: i see what state `isLoading` is supposed to represent, but it doesn't appear that we're actually using it anywhere.  we might want to review the logic around these states

src/routes/vendor/vendor-show.js
```diff
@@ -0,0 +1,87 @@
...
+  get isLoaded() { return false; }
+  get isErrored() { return false; }
+  get isPending() { return !this.isRequested || this.isLoading; }
+
+}
+
+class PendingVendor extends Vendor {
+  get isRequested()  { return true; }
+  get isPending() { return true; }
```
sadatay: since `isPending` is computed, and essentially means "hasnot been requested, or has been requested but is still loading", setting `isLoading` here instead of strong-arming `isPending` would probably reflect the state better

src/routes/vendor/vendor-show.js
```diff
@@ -0,0 +1,87 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import fetch from 'isomorphic-fetch';
+
+import View from '../../components/vendor-show';
+
+class Vendor {
```
sadatay: if we're going to keep these classes inline can we move them to the bottom of the file with a brief comment about what they are and how they're used?

cherewaty:
👍

PR #18, opened 2017-08-09
Display associated packages for a vendor

cherewaty:
localhost-3000-eholdings-vendors-1 iphone 5


src/components/vendor-show/vendor-show.js
```diff
@@ -0,0 +1,86 @@
...
+
+          {vendor.isLoaded && vendorPackages.isLoaded ? (
+            <div>
+              <h3>Packages</h3>
+              <List
+                itemFormatter={renderPackageListItem}
+                items={vendorPackages.packageList}
+                listClass={styles.list}
+              />
+            </div>
```
wwilsman: We should turn this into it's own component so we can reuse it on other pages. 
Something like `<PackagesList packages={...}/>`

tests/vendor-details-test.js
```diff
@@ -31,27 +30,27 @@ describeApplication('VendorDetails', function() {
     });
 
     it('displays the total number of packages', function() {
-      expect(VendorDetailsPage.numPackages).to.equal(vendor.packagesTotal);
+      expect(VendorDetailsPage.numPackages).to.equal(`Total Packages${vendor.packagesTotal}`);
```
sadatay: we copied this style in the package details pr, and i think it's ok to punt on for now, but we should probably figure out a better way to match these eventually.

PR #19, opened 2017-08-10
Package Details Page

sadatay:
**Package Details (Show - Metadata)**
This is the 'show' page for a package, or more specifically a 'vendor package', since there is no API for package as-of-yet.

Currently this only displays a subset of read-only metadata for the package, as well as a link back to the parent vendor.  It makes use of `stripes-components` for some styling, but could probably use a better layout.

As for the list view that should accompany this page, I'm unsure if it should be titles or a list of other vendors providing the package.  I'm pretty sure the former, but if there is no pure "package details" page due to the RM API restriction, then I'm not sure where the latter would go.  We could put both here?

screen shot 2017-08-09 at 9 53 46 pm

### DIFF
--- a/mirage/factories/package.js
+++ b/mirage/factories/package.js
@@ -3,7 +3,13 @@ import { Factory, faker } from 'mirage-server';
 export default Factory.extend({
   packageName: () => faker.commerce.productName(),
   titleCount: 1,
-  isSelected: 0,
+  isSelected: true,
   selectedCount: 1,
-  contentType: () => Math.floor(Math.random() * 8), // valid values: 0-7
+  contentType: () => faker.random.arrayElement([
+    'Online Reference',
+    'Aggregated Full Text',
+    'Abstract and Index',
+    'E-Book',
+    'E-Journal'
+  ])
 });
--- a/src/components/package-show/package-show.js
+++ b/src/components/package-show/package-show.js
@@ -1,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function PackageShow({ vendorPackage }) {
+  return (
+    <div data-test-eholdings-package-details>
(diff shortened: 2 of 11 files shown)


src/components/package-show/package-show.js
```diff
@@ -0,0 +1,51 @@
...
+              </div>
+              <div data-test-eholdings-package-details-titles-total>
+                <KeyValue label="Total Titles" value={vendorPackage.titleCount} />
+              </div>
+              <div data-test-eholdings-package-details-titles-selected>
+                <KeyValue label="Selected Titles" value={vendorPackage.selectedCount} />
+              </div>
+            </div>
+          ) : vendorPackage.isErrored ? (
+            vendorPackage.mapErrors((error, key) => (
```
sadatay: i was thinking about this a bit today - if this block spits out more than one error aren't we going to get react carping about sibling nodes?  i tried wrapping this in a `<div>` earlier but it broke stuff, and i didn't get around to figuring out why.

wwilsman: That's what the `key` is for. React is fine with rendering arrays of nodes, but they have to have unique keys

sadatay: oh, cool!  👍

mirage/factories/package.js
```diff
@@ -3,7 +3,13 @@ import { Factory, faker } from 'mirage-server';
 export default Factory.extend({
   packageName: () => faker.commerce.productName(),
   titleCount: 1,
-  isSelected: 0,
+  isSelected: true,
```
wwilsman: Is this a boolean from the real API or is it an integer?

sadatay: it's boolean in the response from RM-API

PR #24, opened 2017-08-15
Integrate with stripes-connect

wwilsman:
### What I've learned from this:

- The key in the manifest represents a piece of state in the redux store (that gets created when the manifest is processed).
- That piece of the state seems to be overwritten with subsequent calls.
- Because of this, connected component's manifests should have keys unique to their representation of the data.

  - The back button doesn't trigger another request, and so when these keys are not unique, data from `packages` within a vendor would actually be previous `packages` data from our origin page.

### Skipped tests

I've skipped the error tests for now because they will have to be reworked to align with stripes-connect.

-  A manifest can include an `@errorHandler` key whose value is a function called with a stripes error object. [[docs]](https://github.com/folio-org/stripes-connect/blob/master/doc/api.md#error-handling)

- Since the manifest is a static property, this handler won't easily be able to render errors inside of our components.

  - An idea for this is to create a "service" type class that when rendered will output specific errors from some static property. This static property can be populated via a static method that can be passed around to various `@errorHandler` manifests
  
  - This doesn't have to be a "service" type class. We could probably implement something similar by passing static properties from our routes as component properties to it's children. We would just have to trigger a rerender somehow when that static method is called

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,9 +10,20 @@ import CustomerResourceShow from './routes/customer-resource/customer-resource-s
 
 export default class EHoldings extends Component {
   static propTypes = {
+    stripes: PropTypes.shape({
+      connect: PropTypes.func.isRequired
+    }).isRequired,
     match: PropTypes.shape({
       path: PropTypes.string.isRequired
     }).isRequired
+  }
+
+  constructor(props) {
+    super(props);
+    this.ConnectedApp = props.stripes.connect(App);
+    this.ConnectedVendorShow = props.stripes.connect(VendorShow);
+    this.ConnectedPackageShow = props.stripes.connect(PackageShow);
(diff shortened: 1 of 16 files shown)


src/index.js
```diff
@@ -9,19 +9,29 @@ import PackageShow from './routes/package/package-show';
...
     match: PropTypes.shape({
       path: PropTypes.string.isRequired
     }).isRequired
   }
 
+  constructor(props) {
+    super(props);
+    this.ConnectedApp = props.stripes.connect(App);
+    this.ConnectedVendorShow = props.stripes.connect(VendorShow);
+    this.ConnectedPackageShow = props.stripes.connect(PackageShow);
```
sadatay: while you could argue that `ConnectedSomething` is clearer semantically, i think you can get away with `this.App = props.stripes.connect(App)`.  it should still be pretty clear what's going on and it looks a little bit cleaner.

PR #32, opened 2017-08-17
Crude Searches

wwilsman:
Thought maybe we could get packages and titles searches out before the demo.

**Lacking Tests** for the two new search pages. But since the vendor search tests are mostly skipped right now, I didn't think it would be too big of a deal.

These search pages should be refactored anyway. We will definitely write tests during that.

### DIFF
--- a/src/routes/search/search-vendors.js
+++ b/src/routes/search/search-vendors.js
@@ -73,6 +73,28 @@ export default class SearchVendors extends Component {
     return !searchVendors || searchVendors.isPending;
   }
 
+  renderButtonGroup() {
+    const { location, history } = this.props;
+    const goTo = (url) => () => history.push(url);
+    const isActive = (url) => url === location.pathname;
+
+    const searchButtons = [
+      ['Vendors', '/eholdings/vendors'],
+      ['Packages', '/eholdings/packages'],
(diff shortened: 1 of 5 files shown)


src/routes/search/search-vendors.js
```diff
@@ -73,6 +73,28 @@ export default class SearchVendors extends Component {
...
+    const isActive = (url) => url === location.pathname;
+
+    const searchButtons = [
+      ['Vendors', '/eholdings/vendors'],
+      ['Packages', '/eholdings/packages'],
+      ['Titles', '/eholdings/titles']
+    ];
+
+    return (
+      <div style={{ float: 'left', marginRight: '1rem' }}>
```
cherewaty: Inline styles are fine for the moment, but ewwwww

wwilsman: Yea, I didn't want to add a `search.css` file in the routes and also didn't want to bother breaking out another component just to float the buttons next to the form. 😅

sadatay:
:fire: :fire: :fire: nice work

any reason these are broken out into a `/search` directory instead of grouping them with the resources they're searching?  not a blocker, just wanted to float the idea.  otherwise looks solid :shipit:

PR #37, opened 2017-08-24
Upgrade stripes-core

cherewaty:
Brings in https://github.com/folio-org/stripes-core/pull/27 and https://github.com/folio-org/stripes-core/pull/33; keeps locked into the Frontside fork for now. We'll switch to upstream master with the next release tag.

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -6,6 +6,9 @@ export default function () {
   this.urlPrefix = okapi.url;
   this.get('/_/version', () => '0.0.0');
   this.get('/_/proxy/modules', []);
+  this.get('/saml/check', {
+    ssoEnabled: false
+  });
 
   // e-holdings endpoints
   this.namespace = 'eholdings';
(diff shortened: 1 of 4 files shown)


mirage/config.js
```diff
@@ -6,6 +6,9 @@ export default function () {
   this.urlPrefix = okapi.url;
   this.get('/_/version', () => '0.0.0');
   this.get('/_/proxy/modules', []);
+  this.get('/saml/check', {
+    ssoEnabled: false
```
cherewaty: I should note that I'm not actually sure if that's the correct response for that endpoint, but it seems to clear up the errors that were being thrown without it.

tests/stripes.config.js
```diff
@@ -18,6 +18,7 @@ export const okapi = { url: 'https://okapi.com', tenant:'tests' };
 export const config = {
   // autoLogin: { username: 'diku_admin', password: 'admin' }
   // logCategories: 'core,redux,connect,connect-fetch,substitute,path,mpath,mquery,action,event,perm,interface,xhr'
+  logCategories: '',
```
cherewaty: By default all redux events were being logged, which made for a very chatty console. I've got no problem tweaking this in the future.

PR #38, opened 2017-08-28
[WIP] Select / deselect package titles (customer resources)

cowboyd:
One of the primary use-cases for the e-holdings application is for digital resource librarians to manage their holdings. To do this they need the ability to mark a title as being part of their holdings (toggling it on and off);

This adds a reducer to the CustomerResourceShowRoute which handles all of the state changes. Currently, all of the side-effects are handled in `mapDispatchToProps`, but we should consider moving this over to using `redux-observable`.

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -47,6 +47,18 @@ export default function () {
     return new Response(200, getHeaders(request), titles.find(matchingCustomerResource.titleId));
   });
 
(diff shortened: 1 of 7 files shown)


cherewaty:
Replaced by https://github.com/thefrontside/ui-eholdings/pull/53

PR #39, opened 2017-08-28
Add custom coverage and hidden metadata to packages

cherewaty:
Required a lot of custom serialization on the mirage side for packages, since `visibilityData`and `customCoverage` are essentially records embedded within a package.

### isHidden
localhost-3000-eholdings-vendors-1- iphone 5

### customCoverage
localhost-3000-eholdings-vendors-1-packages-1 iphone 5

### DIFF
--- a/mirage/serializers/package.js
+++ b/mirage/serializers/package.js
@@ -1,0 +1,38 @@
+import { Serializer } from 'mirage-server';
+
+export default Serializer.extend({
+  embed: true,
+
+  include: ['visibilityData', 'vendor'],
+
+  serialize(response) {
+    let json = Serializer.prototype.serialize.apply(this, arguments);
+    let keyForPrimaryResource = this.keyForResource(response);
+    let unSideloadedJson = json[keyForPrimaryResource];
+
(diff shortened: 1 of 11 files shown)


cherewaty:
Does not include any internationalization (or any formatting, for that matter) of date strings.

mirage/serializers/package.js
```diff
@@ -0,0 +1,39 @@
+import { Serializer } from 'mirage-server';
+
+export default Serializer.extend({
+  embed: true,
+
+  include: ['customCoverage', 'vendor', 'visibilityData'],
+
+  serialize(response) {
+    let json = Serializer.prototype.serialize.apply(this, arguments);
```
cowboyd: Is it a syntax error to use `this.super.serialize(response)`

cherewaty: http://www.ember-cli-mirage.com/docs/v0.3.x/serializers/#serializeresponse-request Mirage borrows Backbone's implementation of extend, so `this.super.serialize(response)` ends up throwing `TypeError: Cannot read property 'serialize' of undefined`

PR #47, opened 2017-08-31
[WIP] Use stripes to format date for locale

cherewaty:
I created a `formatDate()` available in the app's React context. It uses the locale defined by `stripes` to localize dates.


cherewaty:
Replaced with https://github.com/thefrontside/ui-eholdings/pull/49

PR #49, opened 2017-09-05
Format dates with react-intl

cherewaty:
Replaces https://github.com/thefrontside/ui-eholdings/pull/47

Stripes provides `intl`, so I made it available to our app via context. We don't have to know what the stripes `locale` is with this setup.

`formatISODateWithoutTime()` should be pulled out of the `PackageShow` component when it's needed in the `CustomerResourceShow` component.

### DIFF
--- a/mirage/factories/package.js
+++ b/mirage/factories/package.js
@@ -56,8 +56,8 @@ export default Factory.extend({
   withCustomCoverage: trait({
     afterCreate(packageObj, server) {
       let customCoverage = server.create('custom-coverage', {
-        beginCoverage: () => faker.date.past(),
-        endCoverage: () => faker.date.future()
+        beginCoverage: () => faker.date.past().toISOString().substring(0,10),
+        endCoverage: () => faker.date.future().toISOString().substring(0,10)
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "test": "karma start"
   },
   "devDependencies": {
-    "@folio/stripes-core": "thefrontside/stripes-core#master",
+    "@folio/stripes-core": "2.7.0",
```
wwilsman: We'll need #master until they release the changes to the redux store setup

cherewaty: @wwilsman is that this PR? https://github.com/folio-org/stripes-core/pull/38 - if so, I'll change to the commit hash on `folio-org/stripes-core`; I'm very hesitant to point to `folio-org/stripes-core/master`.

wwilsman: That's the one. :+1:

cowboyd: Why not use the release version, or `#master` on our fork ?

cherewaty: Yup, going to switch back to #master on our fork.

PR #53, opened 2017-09-07
Select / deselect package titles (customer resources)

sadatay:
One of the primary use-cases for the e-holdings application is for
digital resource librarians to manage their holdings. To do this they
need the ability to mark a title as being part of their
holdings (toggling it on and off);

This work, along with recent changes to stripes-core, leverages the
concept of 'epics' from 'redux-observable' to ensure async actions
resolve sanely.

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -46,6 +46,18 @@ export default function () {
     }));
   });
 
+  this.put('/vendors/:vendorId/packages/:packageId/titles/:titleId', ({ customerResources, titles }, request) => {
+    let matchingCustomerResource = customerResources.findBy({
+      packageId: request.params.packageId,
+      titleId: request.params.titleId
+    });
+
+    let { isSelected } = JSON.parse(request.requestBody);
+    matchingCustomerResource.update('isSelected', isSelected).save();
+
+    return new Response(200, getHeaders(request), '');
+  });
+
   // Title resources
   this.get('/titles', ({ titles }, request) => {
     const filteredTitles = titles.all().filter((titleModel) => {
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
     "webpack-dev-server": "^2.5.0"
   },
   "dependencies": {
-    "isomorphic-fetch": "^2.2.1"
+    "isomorphic-fetch": "^2.2.1",
+    "redux-actions": "^2.2.1"
   },
   "stripes": {
     "type": "app",
--- a/src/routes/customer-resource/customer-resource-show.js
+++ b/src/routes/customer-resource/customer-resource-show.js
@@ -1,10 +1,28 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import fetch from 'isomorphic-fetch';
+import { connect } from 'react-redux';
+import { handleActions } from 'redux-actions';
+import  { Observable }  from 'rxjs/Observable';
+import { bindActionCreators } from 'redux'
+
+import 'rxjs/add/observable/from';
+import 'rxjs/add/observable/of';
+import 'rxjs/add/operator/catch';
+import 'rxjs/add/operator/switchMap';
 
 import View from '../../components/customer-resource-show';
 
-export default class CustomerResourceShowRoute extends Component {
+const ActionTypes = {
+  'CUSTOMER_RESOURCE_SHOW_ERROR': 'CUSTOMER_RESOURCE_SHOW_ERROR',
+  'CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED_REJECT': 'CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED_REJECT',
+  'CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED_RESOLVE': 'CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED_RESOLVE',
+  'CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED': 'CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED',
+  'CUSTOMER_RESOURCE_SHOW_LOADED': 'CUSTOMER_RESOURCE_SHOW_LOADED',
+  'CUSTOMER_RESOURCE_SHOW_LOAD': 'CUSTOMER_RESOURCE_SHOW_LOAD'
+};
+
+class CustomerResourceShowRoute extends Component {
   static propTypes = {
     match: PropTypes.shape({
       params: PropTypes.shape({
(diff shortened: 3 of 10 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "test": "karma start"
   },
   "devDependencies": {
-    "@folio/stripes-core": "thefrontside/stripes-core#master",
+    "@folio/stripes-core": "thefrontside/stripes-core",
```
wwilsman: Shouldn't we point to the official repo here?

src/routes/customer-resource/customer-resource-show.js
```diff
@@ -1,53 +1,160 @@
...
+
+    const ToggleSelectionEpic = (action$, { getState }) =>
+          action$.ofType(ActionTypes.CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED)
+          .mergeMap(action => {
+            let state = getState().customerResourceShow;
+            let body = JSON.stringify({isSelected: state.isSelected});
+            return Observable.from(fetch(state.endpoint, { headers, method: 'PUT', body }));
+          })
+          .map(json => ({ type: ActionTypes.CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED_RESOLVE }))
+          .catch(error => ({type: ActionTypes.CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED_REJECT }));
```
wwilsman: This will actually terminate the action stream when it's set on the `action$` itself. [Source](https://redux-observable.js.org/docs/recipes/ErrorHandling.html)

src/routes/customer-resource/customer-resource-show.js
```diff
@@ -1,53 +1,160 @@
...
+
+
+    const ResourceLoadedEpic = (action$, { getState }) =>
+          action$.ofType(ActionTypes.CUSTOMER_RESOURCE_SHOW_LOAD)
+          .mergeMap(action => {
+            let { endpoint } = getState().customerResourceShow;
+            return Observable.from(fetch(endpoint, { headers }).then(response => response.json()));
+          })
+          .map(json => ({ type: ActionTypes.CUSTOMER_RESOURCE_SHOW_LOADED, payload: json }))
+          .catch(error => ({type: ActionTypes.CUSTOMER_RESOURCE_SHOW_ERROR }));
```
wwilsman: Same comment here about terminating the `action$`

src/routes/customer-resource/customer-resource-show.js
```diff
@@ -1,53 +1,160 @@
...
+        let endpoint = `${okapi.url}/eholdings/vendors/${vendorId}/packages/${packageId}/titles/${titleId}`;
+        return {...state, endpoint};
+      },
+      [ActionTypes.CUSTOMER_RESOURCE_SHOW_LOADED]: (state, action) => {
+        let resource = action.payload.customerResourcesList[0];
+
+        // slightly hacky, but we're flattening the customerResource
+        // onto the title to create a package title
+        let title = action.payload;
+        delete title.customerResourcesList;
```
wwilsman: Instead, we could do this:

```js
let { customerResourcesList, ...title } = action.payload;
let resource = customerResourcesList[0];
```

src/routes/customer-resource/customer-resource-show.js
```diff
@@ -1,53 +1,160 @@
...
+// These are inline action creators
+const loadRecord = (params) => ({
+  type: 'CUSTOMER_RESOURCE_SHOW_LOAD', params
+});
+const toggleSelected = (event) => ({
+  type: 'CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED'
+});
+
+function mapDispatchToProps(dispatch) {
+  return bindActionCreators({ loadRecord, toggleSelected }, dispatch);
```
wwilsman: Instead of `mapDispatchToProps` we can just pass `{ loadRecord, toggleSelected }` directly to the second argument of connect to do this automatically

src/routes/customer-resource/customer-resource-show.js
wwilsman: You'll want this chained from `Observable.from(...)` instead

src/routes/customer-resource/customer-resource-show.js
```diff
@@ -1,53 +1,160 @@
...
 import PropTypes from 'prop-types';
 import fetch from 'isomorphic-fetch';
+import { connect } from 'react-redux';
+import { handleActions } from 'redux-actions';
+import  { Observable }  from 'rxjs/Observable';
+import { bindActionCreators } from 'redux'
+
+import 'rxjs/add/observable/from';
+import 'rxjs/add/operator/catch';
+import 'rxjs/add/operator/switchMap';
```
elrickvm: Are we using switchMap?

wwilsman: We should. That way concurrent requests only result in a single `LOADED` action

package.json
cherewaty: Let's leave it on master of our fork until the next upstream release.

mirage/config.js
```diff
@@ -46,6 +46,18 @@ export default function () {
     }));
   });
 
+  this.put('/vendors/:vendorId/packages/:packageId/titles/:titleId', ({ customerResources, titles }, request) => {
+    const matchingCustomerResource = customerResources.findBy({
```
cowboyd: Let's use `let` or `const`, but not both for intermediate computations.

My vote is that we use `const` for "true" constants: PI, some URL, etc.... that live at the top-level of a module, but for values that are intermediate to a computation, we use `let`

I am, of course, biased in this matter. So much so, that I wrote an eslint plugin https://www.npmjs.com/package/eslint-plugin-prefer-let

However, I'm more committed to consistency.

src/routes/customer-resource/customer-resource-show.js
```diff
@@ -1,53 +1,160 @@
...
+}
 
-    return showCustomerResource.records.find((title) => {
-      return title.titleId == titleId && title.customerResourcesList.some((pkgTitle) => {
-        return pkgTitle.packageId == packageId && pkgTitle.vendorId == vendorId;
-      });
-    });
-  }
+// These are inline action creators
+const loadRecord = (params) => ({
```
cowboyd: If the variables are not used more than once, Is there any reason not to inline them?

```js
function mapDispatchToProps(dispatch) {
  return bindActionCreators({
    loadRecord(params){ 
      return { type: ActionTypes.CUSTOMER_RESOURCE_SHOW_LOAD, params };
    },
    toggleSelected() {
      return { type: ActionTypes.CUSTOMER_RESOURCE_TOGGLE_SELECTED };
    }
  }, dispatch);
}
```

src/routes/customer-resource/customer-resource-show.js
cowboyd: Is it not chained from `Observable.from` ? 

```js
Observable.from(x).map(y).catch(z);
```

mirage/config.js
wwilsman: I'm ok with restricting `const` to the top-level. My own thoughts on it are: "use `const` if you are not going to change this value later, otherwise use `let`"

src/routes/customer-resource/customer-resource-show.js
wwilsman: This is currently chained from the `action$` stream that results in `Observable.from`. 

From the linked docs: "Here we placed the `.catch()` inside our `.mergeMap()`, but after our AJAX call; this is important because if we let the error reach the `action$.ofType()`, it will terminate it and no longer listen for new actions."

sadatay: since the result of the `mergeMap` is returning an Observable, the way I understood it is that `map` and `catch` were both hanging off that Observable.  in the example they're being called inside the mergeMap and that's not the case here, right?

src/routes/customer-resource/customer-resource-show.js
sadatay: ✨ nice

src/routes/customer-resource/customer-resource-show.js
sadatay: :sparkles: also nice!

src/routes/customer-resource/customer-resource-show.js
wwilsman: Right, they should be called inside of the merge map so the error never reaches the `action$` stream. While the observable is returned inside the stream, it doesn't create a new stream. `mergeMap` returns the current stream with the return value as it's `subject`, in this case that's the `action$.ofType` stream with the observable as it's `subject`.

What we want to return inside of `mergeMap` is an observable that handles it's own stream errors

wwilsman: I hope that made sense

src/routes/customer-resource/customer-resource-show.js
sadatay: so i read up on `switchMap` and vaguely get it, but far from completely understand the difference.  i converted the `mergeMap`s to `switchMap`s and everything still seems sturdy.

src/routes/customer-resource/customer-resource-show.js
cowboyd: Oops, I read it wrong. Yes, the map and catch are hanging off of the `switchMap`, for some reason I could swear they were off the the ajax observable.

PR #57, opened 2017-09-11
Refactor customer resource show route 🦆

wwilsman:
This pulls the original reducer and epics out into their own "duck" file to be used with the `eholdings` root reducer / epic, and splits the customer resource model's state into `record` and `toggle` pieces to reflect the two different operations.

This is related to stripping `stripes-connect` from the app, as all other routes will have their own reducers and epics defined in the `redux` directory as well.

### DIFF
--- a/src/routes/customer-resource/customer-resource-show.js
+++ b/src/routes/customer-resource/customer-resource-show.js
@@ -2,25 +2,9 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import fetch from 'isomorphic-fetch';
 import { connect } from 'react-redux';
-import { handleActions } from 'redux-actions';
-import  { Observable }  from 'rxjs/Observable';
-import { bindActionCreators } from 'redux'
-
-import 'rxjs/add/observable/from';
-import 'rxjs/add/observable/of';
-import 'rxjs/add/operator/catch';
-import 'rxjs/add/operator/switchMap';
+import { requestResource, toggleSelected } from '../../redux/customer-resource';
 
 import View from '../../components/customer-resource-show';
-
-const ActionTypes = {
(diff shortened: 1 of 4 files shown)


src/routes/customer-resource/customer-resource-show.js
```diff
@@ -145,22 +33,11 @@ class CustomerResourceShowRoute extends Component {
...
-const loadRecord = (params) => ({
-  type: 'CUSTOMER_RESOURCE_SHOW_LOAD', params
-});
-const toggleSelected = (event) => ({
-  type: 'CUSTOMER_RESOURCE_SHOW_TOGGLE_SELECTED'
-});
-
-export default connect(mapStateToProps, { loadRecord, toggleSelected })(CustomerResourceShowRoute);
+export default connect(
+  ({ eholdings: { customerResource }}) => ({
```
sadatay: i'm not sure i fully understand this line, particularly the `eholdings` bit.  nearest i can guess is that this namespaces this component's state under an `eholdings` key?

wwilsman: So with #52, I moved the `addReducer`/`addEpic` stuff to the root of our app under `this.addReducer('eholdings', rootReducer)`. This way when our components are initially mounted we have access to the default state and any other state other components may interact with.

Since `addReducer` adds the key `eholdings` to the state, all of our leaf reducers become part of that `eholdings` piece.

This line simply destructures that bit. Long form would be:

```js
(state) => {
  let customerResource = state.eholdings.customerResource
  ...
}
```

sadatay: 👍 awesome.  that's much better than scattering the `addEpic` calls throughout the routes

sadatay:
🦆 looks groovy to me

wwilsman:
@sadatay you should hit that approve button. :wink:

PR #59, opened 2017-09-11
Refactor search state to use promise-like booleans

wwilsman:
Renamed the following search state properties and added `isResolved`

```
isLoading -> isPending
isErrored -> isRejected
records -> content
```


wwilsman:
@cherewaty Make it more promise-like. Eventually all request can share the same request epic (and maybe reducer logic) if they all have `content` instead of `records`/`record`/`vendor`/`package`/`whatever`

PR #60, opened 2017-09-11
Enforce ESLint rules

cherewaty:
When running the dev server, ESLint will emit errors:
screen shot 2017-09-11 at 4 48 49 pm

`yarn test` will not run `ESLint` assertions, but `yarn lint` will.

Travis CI builds will fail with ESLint errors: https://travis-ci.org/thefrontside/ui-eholdings/builds/274360479

Only pays attention to the `src` directory for now. We should gradually transition over to using the same ruleset as `stripes-core`.

### DIFF
--- a/src/redux/search.js
+++ b/src/redux/search.js
@@ -5,6 +5,7 @@ import 'rxjs/add/observable/from';
 import 'rxjs/add/operator/switchMap';
 import 'rxjs/add/operator/pluck';
 import 'rxjs/add/operator/catch';
+import merge from 'lodash/merge';
 
 // action types
 const EHOLDINGS_SEARCH = 'EHOLDINGS_SEARCH';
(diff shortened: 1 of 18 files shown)


src/redux/search.js
```diff
@@ -5,6 +5,7 @@ import 'rxjs/add/observable/from';
 import 'rxjs/add/operator/switchMap';
 import 'rxjs/add/operator/pluck';
 import 'rxjs/add/operator/catch';
+import _ from 'lodash';
```
wwilsman: We should actually just import `merge` here since that's all that's used in this file.

```js
import merge from 'lodash/merge';
```

src/redux/search.js
```diff
@@ -69,9 +70,8 @@ export function searchEpic(action$, { getState }) {
     .switchMap((action) => {
       const { okapi } = getState();
       const { searchType, query, options } = action;
-      const { endpoint, defaults, recordsKey = searchType } = options;
+      const { endpoint, recordsKey = searchType } = options;
 
-      const searchQuery = { ...query, ...defaults };
       const url = `${okapi.url}/${endpoint}?${queryString.stringify(query)}`;
```
wwilsman: `defaults` is totally meant to be used (since RM-API requires certain params). This is actually not working in production now because of my mistake here.

This line should instead be: `queryString.stringify(searchQuery)`

wwilsman: Although it should be:
```js
const searchQuery = { ...defaults, ...query }
```
So the query isn't overridden by the defaults

cherewaty: Fixed.

PR #64, opened 2017-09-13
Fix convergent negation test

wwilsman:
This fixes tests similar to this: 

```js
it('does not display some thing', function() {
  expect(PageObject.someValue).to.equal('');
});
```

With convergent assertions, these will still pass because `someValue` hasn't had a chance to update to the side effects from `beforeEach` (i.e. loading a record).

To see this, simply move one of these negation tests to someplace it should fail, and **it will still pass**.

We could assert that a record has been loaded before our assertions are run, however sometimes the DOM may be milliseconds behind in rendering and the convergent assertion will still pass.

Instead we should use `it.still` which assures that the assertion is true for a given amount of time:

```js
it.still('does not display some thing', function() {
  expect(PageObject.someValue).to.equal('');
});
```

However, by default, the timeout is 200ms. And so for each `it.still`, our tests would become 200ms slower.

The most optimal solution would be to get mocha to run our tests in parallel, after side effects. This way only an entire suit would be delayed by 200ms no matter how many `it.still`s were used inside.

### DIFF
--- a/tests/customer-resource-show-embargo-test.js
+++ b/tests/customer-resource-show-embargo-test.js
@@ -52,11 +52,11 @@ describeApplication('CustomerResourceShowEmbargos', function() {
       });
     });
 
-    it('does not display the managed embargo section', function() {
+    it.still('does not display the managed embargo section', function() {
       expect(CustomerResourceShowPage.managedEmbargoPeriod).to.equal('');
     });
 
-    it('does not display the custom embargo section', function() {
+    it.still('does not display the custom embargo section', function() {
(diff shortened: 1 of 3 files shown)


cowboyd:
What about passing an option:

```
it.still('is in flight', function() {
  //...run assection
}, 50);
```

wwilsman:
@cowboyd That actually is an available option already implemented in `tests/it-will.js`. It just defaults to 200. We can change that to default to 50. But is 50ms enough to verify something has _not_ loaded?

wwilsman:
:point_up: Lowered default timeout of `it.still` to 50ms

PR #65, opened 2017-09-13
Remove stripes connect

wwilsman:
After removing stripes connect and using a reducer/epic combo from one of the routes, it became evident that most of our redux code will be centered around making and handling requests.

All pieces of our state had a similar structure to each other:

```js
{
  isPending: false,
  isResolved: false,
  isRejected: false,
  content: null,
  error: null
}
```

And all of the epics also had a similar structure to each other:

```js
function epic(action$, { getState }) {
  return action$.ofType(SOME_REQUEST_TYPE)
    .switchMap((action) => {
      let { okapi } = getState();
      let url = buildOkapiEndpoint(okapi, action);
      let headers = { 'X-Okapi-Tenant': okapi.tenant };
      let request = fetch(url, { headers, /* ... */ });

      return Observable.from(request.then((res) => res.json())
        .map((payload) => ({ type: RESOLVE_REQUEST, payload })
        .catch((error) => Observable.of({ type: REJECT_REQUEST, error });
    });
}
```

I wrapped these similar structures in `createRequestReducer` and `createRequestEpic` so we can import them into other duck files to create specific resource request ducks.

They're used like so:

```js
/* /redux/vendor.js */

import { 
  createRequestCreator,
  createRequestReducer, 
  createRequestEpic 
} from './request';

// the name passed to the creator should match the epic/reducer
export const getVendor = createRequestCreator('vendor', /* { method: 'GET' } */);

export const vendorReducer = createRequestReducer({
  name: 'vendor',                                      // name must match the epic name
  initialContent: {},                                  // initial `content` property
  handleActions: {},                                   // other actions to respond to
  handleRequests: {},                                  // other requests to respond to
  initialState: {}                                     // initial state to merge with defaults
});

export const vendorEpic = createRequestEpic({
  name: 'vendor',                                      // name must match the reducer name
  defaultParams: false,                                // optional default query params (false means no query)
  endpoint: (data) => `/eholdings/vendors/${data.id}`, // provide or generate an endpoint 
  normalize: (payload) => payload[keyForRecord],       // transform incoming data
  serialize: (data) => data                            // transform outgoing data
});
```
```js
/* /routes/vendor/vendor-show.js */

import { getVendor } from '../../redux/vendor';

/* ... */
  componentWillMount() {
    let { vendorId } = this.props.match.params;
    this.props.getVendor({ id: vendorId });
  }
/* ... */

export default connect(
  ({ eholdings: { vendor }}) => ({ vendor }),
  { getVendor }
)(VendorShowRoute);
```

### DIFF
--- a/src/redux/request.js
+++ b/src/redux/request.js
@@ -12,8 +12,21 @@ export const REQUEST_MAKE = '@@ui-eholdings/REQUEST_MAKE';
 export const REQUEST_RESOLVE = '@@ui-eholdings/REQUEST_RESOLVE';
 export const REQUEST_REJECT = '@@ui-eholdings/REQUEST_REJECT';
 
-// action creators
+/**
+ * Creates an action creator to start a request for a specific named epic
+ *
+ * @param {string} name - name of the request epic to trigger
+ * @param {object} [options={}] - additional options for this request
+ * (currently only `method` is supported)
+ * @returns {function} named request action creator
+ */
 export function createRequestCreator(name, options = {}) {
+  /**
+   * Named request action creator
+   *
+   * @param {object} [data={}] - data to provide to the request (query
+   * params for GET/DELETE; body data for PUT/POST)
+   */
   return (data = {}) => ({
     type: REQUEST_MAKE,
     name,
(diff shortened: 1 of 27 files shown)


src/redux/request.js
```diff
@@ -0,0 +1,166 @@
...
+import startsWith from 'lodash/startsWith';
+import { handleActions as reduxHandleActions } from 'redux-actions';
+import { Observable } from 'rxjs/Observable';
+import 'rxjs/add/operator/switchMap';
+import 'rxjs/add/observable/from';
+import 'rxjs/add/observable/of';
+import 'rxjs/add/operator/catch';
+
+// action types
+export const REQUEST_MAKE = '@@ui-eholdings/REQUEST_MAKE';
```
cowboyd: What's the convention around the `@@` sigil. I've seen it used for internal redux events.

wwilsman: It's basically the agreed-on convention of namespacing actions.

I think since this app is part of another larger ecosystem, we should namespace our actions to prevent our reducers being triggered from other app's poorly-named actions.

src/redux/request.js
```diff
@@ -0,0 +1,271 @@
...
+ * In order for request reducers to receive actions from a specific
+ * request epic, the names of each reducer/epic combo must match
+ * exactly. Additonally, it must be unique to a reducer/epic combo,
+ * otherwise other epics with the same name will trigger the reducer.
+ *
+ * if `handleActions` is provided, other actions can be included in
+ * the reducer to further alter it's state. When one of the request
+ * actions are provided, the resulting state is reduced again by the
+ * default request action handler to maintain the promise-like
+ * properties of the state.
```
cowboyd: An example of this would be awesome. Also, something indicating which reduction happens _first_

wwilsman:
I forgot to mention this in the description, but this PR also prints out our server errors and un-skips the related error tests.

src/redux/request.js
```diff
@@ -0,0 +1,271 @@
...
+ * empty object. Key-values will be used as defaults for the query.
+ *
+ * @param {string} name - name of this request epic; must match a
+ * corrisponding request reducer
+ * @param {string|function} [endpoint=''] - string or generator which when
+ * combined with `okapi.url` will become the request URL
+ * @param {object|false} [defaultParams=false] - when `false` this request has
+ * no query options; when an object is provided, it will be merged with
+ * `data` before being stringified and appended to GET and DELETE URLs
+ * @param {function} [normalize=identity] - transforms data being
```
cowboyd: Let's name this operation `deserialize` so that they are clear inverses.

src/redux/request.js
```diff
@@ -0,0 +1,271 @@
...
+        } else if (defaultParams) {
+          let query = { ...defaultParams, ...data };
+
+          if (Object.keys(query)) {
+            let searchQuery = '';
+
+            // "hack" to prevent query-string from removing/encoding null searches
+            if (query.search === '%00') {
+              let { search, ...q } = query;
+              searchQuery = `search=${search}&${queryString.stringify(q)}`;
```
cowboyd: I know that the RMAPI uses the `search` parameter, but the standard web convention is `q`. It's an issue for our API, but I think we should use it in the URL, and also in our API. (Also, pretty sure that this is what CQL uses)

src/routes/customer-resource/customer-resource-show.js
```diff
@@ -14,31 +14,42 @@ class CustomerResourceShowRoute extends Component {
...
     let { vendorId, packageId, titleId } = this.props.match.params;
-    this.props.requestResource(vendorId, packageId, titleId);
+    this.props.makeRequest('customer-resource', { vendorId, packageId, titleId });
   }
 
+  toggleSelected = () => {
+    let { vendorId, packageId, titleId } = this.props.match.params;
+    let { isSelected } = this.props.customerResource.content;
+
+    this.props.makeRequest('customer-resource-toggle-selected', {
```
cowboyd: I go back and forth on these. It feels like the strings such as `customer-resource-toggle-selected'` belongs to the ducks. Have you thought about also `createRequestActions` also to sit along side `createRequestEpic` and `createRequestReducer`?

wwilsman: I did think about that. Should that be a helper, or maybe just a simple:

```js
const toggleSelected = makeRequest.bind(null, 'customer-resource-toggle-selected');
```

src/redux/request.js
wwilsman: I like `q` much better, but as of right now, there is no mapping between front-end query params and back-end query params (I can add it if it's worth it). 

I think we could just keep this until we implement `q` in our own API. That way we can remove this whole `null` hack and have mod-kb-ebsco handle that instead too.

src/routes/customer-resource/customer-resource-show.js
cowboyd: Yeah, I like that, althoug the bind syntax is a tad esoteric, we can just use a function literal.

PR #70, opened 2017-09-14
Fix some package / package-title attribute displays

cherewaty:
Solves two problems shown in the screenshot below:

- Don't show content type if it's not returned from RM API.
- If a type or subtype of an identifier is outside of the bounds of what we know, call it an "Unknown Identifier." We were being returned ids not defined in the RM API docs https://developer.ebsco.com/rmapireference#identifier. It would be much better to receive the string describing the identifier from the RM API, instead of having to manually map ids.

screen shot 2017-09-14 at 1 19 03 pm


sadatay:
🐳 👍

PR #72, opened 2017-09-14
Ignore identifiers that aren’t ISSN or ISBN

cherewaty:
We only want to see these identifiers.

Subtype | Type | UI Display Label
-- | -- | --
Print = 1 | ISSN = 0 | ISSN (Print)
Print = 1 | ISBN = 1 | ISBN (Print)
Online= 2 | ISSN = 0 | ISSN (Online)
Online = 2 | ISBN = 1 | ISBN (Online)

(I'm also showing ISSN and ISBN with subtype 0.)

### DIFF
--- a/tests/title-show-test.js
+++ b/tests/title-show-test.js
@@ -23,7 +23,8 @@ describeApplication('TitleShow', function() {
       this.server.create('identifier', { type: 1, subtype: 1, id: '978-0547928210' }),
       this.server.create('identifier', { type: 1, subtype: 1, id: '978-0547928203' }),
       this.server.create('identifier', { type: 1, subtype: 2, id: '978-0547928197' }),
-      this.server.create('identifier', { type: 1, subtype: 0, id: '978-0547928227' })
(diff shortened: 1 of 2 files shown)


tests/title-show-test.js
```diff
@@ -23,7 +23,8 @@ describeApplication('TitleShow', function() {
       this.server.create('identifier', { type: 1, subtype: 1, id: '978-0547928210' }),
       this.server.create('identifier', { type: 1, subtype: 1, id: '978-0547928203' }),
       this.server.create('identifier', { type: 1, subtype: 2, id: '978-0547928197' }),
-      this.server.create('identifier', { type: 1, subtype: 0, id: '978-0547928227' })
+      this.server.create('identifier', { type: 1, subtype: 0, id: '978-0547928227' }),
+      this.server.create('identifier', { type: 8, subtype: 49, id: 'someothertypeofid' })
```
wwilsman: Was there meant to be a matching test for this?

cherewaty: Yep, I totally meant to do that. Added.

PR #73, opened 2017-09-19
View and set backend settings

sadatay:
As it stood, the value for the RMAPI customer ID and API key were per
deployment. Now however, it can be configured via the
/eholdings/configuration endpoint so that each tenant can maintain
their own set of credentials. We needed a way to detect if the
eholdings app is usable, and if its backend is configured
correctly, and also to be able to change credentials.

This change checks on every eholdings application entry to see if
infact there _is_ a  backend for eholdings and gives you instruction
on how to download it. If there is a backend, it will ensure that it
is configured properly, and if not, it will direct the user to a new
settings screen with which a tenant admin can view the current
configuration and make changes where necessary.

In order to support development against server configurations which
are hard to arrive at (such as a borked backend) the ability to start
the app in different scenarios was added. For example to start the app
without a backend, you would invoke the command:

```
yarn start --mirage no-backend
```

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -1,11 +1,31 @@
 import { okapi } from 'stripes-config'; // eslint-disable-line import/no-unresolved
 import { Response } from 'mirage-server';
 
-export default function () {
-  // Okapi configs
+export default function configure() {
   this.urlPrefix = okapi.url;
+
+  // okapi endpoints
   this.get('/_/version', () => '0.0.0');
-  this.get('/_/proxy/modules', []);
+
+  this.get('/_/proxy/modules', [{
+    id: 'mod-kb-ebsco',
+    name: 'kb-ebsco'
+  }]);
+
+  this.get('/_/proxy/modules/mod-kb-ebsco', {
+    id: 'mod-kb-ebsco',
+    name: 'kb-ebsco',
+    provides: [{
+      id: 'eholdings',
+      version: '0.1.0',
+      handlers: [{
+        methods: ['GET', 'POST', 'PUT', 'PATCH', 'DELETE'],
+        pathPattern: '/eholdings/*'
+      }]
+    }],
+    permissionSets: []
+  });
+
   this.get('/saml/check', {
     ssoEnabled: false
   });
@@ -44,6 +64,32 @@ export default function () {
 
   // e-holdings endpoints
   this.namespace = 'eholdings';
+
+  // status
+  this.get('/status', {
+    data: {
+      type: 'status',
+      attributes: {
+        'is-configuration-valid': true
+      }
+    }
+  });
+
+  // configuration
+  this.get('/configuration', {
+    data: {
+      type: 'configuration',
+      attributes: {
+        'customer-id': 'some-valid-customer-id',
+        'api-key': 'some-valid-api-key'
+      }
+    }
+  });
+
+  this.put('/configuration', (_, request) => {
+    let data = JSON.parse(request.requestBody);
+    return data;
+  });
 
   // Package resources
   this.get('/packages', ({ packages }, request) => {
--- a/mirage/scenarios/unconfigured-backend.js
+++ b/mirage/scenarios/unconfigured-backend.js
@@ -1,0 +1,20 @@
+export default function unconfiguredBackendScenario(server) {
+  server.get('/status', {
+    data: {
+      type: 'status',
+      attributes: {
+        'is-configuration-valid': false
+      }
+    }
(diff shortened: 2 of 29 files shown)


mirage/scenarios/unconfigured-backend.js
```diff
@@ -0,0 +1,6 @@
+export default function noBackendScenario(server) {
```
wwilsman: This should be `unconfiguredBackend`

src/redux/request.js
```diff
@@ -241,14 +241,15 @@ export function createRequestEpic({
...
+        let state = getState();
+        let { okapi } = state;
         let { method = 'GET' } = options;
         let body, headers = { 'X-Okapi-Tenant': okapi.tenant };
         let url = endpoint;
 
         // endpoint url
         if (typeof endpoint === 'function') {
-          url = `${okapi.url}/${endpoint(data, options)}`;
+          url = `${okapi.url}/${endpoint(data, options, state)}`;
```
wwilsman: This change looks left over since it's unused

src/routes/application.js
```diff
@@ -0,0 +1,70 @@
...
+  renderApplication() {
+    let { children, application } = this.props;
+
+    if (application.isPending) {
+      return (<div>...loading</div>);
+    } else if (application.isResolved) {
+      let config = application.content;
+
+      return config.isValid ? children : this.renderNotValidConfig();
+    }
```
wwilsman: We need to handle `isRejected` as well

```js
if (application.isPending) {
  return (<div>...loading</div>);
} else if (application.isRejected) {
  return this.renderNotValidConfig(); // or this.renderConfigError()
} else if (application.isResolved) {
  return !application.content.isValid
    ? this.renderNotValidConfig()
    : children;
}
```

src/routes/application.js
```diff
@@ -0,0 +1,70 @@
...
+
+  renderNotValidConfig() {
+    return (
+      <div className="eholdings-unconfigured-backend" data-test-eholdings-unconfigured-backend>
+        <h1>Error: Knowledge Base Not Configured</h1>
+        <p>The E-Holdings application detected the presence of a knowledge base, but the knowledge base does not appear to be configured.  Please configure the knowledge base with your credentials on the settings page..</p>
+      </div>
+    );
+  }
+}
```
wwilsman: Can we get some screenshots of these in the PR? 😉

tests/backend-configuration-test.js
```diff
@@ -0,0 +1,79 @@
...
+        });
+
+      });
+    });
+
+    describe('configuring the mod-go-kb backend', function() {
+      it('has yet to be implemented');
+    });
+  }
+});
```
wwilsman: We should also add a test + scenario for the backend sending an error response (handling `.isRejected`).

tests/helpers.js
```diff
@@ -5,8 +5,10 @@ import chaiJquery from 'chai-jquery';
 import $ from 'jquery';
 import { render, unmountComponentAtNode } from 'react-dom';
 import { convergeOn } from './it-will';
+import startMirage from '../mirage/start';
 import TestHarness from './harness';
 
+
```
wwilsman: ✂️

src/redux/request.js
```diff
@@ -241,7 +241,8 @@ export function createRequestEpic({
     action$.ofType(REQUEST_MAKE)
       .filter((action) => action.name === name)
       .switchMap(({ data = {}, options = {} }) => {
-        let { okapi } = getState();
+        let state = getState();
+        let { okapi } = state;
```
wwilsman: This is unnecessary as well now. You can just ignore the changes in this whole file

cherewaty:
Blocked by https://github.com/thefrontside/mod-kb-ebsco/pull/11

cherewaty:
LtR <-> RtL

http://g.recordit.co/YBPxmMOARv.gif

mirage/config.js
```diff
@@ -1,11 +1,31 @@
...
+  }]);
+
+  this.get('/_/proxy/modules/mod-kb-ebsco', {
+    id: 'mod-kb-ebsco',
+    name: 'kb-ebsco',
+    provides: [{
+      id: 'eholdings',
+      version: '0.1.0',
+      handlers: [{
+        methods: ['GET', 'POST', 'PUT', 'PATCH', 'DELETE'],
```
sadatay: this isn't quite accurate, but should be fine for how we're using it

PR #77, opened 2017-09-27
Rework search layout

cherewaty:
Now responsive with two panes. In the middle range the search pane will pop out with a vignette behind it.

Biggest size:
localhost-3000-eholdings-search-packages-search e ipad

Middle size:
localhost-3000-eholdings-search-packages-search e ipad 1

Smallest size:
localhost-3000-eholdings-search-packages-search e iphone 5

### DIFF
--- a/src/components/search-form/search-form.css
+++ b/src/components/search-form/search-form.css
@@ -1,13 +1,12 @@
 @import '@folio/stripes-core/src/components/variables';
 
 .search-form-container {
-  margin: 1em 0 0.5em 0;
-  width: 100%;
+  padding: 1em;
 }
 
 .search-switcher {
-  float: left;
   margin: 0 1em;
+  width: 100%;
 
   & a {
     display: inline-block;
--- a/src/components/search-form/search-form.css
+++ b/src/components/search-form/search-form.css
@@ -68,7 +68,17 @@
 }
 
 .search-submit {
-  display: none;
+  background-color: var(--primary);
+  border: 1px solid var(--primary);
+  border-radius: 0.4em;
+  color: #fff;
+  line-height: 2em;
+  margin-top: 1em;
+  width: 100%;
+
+  &:hover{
+    background-color: color(var(--primary) shade(8%));
+  }
 }
 
 .search-results-list {
--- a/src/components/search-form/search-form.css
+++ b/src/components/search-form/search-form.css
@@ -5,51 +5,70 @@
 }
 
 .search-switcher {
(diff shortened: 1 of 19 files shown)


src/redux/request.js
```diff
@@ -11,6 +11,7 @@ import 'rxjs/add/operator/catch';
 export const REQUEST_MAKE = '@@ui-eholdings/REQUEST_MAKE';
 export const REQUEST_RESOLVE = '@@ui-eholdings/REQUEST_RESOLVE';
 export const REQUEST_REJECT = '@@ui-eholdings/REQUEST_REJECT';
+export const REQUEST_CLEAR = '@@ui-eholdings/REQUEST_CLEAR';
```
sadatay: not sure how i feel about the naming here.  the other `REQUEST_` constants deal with http requests, and also set special state in the reducer based on those external interactions.  `REQUEST_CLEAR` doesn't quite fit that definition, since it doesn't make any kind of request and simply manipulates redux state.

cherewaty:
Added commit with prettier form style (and hid the search submit button, which reflects the desired prototypes but I need to ask around about the accessibility implications of).

localhost-3000-eholdings-search-titles

sadatay:
looks good to me overall :shipit:.  thanks for pulling my CSS coat!

src/components/search-form/search-form.css
```diff
@@ -1,56 +1,74 @@
...
-  margin-left: 0.5em;
-  background-color: var(--primary);
-  border: 1px solid var(--primary);
-  border-radius: 0.4em;
-  color: #fff;
-
-  &:hover{
-    background-color: color(var(--primary) shade(8%));
-  }
+  display: none;
```
wwilsman: Yea, this is def not accessible.

If it _really_ needs to be hidden, you should use some screen-reader friendly css:

```css
.visuallyhidden { 
  position: absolute; 
  overflow: hidden; 
  clip: rect(0 0 0 0); 
  height: 1px; 
  width: 1px; 
  margin: -1px; 
  padding: 0; 
  border: 0; 
}
```

But imho, this is not ideal. How would a mouse-jockey click search? Where's the indication that you have to press enter to perform the search?

cherewaty: @wwilsman yeah, let's just kick the can down the road and stick with a solution that we _know_ works. I put the button back. I'll follow up with other FOLIO UX folks on what our standard will be. 

localhost-3000-eholdings-search-packages iphone 5

PR #81, opened 2017-10-04
Lint ALL the things

cherewaty:
Enforces linting for the `demo`, `mirage`, and `tests` directories in addition to `src` that was already being checked. For good measure, I also fixed lint issues with `karma.conf.js`.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -21,16 +21,14 @@ webpackConfig.resolve.alias['stripes-config'] = path.resolve(__dirname, 'tests/s
 // `stripes-config` module does not get overriden.
 // see `@folio/stripes-config`
 // see `@folio/stripes-core/webpack.config.base.js`
-webpackConfig.module.rules = webpackConfig.module.rules.filter(rule => {
-  return !rule.use || !rule.use.some(use => use.loader === '@folio/stripes-config');
-});
+webpackConfig.module.rules = webpackConfig.module.rules.filter(rule => !rule.use || !rule.use.some(use => use.loader === '@folio/stripes-config'));
 
 // make sure that the NODE_ENV is available in browser code.
 webpackConfig.plugins.push(new webpack.EnvironmentPlugin({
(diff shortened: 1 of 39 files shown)


karma.conf.js
```diff
@@ -21,16 +21,14 @@ webpackConfig.resolve.alias['stripes-config'] = path.resolve(__dirname, 'tests/s
 // `stripes-config` module does not get overriden.
 // see `@folio/stripes-config`
 // see `@folio/stripes-core/webpack.config.base.js`
-webpackConfig.module.rules = webpackConfig.module.rules.filter(rule => {
-  return !rule.use || !rule.use.some(use => use.loader === '@folio/stripes-config');
-});
+webpackConfig.module.rules = webpackConfig.module.rules.filter(rule => !rule.use || !rule.use.some(use => use.loader === '@folio/stripes-config'));
```
wwilsman: Whoa, this is kinda hard to read imo. Can we make this more readable but still respect eslint?

mirage/config.js
```diff
@@ -14,32 +13,24 @@ export default function () {
...
-  });
+  this.get('/vendors/:vendorId/packages/:packageId', ({ packages }, request) => packages.findBy({
+    vendorId: request.params.vendorId,
+    id: request.params.packageId
+  }));
 
-  this.get('/vendors/:vendorId/packages/:packageId/titles', ({ customerResources }, request) => {
-    return customerResources.where({ packageId: request.params.packageId });
-  });
+  this.get('/vendors/:vendorId/packages/:packageId/titles', ({ customerResources }, request) => customerResources.where({ packageId: request.params.packageId }));
```
wwilsman: Can we just disable this eslint rule? having _really_ long lines like this is way worse imo.

cherewaty: I'm fine with turning off `arrow-body-style` - now both these are back to more legible code over multiple lines.

PR #83, opened 2017-10-04
Fix frontpage mobile styles

wwilsman:
Making sure the nav bar is visible and having contained apps scroll underneath it seems like it should be part of `stripes-core`. However there is currently no such enforcements in place (as there is no comprehensive responsive styles in most of stripes).

App's typically use `Pane` and `Paneset` components to render their content, and these components _do_ take the header height into account in their styling. However, our `front` component does not use a `Pane`, and so we need to ensure that the header height is taking into account ourselves. 

### Before

before

### After

after

### DIFF
--- a/demo/front.css
+++ b/demo/front.css
@@ -1,7 +1,9 @@
 .front {
-  margin: 3em auto;
+  height: calc(100vh - 44px);
+  margin: 0 auto;
   max-width: 50em;
-  padding: 0 2em;
+  overflow: auto;
+  padding: 3em 2em 0;
 }
 
 .logo {


cowboyd:
> However, our front component does not use a Pane, and so we need to ensure that the header height is taking into account ourselves.

Should we just do that?

wwilsman:
@cowboyd We aren't using `stripes-components` Pane's in our app either. We're implementing our own for now due to lack of responsive stuff

cowboyd:
Sounds like they're a pane in the ass.

PR #86, opened 2017-10-10
Hide search pane vignette at large sizes

cherewaty:
If the browser window was resized, the search pane vignette (the partially transparent background between the search form and the results) was still showing up. It should never appear above the largest breakpoint.

### DIFF
--- a/src/components/search-pane-vignette/search-pane-vignette.css
+++ b/src/components/search-pane-vignette/search-pane-vignette.css
@@ -12,3 +12,9 @@
     display: none;
   }
 }
+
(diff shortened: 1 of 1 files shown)


src/components/search-pane-vignette/search-pane-vignette.css
```diff
@@ -12,3 +12,9 @@
     display: none;
   }
 }
+
+@media(min-width: 961px){
```
cherewaty: Future Todo: break out breakpoints into variables.

PR #89, opened 2017-10-18
Create toggle switch component

carolegodfrey:
Updated Customer Resource Selected Toggle by moving it into a separate component and adding animation as shown below:
http://g.recordit.co/WOGSDUvzrr.gif


sadatay:
not sure what the flakey test is about.  otherwise everything looks good to me, but we should investigate that failure.  could just need a rebase?  or it could require some changes to the customer resource page object

.gitattributes
```diff
@@ -0,0 +1 @@
+*.js text eol=lf
```
cherewaty: This makes line endings sane on Windows (so it's not trying to convert them into CRLF): https://github.com/Microsoft/vscode/issues/2957

PR #94, opened 2017-10-23
Three pane search layout

wwilsman:
## Purpose
Add a details pane to the search layout

## Approach
The main child routes for each resource is now nested under the search route. This search route detects the presence of the `searchType` query parameter and will render the search form and results. When the `searchType` parameter is missing, it will instead just render it's children (which are the various show routes).

The reason to pull the search type out into it's own parameter is so we can view a resource unrelated to the type of resource we're searching at the same time. For example, while looking at the details of a package, you might want to search for different vendors. With the previous route system there is no way to view `packages/:packageId` while searching at `/search/vendors`. The new query parameter for search type allows this via `packages/:packageId?searchType=vendors`.

~Additionally to this route change, the initial load of any of the search routes will now perform an empty search, returning the first page of all results from the API.~

## Todos
- [x] ~Look into a bug where when switching between search types, the new empty search is performed but the results component is not re-rendered when the search is resolved~

  When this new search happens via `componentWillReceiveProps` it seems as though the `connect` function does not receive the updated state during the request actions.

  When a request is fired this way, a re-render does happen with every action, however the search leaf returned from the `connect` function remains the same. Once the request is complete, switching to another search type and back will trigger the new search state to be rendered.

  For now, this auto-search feature is removed so we can push out the three-pane search layout.

  Additionally, the initial empty search results for packages do not seem to have names. So the results list only displays "Not Selected" for those packages.

## Screenshots
nv8xjj31g1

localhost-3000-eholdings-vendors-1-search e searchtype vendors ipad

localhost-3000-eholdings-vendors-1-search e searchtype vendors ipad 1

### DIFF
--- a/src/components/package-show.js
+++ b/src/components/package-show.js
@@ -108,7 +108,7 @@ export default function PackageShow({ vendorPackage, packageTitles }, { intl }) 
           {vendorPackage.error.length ? vendorPackage.error[0].message : vendorPackage.error.message}
         </p>
       ) : (
-        <Icon icon='spinner-ellipsis' />
+        <Icon icon="spinner-ellipsis" />
       )}
     </div>
   );
--- a/src/components/package-show.js
+++ b/src/components/package-show.js
@@ -2,8 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { Link } from 'react-router-dom';
-import Paneset from '@folio/stripes-components/lib/Paneset';
-import Pane from '@folio/stripes-components/lib/Pane';
 import Icon from '@folio/stripes-components/lib/Icon';
 import KeyValueLabel from './key-value-label';
 import List from './list';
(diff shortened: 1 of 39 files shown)


src/components/package-show.js
```diff
@@ -22,100 +20,96 @@ export default function PackageShow({ vendorPackage, packageTitles }, { intl })
...
-        <Pane defaultWidth="100%">
-          {vendorPackage.isResolved ? (
-            <div>
-              <div style={{ margin: '2rem 0' }}>
-                <KeyValueLabel label="Package">
-                  <h1 data-test-eholdings-package-details-name>
-                    {packageRecord.packageName}
-                  </h1>
-                </KeyValueLabel>
+    <div style={{ padding: '0 1rem' }} data-test-eholdings-package-details>
```
sadatay: i'm not the most proficient CSS person, but aren't inline styles generally frowned upon?  or are there certain cases where using them is best practice?

wwilsman: You're right, but there are already inline styles in this component, and so we came to the conclusion that most of the styling will need some polish anyway

sadatay: also, i know i've seen `em` used in other parts of the application.  is it kosher to mix-and-match `rem` and `em`, or is it better to uniformly use one of them?  i had to google them to understand what they actually meant, but if i'm understanding it correctly it sounds like we should consistently use one or the other

wwilsman: In my opinion, generally use `rem`, but `em` can be helpful when basing proportions on the current text size. Like padding for a button.

src/components/package-show.js
```diff
@@ -22,100 +20,96 @@ export default function PackageShow({ vendorPackage, packageTitles }, { intl })
...
             </div>
-          ) : vendorPackage.isRejected ? (
-            <p data-test-eholdings-package-details-error>
-              {vendorPackage.error.length ? vendorPackage.error[0].message : vendorPackage.error.message}
-            </p>
+          ) : packageTitles.isResolved ? (
+            <p>No Titles Found</p>
           ) : (
-            <Icon icon='spinner-ellipsis' />
+            <Icon icon="spinner-ellipsis" />
```
sadatay: how come we switched to double quotes here, but we're using single quotes on line 111?  i could've sworn ESLint enforced one or the other.  it's obviously not the end of the world if we aren't consistent in our quoting style, it's just puzzling

wwilsman: That is strange that eslint didn't pick this up. I usually go with double quotes for JSX fwiw.

src/components/package-show.js
cherewaty: Certainly should not be permanent.

sadatay:
i'm not going to pretend to fully understand this, and i think it might be a good idea to either document it or do a little knowledge sharing sesh.  although i don't always know what i'm looking at logically, the code itself seems pretty clean.  i spun the branch up locally and messed around with it and it seems to work correctly.  so that's good enough for me, and about as good of a review as i think i'm capable of giving.  

it's a +1 from me but probably should get some 👀  from someone with a better understanding as well.

src/components/package-show.js
```diff
@@ -22,100 +20,96 @@ export default function PackageShow({ vendorPackage, packageTitles }, { intl })
...
-                      />
-                    ))}
-                  </List>
-                </div>
-              ) : packageTitles.isResolved ? (
-                <p>No Titles Found</p>
-              ) : (
-                <Icon icon="spinner-ellipsis" />
-              )}
+          {packageTitles.isResolved && packageTitles.content.length ? (
```
cowboyd: I don't like the blippy-ness of the title flash in sometimes. It feels discombobulating. This isn't for this PR, but maybe we can have some sort of timeout, where we make the requests at the same time, and if they resolve within 1000ms of each other, we render them at the same time, otherwise we show the separate loading spninner.

src/components/package-show.js
cowboyd: Which is it?

src/routes/search.js
```diff
@@ -102,32 +117,49 @@ class SearchRoute extends Component {
...
+        search: queryString.stringify({
+          ...search[type].query,
+          searchType: type
+        })
       }
     }), {});
 
+    let Results = null;
+    if (searchType === 'vendors') {
+      Results = VendorSearchResultsRoute;
```
cowboyd: I don't think that any of these are actual routes any more, right? They're just pure function components. Probably better to put them as just that.

src/routes/search.js
```diff
@@ -136,9 +168,7 @@ class SearchRoute extends Component {
               onSearch={this.handleSearch}
             />
           )}
-        >
-          {this.props.children}
```
cowboyd: Why not use `SearchPaneset`'s children to pass embedded components instead of the `resultsView` attribute. Actually I think I see. Is it because there are two "slots" results and detail?

Incidentally, this is exactly what named blocks are for in Ember https://github.com/machty/rfcs/blob/named-blocks/text/0000-named-blocks.md

src/routes/search.js
```diff
@@ -101,46 +113,92 @@ class SearchRoute extends Component {
...
+    let { location, children } = this.props;
+    let { searchType, search } = this.state;
+
+    if (searchType) {
+      let hideDetails = /^\/eholdings\/?$/.test(location.pathname);
+      let { query } = search;
+
+      return (
+        <div data-test-eholdings>
+          <SearchPaneset
```
cowboyd: It might even make sense to delegate to a search view at this point. and put the paneset in that

wwilsman:
@cowboyd RE: Creeping asynchrony

I'm currently working on a component to abstract this logic out of all of these places

src/routes/search.js
wwilsman: That's what this is. The only thing that this route renders is the search paneset with the search form. I suppose we could just call it `searchView` and include the form inside of that component. But for this route that paneset is essentially the only view.

wwilsman:
I did a search for single quote properties and these loading spinners are the only components. I bet that one made it in and the rest were copy and paste. I amended them to be double quotes

src/routes/search.js
wwilsman: Yes, it is because there are two "slots" to render. The details and results are side-by-side in the UI and the logic for showing either or both is inside the `SearchPaneset` component

src/routes/search.js
wwilsman: I actually already did that in another branch I'm working on for the async abstraction. Moved those changes over to this PR. 👍

src/routes/search.js
cowboyd: The main reason I suggest it is that `renderResults` also feels like a component, or "view-y"

wwilsman: True. It used to be just a simple ternary in the render method, but I tried to pulled out most search-type specific things into their own methods (besides lines 97-99).

Essentially, it was this:
``` javascript
  resultsView={(isResolved || isPending || isRejected) && ResultsComponent}
```

It could also be something like `getResultsComponent` which would return the component constructor and the render method can give it props

wwilsman: I guess we could argue that this route is responsible for _multiple_ views. Search, Results, and Details.

PR #97, opened 2017-10-25
Bulletproof against missing contentType

cherewaty:
## Purpose
Sometimes a customer resource / package-title will come back from the RM API without a `contentType`. The implementation in https://github.com/thefrontside/ui-eholdings/pull/96 ignored that case, and we would get this error:

folio frontside io-eholdings-vendors-19-packages-2516-titles-480368 iphone 5

## Approach
Put in some checks against a blank or missing `contentType` for both packages and customer resources.

### DIFF
--- a/src/redux/utilities.js
+++ b/src/redux/utilities.js
@@ -10,5 +10,9 @@ export default function formatContentType(contentType) {
     onlinereference: 'Online Reference'
   };
 
-  return contentTypes[contentType.toLowerCase()];
+  if (contentType) {
+    return contentTypes[contentType.toLowerCase()];
+  } else {
+    return contentType;
+  }
 }
--- a/src/redux/utilities.js
+++ b/src/redux/utilities.js
@@ -10,8 +10,10 @@ export default function formatContentType(contentType) {
     onlinereference: 'Online Reference'
   };
 
-  if (contentType) {
-    return contentTypes[contentType.toLowerCase()];
+  let contentTypeKey = contentType.toLowerCase();
+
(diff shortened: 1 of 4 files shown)


src/redux/utilities.js
```diff
@@ -10,5 +10,9 @@ export default function formatContentType(contentType) {
     onlinereference: 'Online Reference'
   };
 
-  return contentTypes[contentType.toLowerCase()];
+  if (contentType) {
+    return contentTypes[contentType.toLowerCase()];
```
wwilsman: To further bulletproof this, we could also check if the `contentTypes` hash includes this content type as well. This way if there is ever a new content type that we don't know about, this function won't return `undefined` and will just return the unformatted content type.

cherewaty: :clap: Good idea. Added in new commit.

src/redux/utilities.js
```diff
@@ -10,8 +10,10 @@ export default function formatContentType(contentType) {
     onlinereference: 'Online Reference'
   };
 
-  if (contentType) {
-    return contentTypes[contentType.toLowerCase()];
+  let contentTypeKey = contentType.toLowerCase();
+
+  if (contentType && Object.prototype.hasOwnProperty.call(contentTypes, contentTypeKey)) {
```
wwilsman: Curious, why not just `contentTypes.hasOwnProperty(contentTypeKey)`?

cherewaty: We have this turned on https://eslint.org/docs/rules/no-prototype-builtins

wwilsman: Gotcha

PR #98, opened 2017-10-26
Update Publication Type display to be more user friendly 

carolegodfrey:
## Purpose
Publication Types which are returned from RMAPI are not well formatted for display. Often they do not contain appropriate spacing (for example - StreamingVideo vs Streaming Video).  Update display of  Publication Type to be more user friendly. (related Jira ticket = https://issues.folio.org/browse/UIEH-85)

## Approach
Following similar pattern as what was used for formatting content type for display. Apply  transform of publication type for display when deserializing server response in title and customer-resource.. 

#### TODOS and Open Questions
Serialization step should be moved to back end mod-kb-ebsco since values will be specific to the resource provider.

## Learning
Needed to look into RMPI reference https://developer.ebsco.com/rmapireference to understand specific payloads where pubType is present in server response.  This is a title level attribute

## Screenshots

image

image


src/redux/utilities.js
```diff
@@ -18,3 +18,31 @@ export default function formatContentType(contentType) {
...
+    database: 'Database',
+    journal: 'Journal',
+    newsletter: 'Newsletter',
+    newspaper: 'Newspaper',
+    proceedings: 'Proceedings',
+    report: 'Report',
+    streamingaudio: 'Streaming Audio',
+    streamingvideo: 'Streaming Video',
+    thesisdissertation: 'Thesis & Dissertation',
+    website: 'Web Site',
```
wwilsman: Isn't "Website" more grammatically correct? (without a space)

I googled it and "Web Site" can be _technically_ correct (that's how it was spelled in the 90's), but most places I've seen the word, it's without a space

src/redux/customer-resource.js
```diff
@@ -55,10 +58,15 @@ export const customerResourceEpics = combineEpics(
     deserialize: (payload) => {
       if (payload) {
         let { customerResourcesList, ...title } = payload;
+
         if (customerResourcesList[0] && customerResourcesList[0].contentType) {
           customerResourcesList[0].contentType = formatContentType(customerResourcesList[0].contentType);
         }
 
+        if (title.pubType) {
+          title.pubType = formatPublicationType(title.pubType);
```
cowboyd: This seems more like a view level concern to me. Is there any reason we wouldn't actually format the publication type not inside an epic or a reducer, but in the component that displays it?

(I realize that this is where `formatContentType`, but perhaps we should move both.

src/redux/package.js
```diff
@@ -5,7 +5,7 @@ import {
   createRequestReducer,
   createRequestEpic
 } from './request';
-import formatContentType from './utilities';
+import { formatContentType } from './utilities';
```
cowboyd: Same thing here. Seems like a view concern.

tests/customer-resource-show-test.js
```diff
@@ -49,7 +50,7 @@ describeApplication('CustomerResourceShow', () => {
     });
 
     it('displays the publication type', () => {
-      expect(ResourcePage.publicationType).to.equal(resource.title.pubType);
+      expect(ResourcePage.publicationType).to.equal('Streaming Video');
```
cowboyd: +1 for tests using hard coded values instead of variables. So much clearer what's it's testing and far less susceptible to false positives.

src/redux/customer-resource.js
wwilsman: I think the idea is that this will be a job of `mod-kb-ebsco` very soon

src/redux/utilities.js
carolegodfrey: Good point - I'll ask Khalilah about this - it is how the mapping is defined in the Jira ticket

src/redux/customer-resource.js
carolegodfrey: Oddly enough - We had started by making the changes for formatContentType in the view as you suggest as well. But the thought was that this will move to mod-kb-ebsco (as wil indicates) as the transformation rules/values will be specific to the resource management backend.   So the thought was to move it closer to the backend

cowboyd: Why not just make them in mod-kb-ebsco now?

wwilsman: It's just a passthrough for now, and we aren't handling any data yet. Once we start on the JSON-API conversion, this will be easy enough to plop in.

src/redux/package.js
cherewaty: Will move to mod-kb-ebsco very soon.

src/redux/utilities.js
carolegodfrey: @wwilsman - Ok, confimed with Khalilah - this is a carryover from older legacy _90's_ app and she is onboard with your suggestion.  Thanks for recommendation - will make the changes -http://www.stratecomm.com/faqs/website/

src/redux/customer-resource.js
cowboyd: OK, gotcha.

PR #99, opened 2017-10-26
Remove search type from search route state when the param is removed

wwilsman:
## Purpose
When you click a link that is missing the `searchType` parameter when the parameter was previously present, the search UI should disappear. It doesn't.

## Approach
When there is no `searchType` found during `componentDidReceiveProps`, still set the state to remove the existing search type

### DIFF
--- a/src/routes/search.js
+++ b/src/routes/search.js
@@ -61,7 +61,8 @@ class SearchRoute extends Component {
     let isNewQuery = props.location.search !== this.props.location.search;
     let isNewSearch = search !== this.state.search;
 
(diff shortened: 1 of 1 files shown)


src/routes/search.js
```diff
@@ -74,6 +75,9 @@ class SearchRoute extends Component {
       if (isNewSearch || isNewQuery) {
         this.setState({ query, searchType, search });
       }
+    } else if (this.state.searchType) {
```
cowboyd: I'm pretty sure this warrants a test.

PR #100, opened 2017-10-26
Add tests for search type removal

wwilsman:
## Purpose
PR #99 got merged a little quick before @cowboyd got his review in. We need to test the functionality of that PR to ensure no regressions happen in the future.

## Approach
- Adds a `goBack` helper which navigates backwards through our apps router history using the `TestHarness`'s `history` property.
- Creates additional mirage data during tests to click deeply inside of the preview pane on the various search routes.
- Tests that clicking deeply hides the search UI and navigating back shows it again.

### DIFF
--- a/tests/package-search-test.js
+++ b/tests/package-search-test.js
@@ -9,7 +9,7 @@ describeApplication('PackageSearch', () => {
   let pkgs;
 
   beforeEach(function () {
-    pkgs = this.server.createList('package', 3, 'withVendor', {
+    pkgs = this.server.createList('package', 3, 'withVendor', 'withTitles', {
       packageName: i => `Package${i + 1}`,
       titleCount: 3,
       selectedCount: 1
@@ -64,6 +64,30 @@ describeApplication('PackageSearch', () => {
 
         it('hides the preview pane', () => {
           expect(PackageSearchPage.previewPaneIsVisible).to.be.false;
+        });
+      });
+
+      describe('clicking an item within the preview pane', () => {
+        beforeEach(() => {
+          return PackageSearchPage.clickTitle(0);
+        });
+
+        it('hides the search ui', () => {
+          expect(PackageSearchPage.$root).to.not.exist;
+        });
+
+        describe('and going back', () => {
+          beforeEach(function () {
+            return this.goBack(() => expect(PackageSearchPage.$root).to.exist);
+          });
+
+          it('displays the original search', () => {
+            expect(PackageSearchPage.$searchField).to.have.value('Package');
+          });
+
+          it('displays the original search results', () => {
+            expect(PackageSearchPage.$searchResultsItems).to.have.lengthOf(3);
+          });
         });
       });
     });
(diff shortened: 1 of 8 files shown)


tests/package-search-test.js
```diff
@@ -54,6 +55,30 @@ describeApplication('PackageSearch', () => {
...
+          return PackageSearchPage.clickTitle(0);
+        });
+
+        it('hides the search ui', () => {
+          expect(PackageSearchPage.$root).to.not.exist;
+        });
+
+        describe('and going back', () => {
+          beforeEach(function () {
+            return this.goBack(() => expect(PackageSearchPage.$root).to.exist);
```
cowboyd: What do you think about putting `goBack` on the page? Or somehow getting the page object involved since it really owns the data about whether the goback succeeded. Somethng like `goBackTo(PackageSearchPage)` or something.

wwilsman: I did think about putting `goBack` on the page, but then we'd have to work in our testing context or the history object into the page object via a `beforeEach` or something.

Then I thought about how our visit helper is also not part of the page object even though for each visit we wait for a root element to exist. You could also argue moving all of our navigation helpers into the pages as well.

If we do want the two current navigation helpers to be part of the page objects, without involving the testing context, I think `goBackTo(PageObject)` and `visit(PageObject)` might work out. The implementation could be as simple as converging on the root element of the page object.

Although there might be times when we want to visit a page and wait for something besides the root to be loaded. We can chain our convergences, but I noticed that in practice the chained convergences would have to be aware of how long the previous convergence in the chain took (so it can throw right before the accumulated timeout).

Thoughts?

cowboyd: I like:

```javascript
goBackTo(PageObject.$root) and visit(PageObject.$root)
```

That way, we can change up where it goes pretty easily.

wwilsman: @cowboyd Just ran into a problem implementing this

`PageObject.$root` is a getter. So `visit(path, PageObject.$root)` will evaluate the root element as the argument (when it doesn't exist).

We could just make sure each page object has a `$root` and pass the entire page object into the helper instead. But this limits us if we need to wait for other things besides the root to be rendered. One example I can think of is when Jefferey and I were attempting to auto-load an initial empty search, we needed to wait on that initial search to be finished before the tests could run instead of the root element being rendered. 

However, it looks like we are _currently_ **not** doing anything other than asserting the root element exists. So changing this to always assert the `$root` of the passed `PageObject` exists will work now, just that it could hinder us in the future.

Which brings me back around to chaining our convergences. We could still always assert the root element exists, but chain another convergence from that:

``` js
return this.visit(path, PageObject).then(() => convergeOn(() => {
  expect(PageObject.searchResults).to.have.lengthOf(5);
}));
```

But this presents the problem of the second convergence not knowing when to properly bail due to the previous convergence taking `x` amount of time to converge.

**So**, do I continue to implement the helpers as taking the page objects as their argument (`visit(path, PageObject)` & `goBack(PageObject)`). Or leave them as-is for now until the bigtest-helpers come around to save us?

PR #102, opened 2017-10-31
Switch to simpler stripes setup with webpackOverrides

cherewaty:
Updates to `stripes-core` streamlining done here: https://github.com/folio-org/stripes-core/pull/85

### DIFF
--- a/demo/stripes.js
+++ b/demo/stripes.js
@@ -1,37 +1,13 @@
 #!/usr/bin/env node
 
 const commander = require('commander');
-const webpack = require('webpack');
-const HardSourceWebpackPlugin = require('hard-source-webpack-plugin');
 const path = require('path');
-const fs = require('fs');
-const nodeObjectHash = require('node-object-hash');
-const express = require('express');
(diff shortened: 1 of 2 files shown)


demo/stripes.js
```diff
@@ -1,47 +1,27 @@
...
-
 // Display webpack output to the console
 function processStats(err, stats) {
   if (err) {
     console.error(err); // eslint-disable-line no-console
   }
   console.log(stats.toString({ // eslint-disable-line no-console
     chunks: false,
-    colors: true
+    colors: true,
```
sadatay: this isn't an eslint issue?  thought we had no trailing commas enforced

cherewaty: Nope, we just have `comma-dangle` turned off completely, so our repo has no opinion on the matter. I copied and pasted this from `stripes-core`, which does require dangling commas and enforces it.

cowboyd: 🎵  Mr. Code Dangles.... 🎵

sadatay: 🆒  thanks for the explanation

PR #104, opened 2017-11-01
Switch to new JSON API vendor endpoints

cherewaty:
## Purpose
The `/eholdings/vendors` and `/eholdings/vendors/:id` endpoints are now available as JSON API documents at `/eholdings/jsonapi/vendors` and `/eholdings/jsonapi/vendors/:id`.

## Approach
There's a new mirage serializer only for vendors that extends `JSONAPISerializer`, but not the `ApplicationSerializer`.

#### TODOS and Open Questions
- [x] Merge https://github.com/thefrontside/mod-kb-ebsco/pull/17 first.
- [x] Test this implementation against the deployed endpoints.

### DIFF
--- a/src/components/vendor-search-results.js
+++ b/src/components/vendor-search-results.js
@@ -18,7 +18,7 @@ export default function VendorSearchResults({
     <Icon icon="spinner-ellipsis" />
   ) : isRejected ? (
     <p data-test-vendor-search-error-message>
-      {error.length ? error[0].message : error.message}
+      {error.errors.length ? error.errors[0].title : error.title}
     </p>
   ) : isResolved && !content.length ? (
     <p data-test-vendor-search-no-results>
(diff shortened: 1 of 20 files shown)


src/components/vendor-search-results.js
```diff
@@ -18,7 +18,7 @@ export default function VendorSearchResults({
     <Icon icon="spinner-ellipsis" />
   ) : isRejected ? (
     <p data-test-vendor-search-error-message>
-      {error.length ? error[0].message : error.message}
+      {error.errors.length ? error.errors[0].title : error.title}
```
cherewaty: This isn't great, but I figure we can redo how errors are handled in `createRequestReducer()` after we've switched over to 100% JSON API endpoints.

src/redux/search.js
```diff
@@ -45,18 +46,35 @@ function createSearchEpic(name, {
...
-      search: '',
-      count: 25,
-      offset: 1,
-      orderby: 'relevance',
-      ...defaultParams
-    }
-  });
+  // TODO: when switching package and title search over to JSON API,
+  // only use this version of return createRequestEpic()
+  if (name === 'vendors') {
```
cherewaty: I could've done this more elegantly, but I left it easier to strip out later.

src/components/vendor-search-results.js
wwilsman: Yea, I think we'll have to revisit our request helpers soon anyway (like default JSON serialization/deserialization and better support for updating resources)

PR #105, opened 2017-11-02
Add map operator for Observable

wwilsman:
## Purpose
The action$ in our epic inherits the Observable created in stripes-core. Any operators chained from this action$ must be defined in stripes-core. However when we create our own Observable using `from` or `of` we need to add the operators we expect to use in our own app

## Approach
- Removes `switchMap` as that is now defined in `stripes-core`
- Adds `map` since we need it when we use `Observable.from(request)`

### DIFF
--- a/src/redux/request.js
+++ b/src/redux/request.js
@@ -2,9 +2,9 @@ import queryString from 'query-string';
 import startsWith from 'lodash/startsWith';
 import { handleActions as reduxHandleActions } from 'redux-actions';
 import { Observable } from 'rxjs/Observable';
-import 'rxjs/add/operator/switchMap';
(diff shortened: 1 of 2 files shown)


wwilsman:
This needs our fork of `stripes-core` to be rebased with upstream

elrickvm:
I opened the PR on stripes-core to import switchMap. For some reason i couldn't requests review in that repo but once that is merge then we are good for this.

wwilsman:
@elrickvm actually Hongwei already made a commit to `stripes-core` you can close that PR now

elrickvm:
Yep @wwilsman good to go its closed.

PR #106, opened 2017-11-02
Add Manage Coverage Date display to Title Package Display

carolegodfrey:
## Purpose
Adds display of Managed Coverage Dates to Title Package Display (aka customer resource show page in code). 
Related Jira tickets are
- https://issues.folio.org/browse/UIEH-30 
- https://issues.folio.org/browse/UIEH-32. 
 
Of Note - 
- Managed Coverage Dates are read only. 
- Date format display is defined by publication type (book types display YYYY only, non-books are displayed as MM/DD/YYYY)
- There can be multiple date ranges returned from RMAPI
- Ranges appear comma separated and ordered by most recent coverage date.
- Illegally formed dates from RMAPI are not displayed

## Approach
- Added a coverage-dates component which encapsulates logic to display appropriate date formats
- Moved pub type formatting from redux level back to display level since needed to add logic for if a pub type is a book
- Used moment.js library to test for valid date formats
- Added mirage factory/model/serializer changes for ManagedCoverageList element in customer-resource

#### TODOS and Open Questions
- [ ] Still have open issue of adding div testing element to coverage-dates component. Currently its is set as 'data-test-eholdings-customer-resource-show-managed-coverage-list' but should be passed in as property
- [ ] Changes in mod-kb-ebsco for JSON API will change data model  

## Learning
- Used https://momentjs.com/ for checks for valid date format string.
- Needed to research/understand tasks for additional data model for mirage testing - http://www.ember-cli-mirage.com/docs/v0.3.x/

## Screenshots
managed-coverage
managed-coverage2


src/components/customer-resource-show.js
```diff
@@ -76,6 +78,16 @@ export default function CustomerResourceShow({ customerResource, toggleRequest,
                   </div>
                 </KeyValueLabel>
               ) }
+              {record.managedCoverageList && record.managedCoverageList.length > 0 && isValidCoverageList(record.managedCoverageList) && (
```
sadatay: `record.managedCoverageList.length > 0` can just be `record.managedCoverageList.length` since `0` is falsey

carolegodfrey: Great - thank for catching that!

src/components/customer-resource-show.js
```diff
@@ -8,8 +8,10 @@ import KeyValueLabel from './key-value-label';
 import IdentifiersList from './identifiers-list';
 import ContributorsList from './contributors-list';
 import ToggleSwitch from './toggle-switch';
+import CoverageDates from './coverage-dates';
+import { isBookPublicationType, formatPublicationType, isValidCoverageList } from './utilities';
 
-export default function CustomerResourceShow({ customerResource, toggleRequest, toggleSelected }) {
+export default function CustomerResourceShow({ customerResource, toggleRequest, toggleSelected }, { intl }) {
```
sadatay: why is `intl` separated here?

src/components/customer-resource-show.js
```diff
@@ -143,3 +156,6 @@ CustomerResourceShow.propTypes = {
   toggleRequest: PropTypes.object.isRequired,
   toggleSelected: PropTypes.func.isRequired
 };
+CustomerResourceShow.contextTypes = {
+  intl: PropTypes.object
```
sadatay: it looks like we're getting this from props in some places, context in others.  we should probably pick one or the other, but i dunno enough about the domain to know which is correct

src/redux/utilities.js
```diff
@@ -1,4 +1,4 @@
-export function formatContentType(contentType) {
+export default function formatContentType(contentType) {
```
sadatay: why did this get promoted to a default export?

carolegodfrey: Good catch - happened as a result of the rebase from master - this file was in conflict - checking in change to resolve

src/components/customer-resource-show.js
carolegodfrey: I was following the same usage pattern as existed with package-show.js. It looks like the separation is due to intl being a contextType vs a propType (as other parameters)

src/components/customer-resource-show.js
carolegodfrey: Could use help here as well, it seems contextType has benefits - but need to understand if this is proper usage

carolegodfrey: So researching more on propType vs contextType https://javascriptplayground.com/blog/2017/02/context-in-reactjs-applications/
Updated coverage-dates component to rec'v intl as context (instead of property). Is there somewhere else that looks inconsistent? or is there a best practice to use property instead of context?

src/components/utilities.js
```diff
@@ -0,0 +1,70 @@
...
+
+export function isValidCoverage(coverageObj) {
+  if (coverageObj.beginCoverage) {
+    if (!moment(coverageObj.beginCoverage, 'YYYY-MM-DD').isValid()) { return false; }
+  }
+  if (coverageObj.endCoverage) {
+    if (!moment(coverageObj.endCoverage, 'YYYY-MM-DD').isValid()) { return false; }
+  }
+  return true;
+}
```
cherewaty: I guess we sometimes get back improperly formatted dates from RM API? :(

If that's the case, I guess it's worth adding `moment`. Bummer.

src/components/coverage-dates/coverage-dates.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+CoverageDates.propTypes = {
+  coverageArray: PropTypes.array,
+  id: PropTypes.string,
+  isYearOnly: PropTypes.bool
+};
+
+CoverageDates.contextTypes = {
+  intl: PropTypes.object
```
cherewaty: I think this is probably the right way to deal with `intl` (on the `content`, not as a `prop`).

src/components/utilities.js
carolegodfrey: Yes - unfortunately we can get bad dates back from rmapi - confirmed this fact with Mohan

PR #108, opened 2017-11-06
Eliminate inline styles

cherewaty:
## Purpose
The few regrettable instances of inline styles were useful to get us to a quick demo one time, but let's pull them out into the css build.

## Future
Since I just repeated the same css several times here, it'd be useful to create a couple components to DRY it up, but that can happen as we get closer to a hard-baked `SearchPaneset`.

### Before
no-ragrets-temporary-tattoo

### After
161214416e59f9190930bf731dd2e04a--no-regrets-wrist-tattoo


cherewaty:
Waiting on `stripes-core` PR before I merge.

cherewaty:
I removed the dependency on a newer `stripes-core` version - that work will be in another PR.

PR #110, opened 2017-11-07
Correct manageCoverageList length check > 0 - avoid incorrect display of 0 length in ui

carolegodfrey:
## Purpose
Correct display issue found after release of Managed Coverage Dates on Title Package Display. Specifically a lone 0 appears below manage url in cases where a record has an empty managedCoverageList.

## Approach
Consistent with checks for other lists (such as subject list) 

## Learning
Even though length of ManagedCoverageList of 0 is falsey - because of the context of where it is used in - the JSX it results in the display of 0 in ui (as in below screenshot)

## Screenshots
managedcoveragelist-0-length

and below image shows the offending element in DOM

image

### DIFF
--- a/src/components/customer-resource-show/customer-resource-show.js
+++ b/src/components/customer-resource-show/customer-resource-show.js
@@ -77,7 +77,7 @@ export default function CustomerResourceShow({ customerResource, toggleRequest, 
             </KeyValueLabel>
           ) }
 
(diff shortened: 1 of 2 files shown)


cowboyd:
We should have a test case for this.

carolegodfrey:
Good point - This is an interesting problem to write a test case for - because the managed coverage list surrounding div with label didn't get generated

PR #112, opened 2017-11-13
Adjust to stripes.build promise

cherewaty:
## Purpose
`stripes.build()` and `stripes.serve()` now return promises.

## Side Effects
Because this upgrades `stripes-core`:
- Removes bootstrap global CSS; Source Sans Pro now default font
- `mod-notify` is now being used by `stripes-core`

#### TODOS and Open Questions
- [x] Install `mod-notify` on our Okapi cluster.
- [x] Mock the `mod-notify` endpoints in mirage.

### DIFF
--- a/demo/stripes.js
+++ b/demo/stripes.js
@@ -9,18 +9,23 @@ const fs = require('fs');
 
 commander.version(packageJSON.version);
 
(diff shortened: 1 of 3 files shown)


sadatay:
:shipit:

PR #116, opened 2017-11-16
Use newer stripes-core and stripes-components

cherewaty:
This feels a little gross, but it gets us the latest `stripes-components` without a new release of it.

I pointed the `stripes-components-fork-master` branch of `thefrontside/stripes-core` to `master` of `thefrontside/stripes-components` (instead of the versioned release).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   },
   "devDependencies": {
     "@folio/stripes-core": "thefrontside/stripes-core#master",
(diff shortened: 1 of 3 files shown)


wwilsman:
Should we also add `stripes-components` to our dependencies? I don't necessarily like relying on a package to be included from another package since we import directly from `stripes-components`

cherewaty:
Thanks @wwilsman. That's a much better option. Updated.

PR #117, opened 2017-11-16
Only use breakpoints defined in stripes-components

cherewaty:
Instead of defining our own breakpoints, we should stick to the ones defined in `stripes-components`.


cherewaty:
Future to-dos: set up style linting and figure out a better setup for importing the `stripes-components` variables.

PR #119, opened 2017-11-16
The Great Reducer Reduction

wwilsman:
## Purpose
Moves all resources to be consumed as JSON API and replaces our old redux request helpers (and the 15 reducers/epics) with a single data reducer and request epic.

Also, URLs more closely match the JSON API urls
```
/vendors/19/packages/1125 => /packages/19-1125
```

## Approach
In order to side-load related resources into our state, the previous reducers would have had to been able to influence reducers of other resource types. With the way the requests helpers worked, this would have been possible by adding a `requestHandler` for other requests that could possibly relate to it. But not only did it _look_ wrong, it would have gotten pretty messy, pretty quickly. Another problem with our request helpers was that they were _request_ helpers. So to update a resource, we had to create _yet another_ reducer-epic combo to go with the update request, and hook that up to the resource reducer to manipulate the original record.

This prompted us to look into a single reducer that would have control over all of the resource types. So if any resources were side-loaded, the reducer could simply send that data to the slice of the state responsible for holding onto those records. Having a single place for all records of a resource type also enables us to use data that has already been loaded rather than having to request the data any time we needed it (this makes the app _really_ snappy).

To facilitate all of this, a single reducer and epic were created to manage the state of our data. The reducer is responsible for manipulating the state (immutably of course). And the epic is now only responsible for transforming specific actions into real requests (and dispatching resolve and reject actions to the data reducer). 

To access a record by it's id would be as simple as `data.vendors[vendorId]`, however we still needed to know about the state of the request as well as the record. Plus, retrieving multiple records from a search would involve knowing which records the search request returned.

Enter the resolver. The resolver's job is to retrieve records for a specific _request_. This is important, as we don't know the ids of records returned from a search. But we do know what the search was for that made the request.

To return something meaningful from the resolver instead of a smashed state for the record, request, and any related records, models were introduced. Models allow us to access a record's related resources and even the request object that was created for the record which triggered the epic. Models also provide the ability for default attributes so we no longer need to check if a property exists before accessing it. 

There are some other nice things about the resolver and models that might make this PR longer than it already is, so I'll leave the rest to the Wiki.

#### TODOS
- [x] The backend needs to support including packages with vendors, and customer resources with packages and titles.
- [x] Additionally, the backend also needs to support editing resources via `PUT`

#### Future Concerns
- We're still repeating ourselves in a lot of places while checking if a resource is currently pending or not. This should be abstracted into a component that handles that stuff for us (along with displaying any errors returned from a request).
- While this is all very nice, it _does not_ allow us to make requests to any ol' API. The stuff going on in the reducer is specific to JSON API payloads. And the stuff going on in the request epic is specific to making requests via Okapi.

## Learning
[Redux Data and You (our wiki)](https://github.com/thefrontside/ui-eholdings/wiki/Redux-Data-and-You)


mirage/config.js
```diff
@@ -3,6 +3,7 @@ import { Response } from 'mirage-server';
 
 export default function configure() {
   this.urlPrefix = okapi.url;
+  this.timing = 1000;
```
sadatay: should this remain here?  seems like we don't want this in top-level config, no?

mirage/serializers/customer-resource.js
```diff
@@ -1,30 +1,28 @@
...
+  hash.attributes.packageId = customerResource.package.id;
+  hash.attributes.packageName = customerResource.package.name;
+  hash.attributes.contentType = customerResource.package.contentType;
+  hash.attributes.titleId = customerResource.title.id;
+  hash.attributes.name = customerResource.title.name;
+  hash.attributes.publisherName = customerResource.title.publisherName;
+  hash.attributes.publicationType = customerResource.title.publicationType;
+  hash.attributes.contributors = customerResource.title.contributors;
+  hash.attributes.identifiers = customerResource.title.identifiers;
+  return hash;
```
sadatay: this may no longer be necessary with the new customerResource objects being flattened

src/components/identifiers-list.js
```diff
@@ -3,26 +3,15 @@ import PropTypes from 'prop-types';
...
-    0: 'Empty',
-    1: 'Print',
-    2: 'Online'
-  };
-
   // get rid of identifiers we received that aren't ISSN or ISBN
-  let filteredData = data.filter(identifier => (identifier.type === 0 || identifier.type === 1) && (identifier.type >= 0 || identifier.type <= 2));
+  let filteredData = data.filter(identifier => ['ISSN', 'ISBN'].includes(identifier.type));
 
   // turn type and subtype ids into strings
```
sadatay: minor nit, but this comment isn't quite accurate anymore

mirage/config.js
wwilsman: You right. Slipped in by mistake

mirage/serializers/customer-resource.js
wwilsman: This is the flattening for mirage. Otherwise we'd have to define all of those properties twice and still make sure the customer-resource attributes equal the corresponding title attributes.

cherewaty:
`mod-kb-ebsco` still needs relationship endpoints like `eholdings/jsonapi/packages/COMPOSITE_PACKAGE_ID/customer-resources`

sadatay:
:+1: from me on the most recent commit.  some of the domain-specific knowledge is over my head, but the explanation was very clear and the code seems to be in line with it

cherewaty:
Successfully re-tested against production `mod-kb-ebsco`. :shipit:

PR #120, opened 2017-11-16
Set up stylelintrc

cherewaty:
## Purpose
CSS conventions should be just as consistent as our JS conventions, so let's lint it. https://github.com/stylelint/stylelint

## Approach
This sets up `.stylelintrc` for the repo and fixes the results of running `stylelint "src/**/*.css"`.

I installed a plugin in my editor (https://github.com/stylelint/stylelint/blob/master/docs/user-guide/complementary-tools.md#editor-plugins) to help stick to conventions in realtime.

## Next steps
- Make `stylelint` a part of the CI pipeline.
- Make `stylelint` violations show up with a webpack dev server.
- Make this setup a first-class citizen for the FOLIO ecosystem with `eslint-config-stripes`.

## Screenshots
Sample console output

screen shot 2017-11-16 at 5 09 03 pm

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+  extends: 'stylelint-config-standard'
+};
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
     "lint": "eslint ./"
   },
   "devDependencies": {
-    "@folio/stripes-core": "thefrontside/stripes-core#master",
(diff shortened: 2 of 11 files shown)


cherewaty:
Upon further investigation, `stylelint` isn't set up to elegantly create shared configs (like we did with `eslint-config-stripes`. Until that becomes a reality, we should just instrument FOLIO repos with `stylelint-config-standard`.

PR #121, opened 2017-11-20
Add Back button in eholdings UI

elrickvm:
## Purpose
Users need a way to step out / exit areas of the application and get back to another section as they navigate through the application. It is true that we have the back button in the web browser that will give user this functionality for free. However, as we have a response layout and user can be on varying devices the use of the native-button may not be a option. In addition having a back-button within the UI may feel more intuitive as Users are accustomed to clicking a `X` button to close things, `hamburger` icon to open hidden trays etc.

## Approach
Add a back button to:
* Package-details.
* Vendor-package-title aka customer-resource.
* Vendor-details.
* Title-details.

Create a `<Link />` Component that is used to track if a users location originated from within `eholdings`. That will allow us to render the back button if a User navigated from within `eholdings` but if they for example paste in the `url` to the browser the back-button will not be rendered as they did not originate from within `eholdings`.

## Screenshots

Back-button in action:

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -1,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Link } from 'react-router-dom';
+
+export default function InternalLink({
+  to,
+  ...props
+}) {
+  let location = { state: { eholdings: true } };
+
+  if (typeof to === 'string') {
+    location.pathname = to;
+  } else {
+    location = { ...to, ...location };
+  }
+
+  return (
+    <Link to={location} {...props} />
+  );
+}
+
+InternalLink.propTypes = {
+  to: PropTypes.oneOfType([PropTypes.string, PropTypes.object])
+};
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -3,6 +3,38 @@ import PropTypes from 'prop-types';
 
 import { Link } from 'react-router-dom';
 
+/**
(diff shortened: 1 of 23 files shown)


src/components/link.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Link } from 'react-router-dom';
+
+export default function InternalLink({
+  to,
+  ...props
+}) {
+  let location = { state: { eholdings: true } };
```
wwilsman: Maybe add a comment to this file indicating what we're doing and why

tests/customer-resource-show-test.js
```diff
@@ -111,17 +111,21 @@ describeApplication('CustomerResourceShow', () => {
...
-         * TODO: control timing directly with Mirage
-         */
+        * The expectations in the convergent `it` blocks
+        * get run once every 10ms.  We were seeing test flakiness
+        * when a toggle action dispatched and resolved before an
+        * expectation had the chance to run.  We sidestep this by
+        * temporarily increasing the mirage server's response time
+        * to 50ms.
+        * TODO: control timing directly with Mirage
+        */
```
wwilsman: Looks like your editor accidentally got to this whitespace. Maybe it's because the comment doesn't start with `/**` like other doc-blocks?

PR #122, opened 2017-11-21
Show Search Results Total

wwilsman:
## Purpose
Show the search results total for vendors, packages, and titles

## Approach
Copies how the total number is displayed for ui-users (under the results pane header).

## Screenshots
image

### DIFF
--- a/src/components/search-paneset/search-paneset.css
+++ b/src/components/search-paneset/search-paneset.css
@@ -20,6 +20,10 @@
   }
 }
 
+.search-heading {
+  text-align: center;
+}
+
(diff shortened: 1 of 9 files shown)


wwilsman:
This relies on #119 so we can easily pull the results total from the JSON API payload meta

wwilsman:
Ready for final review 👍

cherewaty:
I added a commit with a newer way to set pane header titles.

cherewaty:
Additional commits removed - I'll do that in another PR, since it involves some test changes.

PR #123, opened 2017-11-23
Show back button only if there is no query string

elrickvm:
## Purpose
Back button should not show if there is a queryString meaning that there was a search performed. There is already an exit button. The back-button is redundant and not useful in this context. A User in this context would just want to "close" this search detail pane completely and not step back.

## Problem
Prior we were using a queryString value that seemed to be only passed in through props on a Vendor and not Package or Title. So on a Package or Title search detail pane there would be a "X" and "<-" rendered in the UI.

## Approach
- Check the router to see if a search property exist.

### Before:

screen shot 2017-11-22 at 5 59 40 pm

### After:
screen shot 2017-11-22 at 7 05 38 pm

### DIFF
--- a/src/components/package-show/package-show.js
+++ b/src/components/package-show/package-show.js
@@ -17,11 +17,11 @@ export default function PackageShow({
   vendorPackage,
   packageTitles,
   toggleRequest,
-  toggleSelected,
-  queryString
+  toggleSelected
 }, { intl, router }) {
   let packageRecord = vendorPackage.content;
-  let historyState = router.history.location.state;
+  const historyState = router.history.location.state;
+  const queryString = router.history.location.search;
 
   return (
     <div>
@@ -142,7 +142,6 @@ PackageShow.propTypes = {
   packageTitles: PropTypes.object.isRequired,
   toggleRequest: PropTypes.object.isRequired,
   toggleSelected: PropTypes.func.isRequired,
-  queryString: PropTypes.string
 };
 
 PackageShow.contextTypes = {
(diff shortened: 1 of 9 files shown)


cherewaty:
Is there a test we could add that would've failed before this fix?

src/components/package-show/package-show.js
```diff
@@ -17,11 +17,11 @@ export default function PackageShow({
...
   packageTitles,
   toggleRequest,
-  toggleSelected,
-  queryString
+  toggleSelected
 }, { intl, router }) {
   let packageRecord = vendorPackage.content;
-  let historyState = router.history.location.state;
+  const historyState = router.history.location.state;
+  const queryString = router.history.location.search;
```
wwilsman: This won’t work when we get into paginating related resources because `location.search` is the entire query param string from the url. 

We’d have to parse it with `query-string` to get the actual `?search` param. 

Since the search route is always rendered, and always parses the params anyway, maybe we could alter the router context to include the parsed params. That would also be helpful for pagination later on.

elrickvm: Is the router context different to `router.route.location.search`?

wwilsman: Exactly. We basically want to know if we're in a 3-pane layout or not. And that is currently determined by the presence of `searchType` in the search route. Other parts of the query string may eventually relate to paginating packages within the vendor detail pane for example, regardless of whether or not we're in a 3-pane layout.

So if we're on a single-pane detail view on page 2 of the related resources, the `location.search` will still be `?page=2` or what have you, but we still want to show the back button.

elrickvm:
Added test coverage in the places where the button leaked into areas where it was not suppose to..

PR #125, opened 2017-12-05
Add query params context to children of the search route

wwilsman:
## Purpose
Eventually, we will need to parse query params in most of our routes so we can paginate and filter related resources on the various detail pages. Since our search route already parses query params, and is always the parent of these other routes, we can simply provide the query params to child components via a child context defined within the search route.

## Approach
Simply define and provide the query params child context within the search route.

Additionally, since the back button should only be visible outside of the three-pane search layout, the `route.location.search` check has been replaced with `queryParams.searchType`. This way we can retain our query params but still show the back button in the absence of a `searchType`.

#### Concerns
- Context in React is _technically_ experimental. But considering we are using it anyway for `router` and `intl`, and other React libraries we rely on use it as well, I doubt it's going anywhere soon.

## Learning
[React context docs](https://reactjs.org/docs/context.html)


wwilsman:
This relies on #119, just because that should be getting merged soon and future work on pagination and filtering will also be based on that branch. Relevant commit for this PR is 87ab466

PR #126, opened 2017-12-06
Update to stripes-connect release

cherewaty:
## Purpose
Something wonky happened with today's `stripes-connect` release where previous commit hashes disappeared.

When running `yarn`:
screen shot 2017-12-05 at 3 53 44 pm

## Approach
It's been fixed in `stripes-core`, but upgrading to the latest there caused some test failures that need more investigation. Instead we can just request the explicit release version in `ui-eholdings`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   },
   "devDependencies": {
     "@folio/stripes-components": "thefrontside/stripes-components#master",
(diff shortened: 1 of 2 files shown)


cherewaty:
Replaced by https://github.com/thefrontside/ui-eholdings/pull/127

PR #127, opened 2017-12-06
Upgrade to latest stripes-core and stripes-components

cherewaty:
## Purpose
Context of why we want to upgrade here: https://github.com/thefrontside/ui-eholdings/pull/126

There's a new reset stylesheet in `stripes-components`, along with CSS updates for buttons, etc.

## Approach
- The front component styles needed a few updates to look ok again.
- Search result lists look a little different, but I didn't make any changes there yet.

## Screenshots
localhost_3000_ iphone 5
localhost_3000_eholdings_searchtype vendors search e iphone 5

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -25,6 +25,8 @@ webpackConfig.module.rules = webpackConfig.module.rules.filter((rule) => {
   return !rule.use || !rule.use.some(use => use.loader === '@folio/stripes-config');
 });
 
+webpackConfig.plugins = [];
+
 // make sure that the NODE_ENV is available in browser code.
 webpackConfig.plugins.push(new webpack.EnvironmentPlugin({
   NODE_ENV: 'test'
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "license": "MIT",
   "scripts": {
     "postinstall": "bin/link-self",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
   "license": "MIT",
   "scripts": {
     "postinstall": "bin/link-self",
-    "start": "node demo/stripes.js dev demo/stripes.config.js",
+    "start": "NODE_ENV=development node demo/stripes.js dev demo/stripes.config.js",
```
cherewaty: Without this change:
screen shot 2017-12-05 at 5 25 46 pm

karma.conf.js
```diff
@@ -25,6 +25,8 @@ webpackConfig.module.rules = webpackConfig.module.rules.filter((rule) => {
   return !rule.use || !rule.use.some(use => use.loader === '@folio/stripes-config');
 });
 
+webpackConfig.plugins = [];
```
cherewaty: https://github.com/folio-org/stripes-core/pull/120 caused our test run to load in multiple `webpack.EnvironmentPlugin`s. The result was a full run of the tests went from about ~1min to ~5min. Clearing out the plugins and starting from scratch resolved the slowdown.

PR #131, opened 2017-12-20
Remove /jsonapi namespace

cherewaty:
Now that we're exclusively using the new `mod-kb-ebsco` JSON API endpoints, we no longer need them sandboxed under `/jsonapi/`.


cherewaty:
Should merge simultaneously with https://github.com/thefrontside/mod-kb-ebsco/pull/46

PR #133, opened 2017-12-21
Search Pagination

wwilsman:
## Purpose
Adds pagination support to the search pages via an infinite-scroll implementation.

### Paginating Data
Pagination is now handled in the `Collection` returned from our resolver. The collections still provide a `map` function to map over the records in a typical way, but now also have `getPage` and `getRecord`.

`getRecord` returns the record at a given offset, or an empty record for the collection type (a collection of `packages` will return an empty `package` model).

`getPage` returns the request and model records for a given page number for the set of records.

The collection determines where to find the other records by using the exact same params used when resolving the collection to resolve other requests for different page numbers.

``` javascript
let collection = resolver.query('packages', { q: 'bio' });
let { request, records } = collection.getPage(2);
// request.params => { q: 'bio', page: 2 }
```

### Pagination Components
Three new components have been added to handle pagination

- `Impagination` - Uses [impagination](https://github.com/flexyford/impagination) to create a dataset that can be mapped over. When the component receives a new `readOffset`, it will update the read offset of it's dataset and in turn call `fetch` when necessary to fetch other pages of data. This component uses `getPage` from our collection to resolve page promises that the dataset requires.

- `ScrollView` - Wraps all of it's children in `<li>` elements that are fixed height and positioned absolutely so that only the in-viewport elements are actually rendered in the DOM. This component calls an `onUpdate` prop whenever the read offset of the list has changed.

- `QueryList` - Combines the two above components to create an infinite-scroll pagination component and shows any error messages encountered as well. This is an abstraction of the three separate `*SearchResults` components we currently have.

### Other Changes
The `*SearchResults` components still exist to provide type-specific properties to the `QueryList` component. However, they have been renamed `*SearchList` to reflect that they aren't simply rendering results, but are actually rendering an infinite-scroll list.

The three `*ListItem` components have also been changed so that they do not render `<li>` elements. This is so `ScrollView` can render and position them instead. Since these components no longer render them, `<li>` elements were added to the detail views to compensate.

### Tests
- Found a false positive with searching for a second time. Searching twice worked, but initial changes I made caused it not to work, but those tests were still passing. Fixed now.
- Added a mirage helper to create search routes so we aren't repeating ourselves too much. The idea being we can also utilize this helper for the nested relationship routes as well as the three top-level search routes.
- Added tests for paginating, but since the sparse renderer doesn't render all of it's list elements, those tests simply check `$searchResultsItems[4]` as the first element in the list. The index `4` is because the sparse renderer has a threshold of +/- 5.

## Learning
- [Impagination](https://github.com/flexyford/impagination)


src/components/query-list/query-list.js
```diff
@@ -0,0 +1,110 @@
...
+            <div className={styles.error} data-test-query-list-error={type}>
+              {state.rejected[0].error[0].title}
+            </div>
+          ) : !state.length ? (
+            <div className={cx('error', 'not-found')} data-test-query-list-not-found={type}>
+              {notFoundMessage}
+            </div>
+          ) : (
+            <ScrollView
+              length={listLength}
```
cowboyd: We should be able to use `state.length` here. If we can't then maybe it's a bug in impagination?

src/components/query-list/query-list.js
```diff
@@ -0,0 +1,110 @@
...
+    notFoundMessage: 'Not Found'
+  }
+
+  state = {
+    readOffset: (this.props.page - 1) * this.props.pageSize
+  };
+
+  handleUpdate = (readOffset) => {
+    let { pageSize, onPage } = this.props;
+    let page = Math.floor(readOffset / pageSize) + 1;
```
cowboyd: If we track the offset in the url, and not the page, then we don't have to do this math.

src/components/query-list/query-list.js
```diff
@@ -0,0 +1,110 @@
...
+      fetch,
+      itemHeight,
+      renderItem
+    } = this.props;
+    let {
+      readOffset
+    } = this.state;
+
+    let listLength = collection.isLoaded ? collection.length : page * pageSize;
+    let totalPages = Math.ceil(listLength / pageSize);
```
cowboyd: If possible, I'd like to push all of this down into the state that is managed by the `Impagination` component, because things like length and total pages really are properties of the dataset. And should be inferrable. If we can't push them down, then we need to ask ourselves how we can improve impagination so that we can.

src/components/scroll-view/scroll-view.js
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    // calculate the positions before filtering
+    let items = Children.map(children, (child, i) => (
+      <li key={i} className={styles['list-item']} style={calcStyle(i)}>
+        {child}
+      </li>
+    ));
+
+    return items.filter((child, i) => {
```
cowboyd: A filter here is still going to have to iterate over each item, right? So if I'm reading this right, then a list with 10,000 items will need to run this filter on every item in the dataset.... on every re-render.

Alternatively, we can map back from `scrollIndex` to a range of items, `[n...m]` and then take that `slice(n,m)` of the dataset state, and then map _those_ records.

src/components/impagination.js
```diff
@@ -0,0 +1,149 @@
...
+  // set the initial read offset of our dataset
+  componentWillMount() {
+    let { readOffset } = this.props;
+    this.dataset.setReadOffset(readOffset);
+  }
+
+  // when we recieve an update, loop over the pages of collections so
+  // we can resolve or reject any pending requests
+  componentWillReceiveProps(nextProps) {
+    let { totalPages, collection } = nextProps;
```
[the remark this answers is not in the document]
wwilsman: It is coming from a calculation in the `QueryList` component, but you're right, it should probably be an attribute on the Collection.

src/components/impagination.js
```diff
@@ -0,0 +1,149 @@
...
+
+    // new, zero length collections are likely entirely new
+    let isNewCollection = !collection.length &&
+      collection.isLoading !== this.props.collection.isLoading;
+
+    // we need to reset the dataset
+    if (isNewCollection) {
```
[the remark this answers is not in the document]
wwilsman: Referential equality is not good enough because whenever any part of our data store changes, a new resolver is created, which in turn creates new Models and Collections. So when any request is resolved, that creates a new Collection.

I did throw away the old dataset before, but then I found `reset()` and though it was a good use here. I'll go back to creating a new dataset.

src/components/impagination.js
```diff
@@ -0,0 +1,149 @@
...
+
+    // ensure that we keep all of our promises
+    for (let page of Object.keys(this.promises)) {
+      let promise = this.promises[page];
+      let { request, records } = collection.getPage(page);
+
+      if (promise) {
+        if (request.isResolved) {
+          promise.resolve(records);
+          delete this.promises[page];
```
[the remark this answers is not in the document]
wwilsman: We don't want to delete it if it is pending or not requested yet at all. So I'd have to add:

```javascript
if (request.isResolved || request.isRejected) {
  delete this.promises[page];
}
```

So we either duplicate the `delete` or duplicate the `if` logic.

src/components/impagination.js
```diff
@@ -0,0 +1,149 @@
...
+
+        // the collection is not resolved or rejected already, make
+        // the request now and handle resolution later on in the
+        // componentWillReceiveProps hook
+      } else {
+        this.props.fetch(page, pageSize);
+
+        // cache this promise's callbacks (and status) so it can be
+        // resolved or rejected later on
+        this.promises[page] = { resolve, reject };
```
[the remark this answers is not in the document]
wwilsman: I can't seem to find it now, but I thought I recalled you mentioning something like `dataset.resolvePage(page, records)` and `dataset.rejectPage(page, error)` or something along those lines. Which would remove the need for our promises. In the case where we go with and API like this, would `fetch` then just return undefined?

src/components/query-list/query-list.js
wwilsman: I'm all for tracking the offset in the url and not the page, but would we have to change it back to page if other people think having `page` in the URL is more useful than `readOffset`? Maybe we should keep this for now and open an issue where other Ebsco folks could chime in.

src/components/scroll-view/scroll-view.js
wwilsman: `slice` definitely makes sense. The logic in `filter` was left over from when I was doing so in the `dataset.state.map()` function.

src/components/impagination.js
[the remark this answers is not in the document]
cowboyd: Yeah, sounds like we need to really think this out. Let's leave it as is, and schedule a retrospective once this gets merged so that we can suggest improvements to impagination.

src/components/query-list/query-list.js
cowboyd: The only reason they might prefer `page` in the url is the comfort that comes with familiarity. In terms of linkability though, `offset` is an order of magnitude more precise because if we're using page as our cursor we can be off by as several viewports worth of records depending on your page size. That's a degraded UX.

My take: we're already demonstrating a better way, and providing a much enhanced experience, so we should push forward with it.

src/components/scroll-view/scroll-view.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  // clean up our resize listener
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.handleListLayout);
+  }
+
+  // Handles setting the read offset based on the current scroll
+  // position. Also calls `onPage` when a new page has crossed the
+  // threshold of what's considered to be active the view
```
sadatay: looks like a minor grammar error here

src/components/title-search-list.js
```diff
@@ -0,0 +1,43 @@
...
+  onPage
+}) {
+  return (
+    <QueryList
+      type="titles"
+      page={parseInt(params.page || 1, 10)}
+      fetch={fetch}
+      collection={collection}
+      onPage={onPage}
+      itemHeight={80}
```
sadatay: magic number?

src/routes/search.js
```diff
@@ -171,21 +109,87 @@ class SearchRoute extends Component {
...
 
   /**
-   * Renders the search results component specific to the current search type
+   * Update the url to match new search params
+   * @param {Object} params - query param object
+   */
+  updateURLParams(params) {
+    let { location, history } = this.props;
+
+    // if the new query is diffent from our location, update the location
```
sadatay: `different`

tests/pages/vendor-search.js
```diff
@@ -93,5 +93,14 @@ export default {
...
     return $('[data-test-eholdings-package-details-back-button]').trigger('click');
+  },
+
+  scrollToOffset(readOffset) {
+    let $list = $('[data-test-query-list="vendors"]').get(0);
+    let rowHeight = $('li', $list).get(0).offsetHeight;
+    let scrollOffset = rowHeight * readOffset;
+
+    $list.scrollTop = scrollOffset;
+    $list.dispatchEvent(new Event('scroll'));
```
sadatay: not worth pulling out just for this, but maybe down the line we could have a shared helpers file for page objects

tests/unit/redux-model-test.js
```diff
@@ -135,21 +125,25 @@ describe('Redux Models', () => {
...
+      attributes: { totalPackages: 1 },
+      relationships: { packages: {} }
+    };
 
-      vendor = new VendorModel({
-        ...VENDOR_JSON,
-        attributes: { totalPackages: 1 },
-        relationships: { packages: {} }
-      }, vendorRequest, resolver);
+    let INCOMPLETE_PACKAGE_JSON = {
```
sadatay: why `let` and then CONSTANT_CASE?

sadatay:
i definitely didn't fully grok this, but on the first pass through the code looks tidy and the comments were extremely helpful

tests/unit/redux-model-test.js
wwilsman: I'm already trained to only use `const` only in the outer scope, so I accidentally `let` these in.

tests/pages/vendor-search.js
wwilsman: Hopefully that's where bigtest can come help :D

src/components/title-search-list.js
wwilsman: Yea, we could wait until mount and get the height of an individual item. But if that one item we choose to measure gets a weird layout, that would result in all of the offset and scroll calculations being off.

I saw one scrolling component that didn't require an `itemHeight` prop, but it also didn't provide an `onScroll` or an `onUpdate` method so we can get the `offset`. So to calculate that we'd have to have an `itemHeight` in that case anyway.

src/components/scroll-view/scroll-view.js
```diff
@@ -105,8 +111,17 @@ export default class ScrollView extends Component {
...
 
-    // slice the visible items and map them to `children`
-    return items.slice(lower, upper + 1).map((item, i) => {
+    // slice the visible items
+    let visible = items.slice(lower, upper + 1);
+
+    // empty arrays can still have lengths
+    if (!Object.keys(visible).length) {
+      // in case of an empty array, fill it with null values
+      visible = visible.fill(null);
```
cowboyd: This looks like a bug in impagination. If the slice isn't there, the result should be an empty array: 
```js
[].slice(5,100) //=> []
```

wwilsman: I believe this _is_ a bug in impagination. The data being sliced is there, but slice still sometimes returns an empty array. 

From what I initially found, `state.length` is correct, but it only has getters for `[0-99]` and `[100-199]` even though the read offset was `> 4000`.

`state.getRecord(offset)` returns the correct record, hence this workaround.

PR #147, opened 2018-01-06
Interrupt package deselection with confirmation modal

cherewaty:
## Purpose
If a user wants to deselect a package, ask them in a confirmation modal whether they actually want to commit that destructive action.

Part of https://issues.folio.org/browse/UIEH-34

## Approach
The key to making this feature work was keeping the state of the selection toggle independent of the model state. With a toggle component, it would've been strange if you clicked it, the toggle didn't move, but you received a confirmation modal. The UI feels more natural when the toggle always moves with a click/touch, but reverts back if the user decides to cancel the action or the server runs into a problem.

The current `stripes-components` modal was incompatible with `ui-eholdings`. I copy-and-pasted it over and made some necessary modifications. That's not sustainable, but requires more thought and care at the FOLIO project-wide level. What needs to be addressed with the `stripes-components` modal:
  1. there's no z-index, so it ends up layered behind the responsive search panes; there's no comprehensive strategy for managing z-indices in FOLIO
  2. there's no footer space defined for buttons; not a dealbreaker like number 1, but inconvenient
  3. mobile-unfriendly
  4. visual styles are less refined than areas of FOLIO that have received some TLC

## Screenshots
rsasj71n7a

### Modal at larger sizes
screen shot 2018-01-06 at 5 43 04 pm

### DIFF
--- a/src/components/modal/modal.js
+++ b/src/components/modal/modal.js
@@ -1,0 +1,126 @@
+/*
+  Forked from stripes-components due to z-index incompatibility.
+  Should revert back to upstream when fixed.
+*/
+
+import React from 'react';
+import { Modal as OverlayModal } from 'react-overlays';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import Icon from '@folio/stripes-components/lib/Icon';
+import css from './modal.css';
+
+const propTypes = {
+  /** Deciding value for rendering the modal(true) or not(false). */
+  open: PropTypes.bool.isRequired,
+  /** Callback that signals intent to close window -
+   * boolean passed to 'open' prop can be set using
+   * callback does not actually close the modal.
+   */
+  onClose: PropTypes.func,
+  /** Callback triggered when modal opens */
+  onOpen: PropTypes.func,
+  /** Size defaults to 60% the scope element's width.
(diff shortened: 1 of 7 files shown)


src/components/package-show/package-show.js
```diff
@@ -1,137 +1,199 @@
...
+  };
+
+  commitSelectionToggle = () => {
+    this.setState({ showSelectionModal: false });
+    this.props.toggleSelected();
+  };
+
+  cancelSelectionToggle = () => {
+    this.setState({ showSelectionModal: false });
+    this.setState({ packageSelected: this.props.model.isSelected });
```
wwilsman: These two `setState` should be combined into one:

```javascript
this.setState({
  showSelectionModal: false,
  packageSelected: this.props.model.isSelected
});
```

cherewaty: Fixed.

src/components/modal/modal.js
```diff
@@ -0,0 +1,121 @@
+import React from 'react';
```
wwilsman: Should we leave a comment in this file outlining that is was copied and why?

cherewaty: Yes. Added.

PR #150, opened 2018-01-08
Show loading indicator for loading the total results count

wwilsman:
## Purpose
Display "Loading..." message in place of the total results count when loading search results.

## Approach
When a collection does not have any loaded records, show the "Loading..." message.

If we used the collection's `isLoading` property, that would return `true` when the requests for the _current page_ in the collection is pending.

The `isLoaded` property (renamed to `hasLoaded`) returns `true` when _any page's_ request for the collection are no longer pending (resolved or rejected).

The previous method for finding the first resolved request in a collection stopped looking for resolved pages after the current page. So if `page=5` is loaded and we view `page=1`, `hasLoaded` would be `false` even though `page=5` has indeed already loaded. 

This method has been updated to keep searching for resolved requests for up to 400 pages. The reason being that we cannot use `this.totalPages` as it uses `this.length` which in turn calls the getter for `request`, thereby entering an infinite loop. This page limit is arbitrary and was only chosen to ensure all pages are accounted for with up to 10,000 total results (with the default 25 record page size). The loop will exit and cache the request once a resolved request has been found.

## Tests
Since `isLoaded` returned `false` when viewing previous pages in the collection, when we scrolled up in a results list it would flash back to "Loading..." even though we already had the total search count. Additionally, the default `50ms` timeout for `it.still` did not catch this scenario since the following request happened after we scrolled and stopped for at least `300ms`. The tests added to account for this include setting the `it.still` timeout to `500ms`.

## Screenshots
loading

### DIFF
--- a/src/components/search-paneset/search-paneset.js
+++ b/src/components/search-paneset/search-paneset.js
@@ -21,6 +21,7 @@ export default class SearchPaneset extends React.Component {
     resultsView: PropTypes.node,
     detailsView: PropTypes.node,
     totalResults: PropTypes.number,
+    isLoading: PropTypes.bool,
     location: PropTypes.object
   };
 
@@ -67,7 +68,8 @@ export default class SearchPaneset extends React.Component {
       resultsType,
       resultsView,
       detailsView,
-      totalResults
+      totalResults,
+      isLoading
     } = this.props;
     let { intl } = this.context;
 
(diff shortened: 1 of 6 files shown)


wwilsman:
Weird that that gif doesn't play. 🤷‍♂️

PR #151, opened 2018-01-08
Editable custom coverage date on package

elrickvm:
## Purpose
A user needs to be able to perform CRUD (create, read, update, and delete) operations on custom coverage dates on a package.

## Approach
Create a custom-coverage-date component that encapsulates those CRUD operations and incorporate those actions into the application. Forms have a lot of state around them state that is 'internal' or a concern to the Form itself and state that is 'external' and a concern to the application and Redux-Store. To aid with this integration of 'internal', 'external' state, crud operations, dispatched actions to store and connecting and communicating with store and the like we employed the use of Redux-Form.

The validation around the custom-coverage-date component complies with https://issues.folio.org/browse/UIEH-50.

## Learning
[Form Design](https://www.lukew.com/resources/web_form_design.asp)
[Redux-Form](https://redux-form.com/7.2.0/)

### DIFF
--- a/src/components/custom-coverage-date/custom-coverage-date.css
+++ b/src/components/custom-coverage-date/custom-coverage-date.css
@@ -11,7 +11,6 @@
 
 .custom-coverage-action-buttons {
   display: flex;
-  flex-direction: row-reverse;
   margin-top: 1em;
 }
 
--- a/src/components/custom-coverage-date/custom-coverage-date.css
+++ b/src/components/custom-coverage-date/custom-coverage-date.css
@@ -1,0 +1,42 @@
+@import '@folio/stripes-components/lib/variables';
+
+.custom-coverage-form,
+.custom-coverage-form-editing {
+  padding: 0.5em;
+}
+
+.custom-coverage-form-editing {
+  background: #e6f3ff;
+}
+
+.custom-coverage-action-buttons {
+  display: flex;
+  flex-direction: row-reverse;
+  margin-top: 1em;
+}
+
+.custom-coverage-date-display {
+  display: flex;
+  align-items: center
+}
+
+.custom-coverage-dates {
+  display: flex;
+  align-items: flex-start;
+}
+
+.custom-coverage-date-clear-row {
+  flex: 0 0 3em;
+  margin-top: 1.5em;
+  text-align: center;
+}
+
+.custom-coverage-date-picker {
+  align-self: flex-start;
+  flex: 1 1 auto;
+  margin-right: 1em;
+}
+
+.custom-coverage-add-button {
+  margin: 1em 0;
+}
--- a/src/components/custom-coverage-date/custom-coverage-date.js
+++ b/src/components/custom-coverage-date/custom-coverage-date.js
@@ -1,0 +1,179 @@
+import React, { Component } from 'react';
+import { Field, reduxForm } from 'redux-form';
+import PropTypes from 'prop-types';
+import moment from 'moment';
+
+import Datepicker from '@folio/stripes-components/lib/Datepicker';
+import IconButton from '@folio/stripes-components/lib/IconButton';
+import Button from '@folio/stripes-components/lib/Button';
+import styles from './custom-coverage-date.css';
+
+import { formatISODateWithoutTime } from '../utilities';
+
+class CustomCoverageDate extends Component {
+  static propTypes = {
+    initialValues: PropTypes.shape({
+      beginCoverage: PropTypes.string,
+      endCoverage: PropTypes.string
+    }).isRequired,
+    handleSubmit: PropTypes.func.isRequired,
+    onSubmit: PropTypes.func.isRequired,
+    pristine: PropTypes.bool,
+    initialize: PropTypes.func,
+    isPending: PropTypes.bool
+  }
+
+  static contextTypes = {
+    intl: PropTypes.object
+  }
+
+  state = {
+    isEditing: false
+  }
+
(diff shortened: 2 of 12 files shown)


src/components/custom-coverage-date/custom-coverage-date.css
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+.custom-coverage-action-buttons {
+  display: flex;
+}
+
+.custom-coverage-date-display {
+  display: flex;
+  // align-items: center;
```
cherewaty: Minor nit: remove comment

src/components/custom-coverage-date/custom-coverage-date.js
```diff
@@ -0,0 +1,170 @@
...
+import Button from '@folio/stripes-components/lib/Button';
+import styles from './custom-coverage-date.css';
+
+class CustomCoverageDate extends Component {
+  static propTypes = {
+    initialValues: PropTypes.shape({
+      beginCoverage: PropTypes.string,
+      endCoverage: PropTypes.string
+    }).isRequired,
+    handleSubmit: PropTypes.func.isRequired,
```
cherewaty: I only see two props passed into `CustomCoverageDate` at https://github.com/folio-org/ui-eholdings/pull/151/files#diff-0d953e650d58ec07823a4b3dc2a91366R8; do these other things need to be props?

elrickvm: These are passed by Redux form implicitied in the background

src/components/custom-coverage-date/custom-coverage-date.js
```diff
@@ -0,0 +1,169 @@
...
+          >
+            Add Custom Coverage
+          </Button>
+        </div>
+      );
+    }
+  }
+}
+
+function validate(values, props) {
```
sadatay: minor nit, but i'd move this above the `render`.  i'd also throw some comments throughout this file, as `redux-form` is not the most obvious thing in the world.

src/components/package-show/package-show.js
```diff
@@ -13,13 +13,15 @@ import List from '../list';
...
+import PackageCustomCoverage from '../package-custom-coverage/';
 import { formatISODateWithoutTime } from '../utilities';
 import styles from './package-show.css';
 
 export default class PackageShow extends Component {
   static propTypes = {
     model: PropTypes.object.isRequired,
-    toggleSelected: PropTypes.func.isRequired
+    toggleSelected: PropTypes.func.isRequired,
+    customCoverageSubmitted: PropTypes.func.isRequired
```
sadatay: as we add more of these maybe we should bundle the field-specific edits into a single PUT with only the changed info.  the backend is smart enough to collapse nil values into the model's pre-existing values.  i dunno how this jives with the redux data layer though

src/components/custom-coverage-date/custom-coverage-date.js
elrickvm: I think it can be moved. it will have to be passed as { validate: this.validate } which may work. And then it can put into the class above render.

elrickvm: How will we reference the validation function from within the class from the redux-form connect method at the end of the file?

src/components/package-show/package-show.js
elrickvm: What do you mean by field specific edits?

sadatay: like rather than have `toggleSelected` and `customCoverageSubmitted` maybe just `packageEdited`.  collect the attributes that have been touched, construct the payload and fire it off to the server.  now that i say it out loud i'm not as sure about it.  i'm just thinking long term all these edits are doing the same thing - sending the full payload to the server with one thing changed and the other things nullified.  could be an opportunity here to DRY it up, but now i'm not so sure that DRYing wouldn't be at the cost of readability

tests/package-show-custom-coverage-test.js
```diff
@@ -37,14 +57,60 @@ describeApplication('PackageShowCustomCoverage', () => {
     it('displays the custom coverage section', () => {
       expect(PackageShowPage.customCoverage).to.equal('7/16/1969 - 12/19/1972');
     });
+
+    // TODO: why use `it.still` here instead of just `it`?
+    // if `it.still` is needed here, why not for the other assertions
+    // in this `describe` block?
```
sadatay: dunno if we need this TODO anymore (which i added in the first place, sorry about that)

tests/pages/package-show.js
```diff
@@ -57,6 +57,42 @@ export default {
...
+      return $('[data-test-eholdings-package-details-save-custom-coverage-button] button').click();
+    });
+  },
+
+  clickCustomCoverageAddButton() {
+    return convergeOn(() => {
+      expect($('[data-test-eholdings-package-details-custom-coverage-button] button')).to.exist;
+    }).then(() => (
+      $('[data-test-eholdings-package-details-custom-coverage-button] button').click()
+    ));
```
sadatay: we should probably extract these into a common click helper.  i'm not insisting on now, but soon.

src/components/package-show/package-show.js
elrickvm: aaaah i see. Then that `packageEdited` function will almost be like a validation function in a sense...That could get crazy really fast with lots of logic in that function rather than have explicit calls to function. But i can see the argument for the one function if its not too long.

cherewaty:
Rebased to include React 16 upgrade.

PR #153, opened 2018-01-08
Use skeleton loading css for list-item loading state

wwilsman:
## Purpose
Show a loading indicator for individual list-items within the search results list.

## Approach
Facebook's approach uses a single background with multiple divs clipping away from each visual element to hide parts of the background. This approach uses multiple css backgrounds, one for each skeleton element, and the shimmer effect is created using a psuedo element with some opacity.

#### Open Questions
- It might be more cross-browser friendly to use multiple divs like facebook (for browsers that do not support multiple backgrounds).

## Learning
[How the Facebook content placeholder works](https://cloudcannon.com/deconstructions/2014/11/15/facebook-content-placeholder-deconstruction.html)

## Screenshots
skeleton

### DIFF
--- a/src/components/package-list-item/package-list-item.css
+++ b/src/components/package-list-item/package-list-item.css
@@ -1,5 +1,78 @@
-.name {
-  overflow: hidden;
-  white-space: nowrap;
-  text-overflow: ellipsis;
+.item {
+  height: 100%;
+
+  & h5, & div {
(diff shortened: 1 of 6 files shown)


cherewaty:
This also makes it really obvious that we need to make some line-height changes on the list items. We'll tackle that soon.

PR #157, opened 2018-01-09
Warn user on deselection of final title in package

sadatay:
When the final title in a package is deselected, the entire package
itself is also deselected.  This will eliminate all package-level
customization, so we warn the user of this before proceeding.

esveeuxwqm

### DIFF
--- a/src/components/customer-resource-show/customer-resource-show.js
+++ b/src/components/customer-resource-show/customer-resource-show.js
@@ -247,17 +247,19 @@ export default class CustomerResourceShow extends Component {
                selects and then immediately deselects the
                customerResource
             */
-            model.package.selectedCount <= 1
-              ? (
-                <span data-test-eholdings-final-title-warning>
-                  Are you sure you want to remove this title from your holdings?
-                  It is also the last title selected in this package. By removing
-                  this title, you will also remove this package from your holdings
-                  and all customizations will be lost.
-                </span>
-              )
-              : 'Are you sure you want to remove this title from your holdings? ' +
-                'All customizations will be lost.'
+            model.package.selectedCount <= 1 ? (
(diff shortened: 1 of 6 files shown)


sadatay:
We weren't really sure the best way to test this.  @wwilsman brought up the possibility of using spans with data attributes to ensure that the proper message is displayed in the proper case, but we weren't sure it was necessary to go that far.  I'm fine with it as-is, but also open to feedback.

src/redux/customer-resource.js
```diff
@@ -7,6 +7,7 @@ class CustomerResource {
   vendorName = '';
   packageId = 0;
   packageName = '';
+  package = belongsTo();
```
cherewaty: Aha! That was the piece I was missing.

cherewaty:
I added some more test coverage to check for each of the different copy variants. The selection test was getting unwieldy, so I split deselection behaviors out into their own separate test.

Since I just added a lot of code, let's get one more 👍 before merging. I nominate @wwilsman.

src/components/customer-resource-show/customer-resource-show.js
```diff
@@ -241,7 +241,28 @@ export default class CustomerResourceShow extends Component {
...
+                  and all customizations will be lost.
+                </span>
+              )
+              : (
+                <span data-test-eholdings-deselect-title-warning>
+                  Are you sure you want to remove this title from your holdings?
+                  All customizations will be lost.
+                </span>
+              )
+          }
```
wwilsman: We can un-nest this by one now that we're using spans.

```js
model.package.selectedCount <= 1 ? (
  <span>...</span>
) : (
  <span>...</span>
)}
```

PR #159, opened 2018-01-10
Wrap settings in Pane for responsiveness

cherewaty:
## Purpose
Because the eHoldings settings form was not in a `Pane`, the form was inaccessible on small screens.

## Approach
Wrapped the form in a `Pane` that has a back button. The back button only appears at small sizes.

## Screenshots
localhost_3000_settings_eholdings iphone 7

### DIFF
--- a/src/components/settings/settings.css
+++ b/src/components/settings/settings.css
@@ -1,8 +1,12 @@
 @import '@folio/stripes-components/lib/variables';
 
 .eholdings-settings {
-  padding: 2em;
+  padding: 0 2em;
   width: 100%;
+
+  & h2 {
+    margin: 0.83em 0;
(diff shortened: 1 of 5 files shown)


cowboyd:
Dumb Q: Why aren't all settings wrapped in a Pane?

cherewaty:
Why aren't all settings wrapped in a pane? Most of the other apps are using the `Settings` component from `stripes-components`. It adds another level of subnavigation with a nested `Paneset`. Those are also lacking in responsive behavior, but those are an issue for another day (only the eHoldings settings and the version info appear at folio.frontside.io's settings).

cowboyd:
Got it!

PR #162, opened 2018-01-15
Track specific attribute changes in redux store

cherewaty:
## Purpose
When saving a change to a package's or customer resource's `isSelected` attribute, we were previously just relying on `model.update.isPending` to determine if the selection toggle should be in a loading state.

That was fine until we started trying to modify model attributes individually. For example, editing custom coverage does not change `isSelected`, and should not trigger a loading state for `isSelected`.

## Approach
Redux Form expects `onSubmit` to return a promise, but that doesn't jive with our use of observables. Instead, we can shove some extra data in the redux store to know which specific attributes have pending changes.

When saving a model, there will now be a `model.update.changedAttributes` object containing the keys of changed attributes, and an array of the old value and new value for each.

#### Example
From the state of an entry in `eholdings.data.packages.requests`:
```
changedAttributes: {
  selectedCount: [7,0]
  isSelected: [true, false]
}
```

#### TODOS and Open Questions
- [x] I didn't create any additional tests for this. Unsure how to proceed with that.
- [x] Should `changedAttributes()` be a property of every request in the redux store? Right now it only appears as part of save requests.

## Learning
Inspired by Ember Data's `changedAttributes()`: 
https://www.emberjs.com/api/ember-data/2.16/classes/DS.Model/methods/changedAttributes?anchor=changedAttributes
https://github.com/emberjs/data/blob/v2.16.0/addon/-private/system/model/internal-model.js#L698

### DIFF
--- a/src/redux/data.js
+++ b/src/redux/data.js
@@ -119,21 +119,24 @@ const reject = (request, errors, data) => ({
  * @param {String} data.type - the resource type
  * @param {Object} data.params - request params
  */
-const makeRequest = (type, data) => ({
-  [data.timestamp]: {
-    timestamp: data.timestamp,
-    type,
-    path: data.path,
-    resource: data.type,
-    params: data.params,
-    isPending: true,
-    isResolved: false,
-    isRejected: false,
-    records: data.params.id ? [data.params.id] : [],
-    meta: {},
-    errors: []
-  }
-});
+const makeRequest = (type, data) => {
+  return {
+    [data.timestamp]: {
+      timestamp: data.timestamp,
+      type,
+      path: data.path,
+      resource: data.type,
+      params: data.params,
+      isPending: true,
+      isResolved: false,
+      isRejected: false,
+      records: data.params.id ? [data.params.id] : [],
+      changedAttributes: data.changedAttributes,
+      meta: {},
+      errors: []
+    }
+  };
+};
 
 /**
  * Helper for retrieving or creating a record from the resource
@@ -198,6 +201,30 @@ const formatErrors = (errors) => {
   }
 };
 
+/**
+ * Helper for calculating the difference between old and new state
+ * with model.save(). Borrows heavily from Ember Data.
+ * @param {Object} oldData - current state of attributes in store
+ * @param {Object} newData - requested new state of attributes
+ * @returns {Object} set of attributes with change
+ */
+const getChangedAttributes = (oldData, newData) => {
+  let diffData = Object.create(null);
+  let newDataKeys = Object.keys(newData);
+
+  for (let i = 0, length = newDataKeys.length; i < length; i++) {
+    let key = newDataKeys[i];
+    if (oldData[key] !== newData[key]) {
+      diffData[key] = {
+        prev: oldData[key],
+        next: newData[key]
+      };
+    }
+  }
+
+  return diffData;
+};
+
 // reducer handlers
 const handlers = {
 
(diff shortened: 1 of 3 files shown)


src/redux/data.js
```diff
@@ -198,6 +201,27 @@ const formatErrors = (errors) => {
...
+ * @returns {Object} set of attributes with change
+ */
+const getChangedAttributes = (oldData, newData) => {
+  let diffData = Object.create(null);
+  let newDataKeys = Object.keys(newData);
+
+  for (let i = 0, length = newDataKeys.length; i < length; i++) {
+    let key = newDataKeys[i];
+    if (oldData[key] !== newData[key]) {
+      diffData[key] = [oldData[key], newData[key]];
```
sadatay: i dunno how i feel about relying on array position to denote something qualitative about data we're storing

sadatay:
re: testing, given the "real world" nature of our tests, one route could be to assert that edits to various fields do not change the page state of other fields.  the sort of base case you're trying to address here is around `isSelected` so you could start there - i.e. augment tests around coverage/visibility to also check that the `isSelected` toggle doesn't animate.  i'm not sure how far it would be necessary to go with this - a cross of every editable field with every other editable field seems like overkill

src/redux/data.js
cherewaty: That's a good point. It was the approach taken in Ember Data's `changedAttributes()`, but that doesn't mean we have to do it identically. Is this preferable?
```
changedAttributes: {
  selectedCount: {
    old: 7,
    new: 0
  ],
  isSelected: [
    old: true, 
    new: false
  ]
}
```

sadatay: to me, yes.  but i'm open to discussion on it.

wwilsman: I like the array notation because it's simple and most of this data stuff is inspired by Ember.

I'm also open to the object notation, as it's a bit more clear. But I'd recommend something like `prev` and `next` since `new` might not be the actual value of the attribute on the record. Like if the record changed elsewhere and is re-requested, `model.isSelected` might not reflect `model.update.changedAttributes.isSelected.new`.

wwilsman:
RE: Testing

Maybe writing unit/integration tests for it? There are currently some existing unit tests in `tests/unit/redux-model-test.js`. Although the setup in that file wouldn't work for this scenario. It's still run along with the other tests, so we have access to mirage; but I don't know how we would wire up the reducer and epic short of just creating a redux store for the purpose of these tests.

If we wanted to test them via the acceptance tests, I think just testing the loading state of selecting a package/title should suffice since we would now be using `&& 'isSelected' in model.update.changedAttributes`. And also as Joe said, add a test with the custom coverage stuff that says "does not trigger the toggle loading state".

wwilsman:
Also, I think leaving `changedAttributes` for updates only makes sense; at least for our use case. Otherwise would we want it to actually show changed attributes between requests? Like if the record was updated elsewhere, would the new request have a populated `changedAttributes` property?

And with that, sometimes the data from the requests are different because of RM API. For example, a search for titles does not contain some data, so a request for a single title after a search would unnecessarily diff the missing attributes.

src/redux/data.js
sadatay: i agree with the `prev` and `next` comment if we go the object route.  i'm down for the array notation so long as we make a comment for future travelers.  it should be pretty clear by the code, especially if this is like a well known facet of ember-data (which i don't really have the context to know), but worth at least a note imo

cherewaty:
Let's go with `prev` and `next`. Change pushed.

Re: testing: we do have assertions in `customer-resource-selection-test.js` and `package-show-selection-test.js` that check `isPending` states, so I'm comfortable this does not introduce any regressions. When I go back to custom coverage work, I'll add assertions there that the selection toggle is not in a loading state.

The one other question: let's leave `changedAttributes()` only in updates.

PR #163, opened 2018-01-15
Add settings.eholdings.enabled permission

MikeTaylor:


MikeTaylor:
Exactly. See https://github.com/folio-org/stripes-core/blob/master/src/components/Settings/Settings.js#L26-L29

PR #164, opened 2018-01-16
Upgrade stripes-core and stripes-components

cherewaty:
## Purpose
Regular update of the Frontside forks of `stripes-core` and `stripes-components` to match upstream.

## Highlights
- Better auth form behavior when errors occur
- Lots of form element style tweaks
- Datepicker display bug fix

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4,7 +4,7 @@
 
 "@folio/stripes-components@^2.0.0", "@folio/stripes-components@thefrontside/stripes-components#master":
   version "2.0.0"
-  resolved "https://codeload.github.com/thefrontside/stripes-components/tar.gz/b287bc46555e77c26f1dc7d223117985764e46ac"
+  resolved "https://codeload.github.com/thefrontside/stripes-components/tar.gz/017f2d467e49b343db553234584d6521eacaf56c"
   dependencies:
     "@folio/stripes-form" "^0.8.2"
     "@folio/stripes-react-hotkeys" "^1.0.0"
@@ -16,6 +16,7 @@
     moment "^2.17.1"
     moment-range "^3.0.3"
     mousetrap "^1.6.1"
+    normalize.css "^7.0.0"
     prop-types "^15.5.10"
     prop-types-extra "^1.0.1"
     react "^15.6.1"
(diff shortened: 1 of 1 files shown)


MikeTaylor:
Remind me, why are you maintaining your own forks? Can we do anything to make it unnecessary for you to take on this additional burden?

cherewaty:
The original reasons to fork `stripes-core` and `stripes-components`:
- Some bug once upon a time that I don't remember, and was fixed.
- Provide a place where Frontside developers can easily collaborate on branches, without needing permissions on upstream.
- Lock in the commit hashes of `stripes-core` and `stripes-components` with `yarn`, so `folio.frontside.io` doesn't get any surprise versions. 

We can probably stop `ui-eholdings` pointing to the fork at this point. The last bullet point can still be achieved pointing to upstream `#master` (not `latest`) instead of the fork. This type of PR would still be needed just to periodically update the version in `yarn.lock`.

What _would_ help a lot: a predictable release cadence for `stripes-core` and `stripes-components`. If we knew that every _x_ weeks (six weeks seems about right)  a new version would be cut of each of those, we would stay on released versions more often.

MikeTaylor:
> Provide a place where Frontside developers can easily collaborate on branches, without needing permissions on upstream.

I don't think there's any reason why Frontsiders shouldn't have permission to push to non-master branches of the folio-org repositories. Would that give you what you need?

I don't think we're _likely_ to move to a fixed release schedule any time soon; but I might be wrong. IIRC, Jason wants to do it that way (but then he never makes releases anyway!)

MikeTaylor:
Anyway, just to be clear -- of course, you are more than welcome to maintain your own forks if you still find that useful! Just looking to see if we can easily save you the work.

cowboyd:
Yeah, I think the big one was that pre-Montreal, we didn't have the ability to create branches within the stripes-core and stripes-component repos so that if we wanted to run a little bit ahead of the current release, it wasn't possible without a fork.

I think having a folio-org repo pointing to an Frontside fork is kinda weird, and should be re-aligned with the `folio-org/stripes-core`.

In the future, what we can do if we need to make changes to `stripes-core` to support work in e-holdings, we can just point to a branch / SHA within `folio-org/ui-eholdings`.

MikeTaylor:
OK, sounds good! Do you now have all the GitHub perms you need?

cherewaty:
I think we've got all the permissions we need now. I'll make another PR to switch off the forks the next time we need to updates the commit SHAs in `yarn.lock`.

MikeTaylor:
Sweet.

PR #167, opened 2018-01-17
Implement Package Level Hiding UI Changes

carolegodfrey:
## Purpose
Implement show/hide functionality at the Package Detail level.  This involves

- adding a toggle to package details that represents the isHidden state, toggle behaves as follows:
1. When the toggle is "off", isHidden=true: label is "Hidden from patrons"
2. When the toggle is "on", isHidden=false: label is "Visible to patrons"
- displaying a message to the right of toggle if package is hidden by ep  (message is Set by System - specific message text to be set in mod-kb-ebsco)
- displaying Hidden message in list of customer resources on Package details page
Also updated Vendor Detail page to show Hidden status on list of displayed packages
Updated Hide/Show toggle (disable if Package is Not Selected)

Related Jira Tickets

- https://issues.folio.org/browse/UIEH-116
- https://issues.folio.org/browse/UIEH-52
- https://issues.folio.org/browse/UIEH-117
- https://issues.folio.org/browse/UIEH-98

## Approach

Implementation is similar to Selected/Not Selected Toggle on Package Details page. Uses same Update Package functionality

#### TODOS and Open Questions

- [x] Still need to implement display of proper hidden message in mod-kb-ebsco for package level. Currently, module displays - "All titles in this package are hidden" if package is hidden and does not distinguish 'Hidden by EP' vs 'Hidden by Customer'
Completed - https://github.com/folio-org/mod-kb-ebsco/pull/67 (mod-kb-ebsco changes merged to master)
- [x] Still see a flickering of message when toggling hide/show setting - possible change at data layer or added status check needed. Note - flicker is visible in gif - not observed when testing
Updates suggested by Jeffrey have been made - flickering no longer observed
- [ ] Observe same issue as experience with selected toggle - when clicked, only first customer resource in list is displayed. Note - this issue is not observed with new pagination functionality

## Screenshots
package-hide-show

### DIFF
--- a/src/components/package-show/package-show.js
+++ b/src/components/package-show/package-show.js
@@ -19,7 +19,8 @@ import styles from './package-show.css';
 export default class PackageShow extends Component {
   static propTypes = {
     model: PropTypes.object.isRequired,
-    toggleSelected: PropTypes.func.isRequired
+    toggleSelected: PropTypes.func.isRequired,
+    toggleHidden: PropTypes.func.isRequired
   };
 
   static contextTypes = {
@@ -30,12 +31,16 @@ export default class PackageShow extends Component {
 
   state = {
     showSelectionModal: false,
-    packageSelected: this.props.model.isSelected
+    packageSelected: this.props.model.isSelected,
+    packageHidden: this.props.model.visibilityData.isHidden
   };
 
   componentWillReceiveProps({ model }) {
     if (!model.isSaving) {
-      this.setState({ packageSelected: model.isSelected });
+      this.setState({
+        packageSelected: model.isSelected,
+        packageHidden: model.visibilityData.isHidden
+      });
     }
   }
 
@@ -63,7 +68,7 @@ export default class PackageShow extends Component {
   render() {
     let { model } = this.props;
     let { intl, router, queryParams } = this.context;
-    let { showSelectionModal, packageSelected } = this.state;
+    let { showSelectionModal, packageSelected, packageHidden } = this.state;
     let historyState = router.history.location.state;
 
     return (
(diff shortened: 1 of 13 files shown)


src/components/package-show/package-show.js
```diff
@@ -138,15 +139,26 @@ export default class PackageShow extends Component {
...
+                data-test-eholdings-package-details-hidden
+                htmlFor="package-details-toggle-hidden-switch"
+              >
+                <h4>{model.visibilityData.isHidden ? 'Hide' : 'Show'}</h4>
+                <div className={styles['flex-container']}>
+                  <div className={styles['flex-item']}>
+                    <ToggleSwitch
+                      onChange={this.props.toggleHidden}
+                      checked={model.visibilityData.isHidden}
+                      isPending={model.update.isPending}
```
cherewaty: Just merged an improvement for this type of thing today: https://github.com/folio-org/ui-eholdings/pull/162.

Now you can say `isPending={model.update.isPending && 'visibilityData' in model.update.changedAttributes}`. That isolates the loading states for the selection toggle and the visibility toggle, so they're not pulsing in a loading state together.

cherewaty:
I got around the flickering of the toggle state for selection with an extra piece of state in the component: https://github.com/folio-org/ui-eholdings/blob/master/src/components/package-show/package-show.js#L33

You could do something along these lines (may not be exact):
```
state = {
    showSelectionModal: false,
    packageSelected: this.props.model.isSelected
    packageHidden: this.props.model.visibilityData.isHidden
  };

componentWillReceiveProps({ model }) {
    if (!model.isSaving) {
      this.setState({
        packageSelected: model.isSelected,
        packageHidden: model.visibilityData.isHidden
      });
    }
  }
```

Then the toggle switch's `checked` is `{this.state.packageHidden}` instead of directly going off the model.

src/components/package-show/package-show.js
carolegodfrey: Thanks for suggestions - will test both of these out!

sadatay:
this looks pretty good to me thusfar 👍

PR #170, opened 2018-01-18
Add editing for package custom coverage

cherewaty:
## Purpose
Builds on the work from https://github.com/folio-org/ui-eholdings/pull/151. Probably replaces it, but we'll leave both open until we get one to the finish line.

## Approach
Major UX changes:
- Instead of a trash can that immediately removes the custom coverage, there's a clear button that just clears out both rows. I think that will work nicer with the multiple-row situation in package-titles. It also makes it so there's always an explicit Save action.
- You can submit blank values in the form.
- All the custom coverage stuff has been moved out to a `CustomCoverageForm` component.
- There's now a background color on the form.

#### TODOS and Open Questions
- [ ] Fix failing test assertions using `fillIn()`.
- [ ] How do we deal with server errors? @elrickvm might now have some insights

## Screenshots
l65ih9xonw

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -159,12 +159,13 @@ export default function configure() {
     });
 
     let body = JSON.parse(request.requestBody);
-    let { isSelected } = body.data.attributes;
+    let { isSelected, customCoverage } = body.data.attributes;
 
     let selectedCount = isSelected ? matchingCustomerResources.length : 0;
 
     matchingCustomerResources.update('isSelected', isSelected);
     matchingPackage.update('isSelected', isSelected);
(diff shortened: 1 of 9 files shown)


elrickvm:
One thing i noticed is that the form validation is somehow not working. So its allowing a blank start date to be submitted that RMAPI does not like allow. I think it has to do with maybe an adjustment that was made for the delete/trash-can/x button to allow those fields to pass without validation.

cherewaty:
Closing in favor of https://github.com/folio-org/ui-eholdings/pull/151

PR #171, opened 2018-01-18
Switch to upstream stripes

cherewaty:
## Purpose
With the `folio-org` `stripes-core` and `stripes-components` compatibility with `ui-eholdings` no longer an issue, there's not any reason to point to `thefrontside` forks of those repos anymore.

## Screenshots
This updates the header to some shiny new styles.

localhost_3000_eholdings_vendors_5_searchtype vendors q e iphone 8

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
     "lint": "eslint ./"
   },
   "devDependencies": {
-    "@folio/stripes-components": "thefrontside/stripes-components#master",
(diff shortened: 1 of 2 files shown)


MikeTaylor:
I've gone ahead and merged this so that, by no longer having multiple instances of stripes-components in my build, I can _finally_ stop seeing the sensationally dumb warning:

> Invalid context `paneset` of type `Paneset` supplied to `Paneset`, expected instance of `Paneset`.

Woo-hoo!

PR #175, opened 2018-01-19
Show/Hide functionality for title-packages

sduvvuri-ebsco:
## Purpose
- Implement Show/Hide functionality for title-packages
- If the package is hidden or customer resource in a package is not selected, a toggle switch which is read-only and cannot be toggled will appear with a message. 
- Message in above case will be "All titles in this package are hidden." if the package is hidden. If the customer resource within package is not selected, there is no message.
- If a package is not hidden and if a package or customer resource is selected, a toggle switch reflecting current state of the title-package appears with the appropriate message, being 'Set by System' if it is 'Hidden by EP' from RM API or none if it is 'Hidden by Customer' or any other message from RM API
JIRA Tickets: 
https://issues.folio.org/browse/UIEH-47
https://issues.folio.org/browse/UIEH-36
Per change in requirement, when isHidden=true,label is "Hidden from patrons" and toggle is off
when isHidden=false, label is "Visible to patrons" and toggle is on

## Approach
- Use the existing toggle-switch component
- Implement the flex-container for having message on the right side of toggle switch
- Make .css modifications to get the toggle-switch to the required color and few modifications to use existing colors in Stripes rather than hard-coding
- Associated unit tests for the above functionality

#### TODOS and Open Questions
- [ ] Upload new recording after getting few more examples from Khalilah

## Learning
- https://reactjs.org/tutorial/tutorial.html

## Screenshots
capture

### DIFF
--- a/src/components/customer-resource-show/customer-resource-show.css
+++ b/src/components/customer-resource-show/customer-resource-show.css
@@ -19,3 +19,13 @@
 .detail-container-header {
   margin: 2rem 0;
 }
+
+.flex-container {
+  display: flex;
+  align-items: center;
+}
+
+.flex-item {
+  margin-right: 10px;
+}
+
--- a/src/components/customer-resource-show/customer-resource-show.js
+++ b/src/components/customer-resource-show/customer-resource-show.js
@@ -19,7 +19,8 @@ import styles from './customer-resource-show.css';
 export default class CustomerResourceShow extends Component {
   static propTypes = {
     model: PropTypes.object.isRequired,
-    toggleSelected: PropTypes.func.isRequired
+    toggleSelected: PropTypes.func.isRequired,
+    toggleHidden: PropTypes.func.isRequired
   };
 
   static contextTypes = {
@@ -29,12 +30,16 @@ export default class CustomerResourceShow extends Component {
 
   state = {
     showSelectionModal: false,
-    resourceSelected: this.props.model.isSelected
+    resourceSelected: this.props.model.isSelected,
+    resourceHidden: this.props.model.visibilityData.isHidden
   };
 
   componentWillReceiveProps({ model }) {
     if (!model.isSaving) {
-      this.setState({ resourceSelected: model.isSelected });
+      this.setState({
+        resourceSelected: model.isSelected,
+        resourceHidden: model.visibilityData.isHidden
+      });
     }
   }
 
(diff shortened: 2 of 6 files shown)


src/components/customer-resource-show/customer-resource-show.css
```diff
@@ -19,3 +19,14 @@
 .detail-container-header {
   margin: 2rem 0;
 }
+
+.flex-container {
+  display: flex;
+  flex-direction: row;
```
elrickvm: `flex-direction: row` is the default with you apply `display: flex` to any element. So i think removing line 25 would have the same effect.

sduvvuri-ebsco: Thanks for your feedback Elrick! You are right, row is the default direction. Testing the modified code now. Will have it pushed up soon.

cherewaty:
Just had a conversation with Khalilah about the labeling of this. Can we flip the boolean around? I think this will be much clearer.

Before:
When the toggle is "off", `isHidden=false`: label is "Show"
When the toggle is "on" (green), `isHidden=true`: label is "Hide"

After:
When the toggle is "off", `isHidden=true`: label is "Hidden from patrons"
When the toggle is "on", `isHidden=false`: label is "Visible to patrons"

src/components/customer-resource-show/customer-resource-show.js
```diff
@@ -29,12 +30,14 @@ export default class CustomerResourceShow extends Component {
...
     showSelectionModal: false,
-    resourceSelected: this.props.model.isSelected
+    resourceSelected: this.props.model.isSelected,
+    resourceHidden: this.props.model.visibilityData.isHidden
   };
 
   componentWillReceiveProps({ model }) {
     if (!model.isSaving) {
       this.setState({ resourceSelected: model.isSelected });
+      this.setState({ resourceHidden: model.visibilityData.isHidden });
```
sadatay: you can combine this with the previous line:
```
this.setState({
  resourceSelected: model.isSelected,
  resourceHidden: model.visibilityData.isHidden
});
```

tests/customer-resource-deselection-test.js
```diff
@@ -137,7 +137,7 @@ describeApplication('CustomerResourceShow Deselection', () => {
 
             it('is not hidden', () => {
               expect(resource.visibilityData.isHidden).to.equal(false);
-              expect(ResourcePage.isHidden).to.equal(false);
+              expect(ResourcePage.isHidden).to.equal(true);
```
cherewaty: I think this should still be `expect(ResourcePage.isHidden).to.equal(false);` - the page helper function might need to be switched around.

tests/pages/customer-resource-show.js
```diff
@@ -72,11 +72,35 @@ export default {
   },
 
   get isHidden() {
-    return $('[data-test-eholdings-customer-resource-show-is-hidden]').length === 1;
+    return $('[data-test-eholdings-customer-resource-toggle-hidden] input').prop('checked');
```
cherewaty: Yes, this needs to be flipped for `CustomerResourceShowPage.isHidden` to make sense. There might be a few other spots that will need updates to assertions after you make this change.

tests/pages/customer-resource-show.js
```diff
@@ -72,11 +72,35 @@ export default {
   },
 
   get isHidden() {
-    return $('[data-test-eholdings-customer-resource-show-is-hidden]').length === 1;
+    return $('[data-test-eholdings-customer-resource-toggle-hidden] input').prop('checked') === false;
```
cherewaty: 👍

tests/customer-resource-show-visibility-test.js
```diff
@@ -44,15 +46,20 @@ describeApplication('CustomerResourceShowVisibility', () => {
       });
     });
 
-    it('does not display the hidden/reason section', () => {
+    it('displays the hide/show toggle with show', () => {
```
cherewaty: Alternate suggestion for this label: `displays the visibility toggle as switched to on`. Open to any other ideas that might make it more clear.

tests/customer-resource-show-visibility-test.js
```diff
@@ -69,25 +76,104 @@ describeApplication('CustomerResourceShowVisibility', () => {
       });
     });
 
-    it('displays the hidden/reason section', () => {
-      expect(CustomerResourceShowPage.isHidden).to.be.true;
+    it('the hide/show toggle with hide', () => {
```
cherewaty: Alternate suggestion for this label: `displays the visibility toggle as switched to on`.

tests/customer-resource-show-visibility-test.js
```diff
@@ -69,25 +76,104 @@ describeApplication('CustomerResourceShowVisibility', () => {
...
+
+      resource.update('visibilityData', visibilityData);
+      resource.save();
+
+      return this.visit(`/eholdings/customer-resources/${resource.id}`, () => {
+        expect(CustomerResourceShowPage.$root).to.exist;
+      });
+    });
+
+    it('the hide/show toggle with hide', () => {
```
cherewaty: Alternate suggestion for this label: `displays the visibility toggle as switched to on`.

tests/customer-resource-show-visibility-test.js
```diff
@@ -69,25 +76,104 @@ describeApplication('CustomerResourceShowVisibility', () => {
...
+        isSelected: true,
+        title
+      });
+
+      return this.visit(`/eholdings/customer-resources/${resource.id}`, () => {
+        expect(CustomerResourceShowPage.$root).to.exist;
+      });
+    });
+
+    it('displays the hide/show toggle with show', () => {
```
cherewaty: I think this label is reversed? `displays the visibility toggle as switched to on` would work.

tests/customer-resource-show-visibility-test.js
```diff
@@ -98,12 +184,76 @@ describeApplication('CustomerResourceShowVisibility', () => {
       });
     });
 
-    it('displays the hidden/reason section', () => {
-      expect(CustomerResourceShowPage.isHidden).to.be.true;
+    it('the hide/show toggle with hide', () => {
```
cherewaty: Alternate suggestion for this label: `displays the visibility toggle as switched to on`.

tests/customer-resource-show-visibility-test.js
```diff
@@ -98,12 +184,76 @@ describeApplication('CustomerResourceShowVisibility', () => {
...
+
+      resource.update('visibilityData', visibilityData);
+      resource.save();
+
+      return this.visit(`/eholdings/customer-resources/${resource.id}`, () => {
+        expect(CustomerResourceShowPage.$root).to.exist;
+      });
+    });
+
+    it('the hide/show toggle with hide', () => {
```
cherewaty: Alternate suggestion for this label: `displays the visibility toggle as switched to on`.

sadatay:
same 👍

PR #180, opened 2018-01-23
Paginate related resources and use common details-view component

wwilsman:
## Purpose

Makes related resources on the various details pages paginated via impagination and our `QueryList` component.

Also extracts all of the common layout and styles into a `DetailsView` component.

## Approach

- Created a common component for the various details pages so this single component can manage loading states, error states, and scroll behavior with related resource lists.

- Made `ScrollView` stateful so that the `offset` and `onUpdate` params can truly be optional.

- Implemented a `slice` method for our collection so that a collection can be directly passed to the `ScrollView` component. This is necessary for a title page where the related resources are not paginated, and instead are simply included with the title request.

- Added the ability to unload records from the store. This allows us to unload records from our route and have the `Impagination` component automatically re-initialized to make new requests for the current read offset. With this comes a `hasUnloaded` property on collections and query requests. `Impagination` will check the current horizon of records for `request.hasUnloaded` to determine if it needs to reload the dataset.

- Optimized `collection.request` and `resolver.getRequest` so that the request getter for collections does not have to loop over a theoretical amount of pages. Instead, it will now just find the first matching resolved request in the store as oppose to searching the store for a request for every page.

- Also fixed a bug in `resolver.getRequest` when matching the `path` of requests where package
customer-resource requests were being returned for title customer-resources because there is no path for title customer-resources. When `path` was `undefined`, the request matcher didn't bother matching on the path. Now it will still attempt to match on the path as long as `path` key exists in the data passed to `getRequest`.

#### TODOS and Open Questions

- Currently, there are no query params being tracked in the URL for related resources. `offset` would clash with a search `offset`, so we'll have to come up with another name for it.

- Coverage is barely passing at ~86% for code branches. I attempted to ignore `mirage/scenarios/default.js` and a few helpers in `tests/helpers.js` that are only being used in development and not actually used at all when tests are run. I tried adding `/* istanbul ignore next */` and `coverageReporter.check.global.excludes` to Karma but neither seemed to change the coverage output (it was still checking for coverage in those places).

- There is also low coverage for `ScrollView` (which has technically increased from master) but the code that isn't covered is simply future-proof code that would allow us to change the readoffset without scrolling the list. We don't currently do that anywhere in the app, but it's there to avoid bugs if we ever actually do that.


src/components/details-view/details-view.css
```diff
@@ -0,0 +1,56 @@
...
+    /* the box-model needs _some_ height that is not 'auto' for flexbox to
+       be able to fill this element to fit the remaining space */
+    min-height: 0px;
+    flex: 1;
+  }
+}
+
+/* like `marginTopHalf` or `marginBottomHalf` from stripes-components,
+   but for right margins instead (and without `!important`) */
+:global(.marginRightHalf) {
```
cherewaty: I don't love adding this global (not a fan of the `stripes-components` ones it's inspired by, either). Could we incorporate this as a shared style somehow without putting into the global namespace?

src/components/details-view/details-view.js
```diff
@@ -0,0 +1,169 @@
...
+
+import PaneHeader from '@folio/stripes-components/lib/PaneHeader';
+import Icon from '@folio/stripes-components/lib/Icon';
+
+import KeyValueLabel from '../key-value-label';
+import styles from './details-view.css';
+
+const cx = classNames.bind(styles);
+
+export default class DetailsView extends Component {
```
sadatay: it might be worth throwing a comment in about what this component does and generally how it works.  the line-by-line comments are helpful, but to someone coming in from the outside it may be difficult to parse what the intent is, how the scroll math/logic works and what things like `sticky` mean in the context of the component.  i had the benefit of riding along while you were writing some of this so i somewhat get the picture, but to someone with fresh eyes this is pretty intense.  it is also pretty clearly delineated in your PR description, so i guess it could be argued that people could look it up there.

src/components/title-show.js
```diff
@@ -0,0 +1,86 @@
...
+      showPaneHeader={!queryParams.searchType}
+      paneHeaderFirstMenu={historyState && historyState.eholdings && (
+        <PaneMenu>
+          <div data-test-eholdings-title-show-back-button>
+            <IconButton icon="left-arrow" onClick={() => router.history.goBack()} />
+          </div>
+        </PaneMenu>
+      )}
+      bodyContent={(
+        <div>
```
sadatay: not a blocker, but i wonder if it's possible to extract the JSX we pass to `bodyContent` to somewhere else - maybe a separate file with an export or a variable somewhere else in the component.  inlining it can make the `DetailsView` component a bit unwieldy with respect to readability imo, especially for larger detail views.

sadatay:
this is some super cool stuff.  my comments were minor things, overall this looks good to me.

src/components/details-view/details-view.js
wwilsman: Good point. I thought I documented this, but guess not

src/components/title-show.js
wwilsman: I thought of this too. I think we can keep them in the same file, but just in a separate function. This way we can read the body function separate, and this prop would look like:
```
bodyContent={titleBodyContent(model)}
```

PR #181, opened 2018-01-24
Run tests in IE, Edge, and Firefox

cherewaty:
## Purpose
We can run our tests in more browsers on BrowserStack, but `allow_failures`.

## Approach
`fast_finish` tells Travis it doesn't have to wait for allowed failures before it can declare the build successful.

## Browser statuses
- Firefox: really close to passing, I would think it's an easy fix
- Edge: having some date issues
- IE11: doesn't even start running the tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,15 @@ language: node_js
 env:
   - BROWSER=Chrome_travis_ci
   - BROWSER=bs_safari_11
+  - BROWSER=bs_firefox_mac
+  - BROWSER=bs_ieEdge_windows
+  - BROWSER=bs_ie11_windows
+matrix:
+  fast_finish: true
+  allow_failures:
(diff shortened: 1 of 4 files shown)


wwilsman:
So it looks like the Firefox issue was two-fold. 

1. The event wasn't being triggered. This was fixed by mimicking react-dom's test suite using `new MouseEvent`.

2. The settings validation tests were testing the border color, which wasn't being returned in the expected format. Fixed this one by testing for the presence of `has-error--` in the className.

cherewaty:
👏 thanks @wwilsman. I added a new commit that no longer allows failures on Firefox.

sadatay:
so failures on firefox and chrome are not allowed but we're letting them slide on edge/ie, correct?  that seems fine to me 👍

PR #182, opened 2018-01-24
Default to okapi.frontside.io in .stripesclirc

wwilsman:
#178 Missed an instance where we need to change the default okapi url


sadatay:
nice catch 👍

PR #184, opened 2018-01-29
Open managed URL in a new tab

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-95, managed URL for a title-package should open in a new tab as opposed to the current behavior of opening in the same tab.

## Approach
- Add target="_blank" to open url in a new tab
- Add associated unit tests

#### TODOS and Open Questions
- [ ] Not sure if these unit tests are necessary and if this is the best approach for testing if the link opens in a new tab.
- [ ] Eventually, we need to implement this as well for this feature to be complete: https://issues.folio.org/browse/UIEH-129

## Learning
https://stackoverflow.com/questions/850058/is-it-possible-to-detect-if-a-user-has-opened-a-link-in-a-new-tab

## Screenshots
ui-eholdings-managed-url

### DIFF
--- a/src/components/customer-resource-show/customer-resource-show.js
+++ b/src/components/customer-resource-show/customer-resource-show.js
@@ -150,7 +150,7 @@ export default class CustomerResourceShow extends Component {
               {model.url && (
                 <KeyValueLabel label="Managed URL">
                   <div data-test-eholdings-customer-resource-show-managed-url>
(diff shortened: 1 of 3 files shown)


cherewaty:
The test addition looks good to me.

When we have the appropriate icon, let's throw an `alt` attribute on it that says "opens in new window".

PR #189, opened 2018-02-01
React16

doytch:
## Purpose
Upgrade to React 16


cherewaty:
@doytch is this still a work-in-progress? Should I close it?

doytch:
Yeah this is in a holding pattern. Basically the goal with moving to R16 is to only define our React dependency in one place, stripes-core. However, in our current state, this won't work with the CI tests we're running on a per-module basis since the modules are tested by themselves. When we `yarn` during CI, we won't install React since it'll be a peer dependency and the tests will fail to execute.

During last Thursday's FE meeting we were thinking that ( @ttamj correct me if I misspeak) we'd set up the testing environment using stripes-cli which would bring in stripes-core which would bring in React. (Bonus: all the testing packages would only need to be defined in one place as well.)

As a result, I'm sitting on the merges for stripes-core and modules that have test suites.

ttamj:
@doytch That sounds about right.  I opened STCLI-19 to track this, though I haven't been able to start it yet.  Would it make sense, in the interim, to move to React 16 in the apps w/ peer dependency as planned, but include the dev dependency in the stripes-* modules for CI?  I would see this as only a stop gap, with the dev dependency removed upon completing STCLI-19 (several other dev dependencies would be removed in the process w/ STCLI-19 anyway).

doytch:
Yeah when we discussed this I figured I'd just hold off to reduce the amount of churn, but you're right - if we'll be nixing a bunch of devDeps in that future PR as well, it'll be the same amount of changes in the end.

So today I'll work on updating the outstanding PRs I've got going and hopefully get R16 merged in as a devDep where necessary.

doytch:
Alright, we should be good for a merge on this now if y'all are, @cherewaty

PR #192, opened 2018-02-01
Add ability to filter title searches

wwilsman:
## Purpose

Adds a filter component to the title search form and allows filtering search results via a `filter` query param.

## Approach

- The search route will provide a filter component based on the current search type

- The filter component will receive a filter property and an `onUpdate` hook to update the form component's state

- The form component simply supplies the filter param on submit, and the search route adds it to the URL and makes the request

## Screenshots

filtering

### DIFF
--- a/src/components/title-search-filters.js
+++ b/src/components/title-search-filters.js
@@ -36,7 +36,7 @@ export default function TitleSearchFilters({
         separator={false}
         header={FilterAccordionHeader}
         displayClearButton={pubtype !== 'all'}
-        onClearFilter={() => onUpdate({ ...filter, type: null })}
+        onClearFilter={() => onUpdate({ ...filter, type: undefined })}
         closedByDefault={false}
       >
         {pubtypeFilters.map(({ label, value }, i) => (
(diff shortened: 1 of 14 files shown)


src/components/title-search-filters.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Accordion, FilterAccordionHeader } from '@folio/stripes-components/lib/Accordion';
+import RadioButton from '@folio/stripes-components/lib/RadioButton';
+
+const pubtypeFilters = [
+  { label: 'All', value: 'all' },
```
cherewaty: This list of filterable types will probably be better off coming from the back end, but hardcoding it in the front-end is probably an acceptable first pass.

src/components/title-search-filters.js
```diff
@@ -0,0 +1,62 @@
...
+        header={FilterAccordionHeader}
+        displayClearButton={pubtype !== 'all'}
+        onClearFilter={() => onUpdate('all')}
+        closedByDefault={false}
+      >
+        {pubtypeFilters.map(({ label, value }, i) => (
+          <RadioButton
+            key={i}
+            name="pubtype"
+            id={`filter-${value}`}
```
cherewaty: We should probably give a more specific id to prevent any collisions. `eholdings-title-search-pubtype-filter-${value}`?

PR #196, opened 2018-02-05
Rewrite impagination component

wwilsman:
## Purpose
When records are resolved, impagination keeps a reference to the record in our store. When the record is updated immutably, the old record becomes stale and is not reflected in UI using the impagination component.

## Approach
By using impagination's state machine directly we can resolve or reject records at will, without promises. This allows us to update any records currently in the load horizon with accurate records from our own data store.

### DIFF
--- a/src/components/impagination.js
+++ b/src/components/impagination.js
@@ -1,24 +1,24 @@
 import { Component } from 'react';
 import PropTypes from 'prop-types';
-import Dataset from 'impagination';
+import State from 'impagination/src/state';
 
 /**
- * Patch the slice method of a Dataset state.
+ * Patch / add methods to a Dataset state
  *
- * Currently it is broken such that state.slice(x, y) will return a
- * slice with `undefined` values if the values of `x` and `y`
- * reference records that are out of the dataset bounds. Instead, the
- * standard JS behaviro is to truncate the slice at the last available
- * record. If there are no avaliable records, then it will be an empty
- * set.
- *
- *
- * @see https://github.com/flexyford/impagination/issues/48
- * @param {Object} datasetState - Dataset state instance
- * @returns {Object} new Dataset state instance
+ * Due to the way the state's immutability is handled, we cannot
+ * extend it to create a patched state class
  */
-function patchDatasetSlice(datasetState) {
-  return Object.create(datasetState, {
+function patchDatasetState(dataset) {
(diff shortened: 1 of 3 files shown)


src/components/impagination.js
```diff
@@ -50,156 +51,113 @@ export default class Impagination extends Component {
...
+    dataset.unrequested.forEach((p) => {
+      // we use a 1-based page number
+      fetch(p.offset + 1, pageSize);
+    });
+
+    // update newly requested pages
+    dataset = dataset.fetch(dataset.unrequested);
+
+    // resolve or reject all pages within the load horizon
+    this.getPagesInHorizon(dataset).forEach((page) => {
```
cowboyd: Do we want to only handle pages in the horizon? or do we want to consider each page with an outstanding request. (isPending)

Also, what about unfetchable pages? I.e. pages outside the load horizon that we can unload.

wwilsman: So we don't want to only consider pending requests, because when a single package is updated, the page's request is still resolved and does not change. So we would still have the problem where selecting a package does not get updated in the list.

If we do this to all pages outside of the load horizon, this gets very slow to update records that are outside of the view after you've scrolled down a long list. I think within the horizon is fine because this component is just for the _display_ of data. And we would never need to know the status of dataset records outside of the view. Especially since our records have their own `isResolved` and `isRejected` properties.

As for unfetchable pages, I do think I should add some logic around unfetching pages with errors. But otherwise what would be the benefit of unloading previous records in the list? If we unload them and you end up on the page of an unloaded record, we lose the snapyness of having that data already available in the store.

cowboyd: I guess the edge-case I'm thinking of is when we have page requested, and then it leaves the horizon. Will it get stuck in a pending state if it never gets resolved? Or will we resolve it once it moves back into the horizon.

Also, this seems like a candidate for moving upstream. Can we wrap our dataset state to provide a `horizonPages` property so that the interface won't change if we open up an upstream PR?

wwilsman: The data in our store will be resolved whenever the request is resolved. The dataset page object will get resolved once it's back in the horizon and the data in our store is also resolved.

Wil do on that `horizonPages` property

cowboyd: maybe call it `pagesWithinHorizon` so that it's more easily understandable.

PR #198, opened 2018-02-08
Install fresher stripes cli, core

cherewaty:
I wanted something small to verify our new Circle CI deploy pipeline is working correctly.

This is a small version bump to `stripes-cli` (plus `stripes-core`). We can also probably start using released versions of the CLI from now on, vs. the tip of `master`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,11 +8,11 @@
   "scripts": {
     "start": "NODE_ENV=development stripescli serve demo/stripes.config.js",
     "build": "stripescli build demo/stripes.config.js dist",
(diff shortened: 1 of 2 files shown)


cherewaty:
Cool, I dropped the explicit `--configFile`.

ttamj:
Also, the CLI is invoked with `stripes` starting in v0.15.0.  This will officially be released as v1.0.0 tonight if I can get to it, otherwise probably tomorrow.  I'll run it through my `testing-cli-changes` branch on ui-eholdings first.

PR #199, opened 2018-02-09
upgrade to @folio/eslint-config-stripes

doytch:
## Purpose
Move to a unified set of linter rules for Stripes repos and be able to pull the latest versions of those rules.

## Approach
Switch to the new folioci-based, @folio-scoped, eslint-config-stripes package.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,7 @@ module.exports = {
       experimentalObjectRestSpread: true
     }
   },
-  extends: ['stripes'],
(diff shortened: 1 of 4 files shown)


doytch:
[This is the failing rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/default-props-match-prop-types.md)

cherewaty:
Fixed the `default-props-match-prop-types` failure and `yarn`-ed (`ui-eholdings` has a `yarn.lock`, which might need some rethinking).

PR #201, opened 2018-02-13
Implement editable custom embargo

sduvvuri-ebsco:
## Purpose
Initially, custom embargo was only readable but could not be edited. This implementation makes the custom embargo editable. 

## Approach
- Custom Embargo Section should show up and be editable only when a title-package is selected, if not, it should be hidden
- If custom embargo is not present, "Add custom embargo" button should show up
   - When the user clicks on the button, a form should open up containing a textfield to input embargo 
      value, a dropdown to select embargo unit, a save button and a cancel button
- If custom embargo is present, that value should be displayed with a edit icon next to it, clicking on which the custom embargo form should open up
- User should be able to enter only positive integers in the text field
- User should be able to select one of None, Days, Weeks, Months, Years as embargo unit
- When user selects "None" textfield should automatically update to "0"
- If user enters 0 and selects a unit other than None, error message shows up preventing the user from clicking "Save" button
- If user enters valid value and selects "None", error message shows up asking the user to select a valid unit
- On clicking "Save", embargo value and unit get updated and show the new value on the screen
- On clicking "cancel", old embargo value is retained and nothing gets updated
- Added associated unit tests

#### TODOS and Open Questions
- [ ] There is a warning in JavaScript console that appears because Stripes UI TextField component does not support numbers as proptypes, and we input only numbers because client side validation checks for it as well as RM API expects a number for custom embargo value. Discussed the fix with Joe and Jeffrey and merged a PR to master in Stripes UI TextField component for supporting numbers and strings https://github.com/folio-org/stripes-components/pull/234
Tested in local using yarn link that this fix actually fixes the JS console warning we have been seeing earlier.

## Learning
https://davidkpiano.github.io/react-redux-form/docs.html
https://facebook.github.io/jest/docs/en/tutorial-react.html

## Screenshots
custom-embargo

### DIFF
--- a/src/components/custom-embargo/custom-embargo.css
+++ b/src/components/custom-embargo/custom-embargo.css
@@ -1,0 +1,49 @@
+.custom-embargo-form,
+.custom-emabrgo-form-editing {
+  padding: 0.5em;
+}
+
+.custom-embargo-form-editing {
+  background: #e6f3ff;
+}
+
+.custom-embargo-add-button {
+  margin: 1em 0;
+}
+
+.custom-embargo-display {
+  display: flex;
+  align-items: center
+}
+
+.custom-embargo {
+  display: flex;
+  align-items: flex-start;
+}
+
+.custom-embargo-action-buttons {
+  display: flex;
+  margin-top: 1em;
+  padding: 0.5em;
+  align-items: flex-start;
+}
+
+.custom-embargo-container {
+  display: flex;
+  align-items: flex-start;
+  padding: 0.5em;
+}
+
+.custom-embargo-text-field {
+  align-self: flex-start;
+  flex: 1 1 auto;
+  margin-right: 1em;
+}
+
+.custom-embargo-component-width {
+  width: 13em;
+}
+
+.error-message {
+  color: red;
+}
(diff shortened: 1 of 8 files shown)


cherewaty:
Relies on https://github.com/folio-org/stripes-components/pull/234

cherewaty:
Now that the change is in to `stripes-components`, you can upgrade the version of `stripes-components` used by `ui-eholdings` to see if that warning goes away:
1. `yarn remove @folio/stripes-components`
2. `yarn cache clean`
3. Discard changes to `package.json` (`@folio/stripes-components` should appear again)
4. `yarn` - you should end up with changes to `yarn.lock`
5. Check in `yarn.lock`

sduvvuri-ebsco:
Thanks so much Jeffrey! tried those steps above and it works!

cherewaty:
The https://github.com/folio-org/ui-eholdings/pull/192 merge caused a `yarn.lock` conflict. Nothing a rebase can't fix.

sduvvuri-ebsco:
thanks Jeffrey, triple checked that the number warning is gone, merging this in

PR #202, opened 2018-02-13
Prevent a JS Console Warning in Details View

sduvvuri-ebsco:
## Purpose
Prevent a warning in JS console getting thrown because a mousewheel event is being triggered with no list element

## Approach
If the list is empty, then do not run the rest of the handleWheel function and return

## Learning
- Consultation with Wil


cherewaty:
👏 thanks for fixing @sduvvuri-ebsco

PR #205, opened 2018-02-14
Use newer version of yarn on Travis

cherewaty:
## Purpose
I noticed we're installing an older version of `yarn` over the one already on Travis.

screen shot 2018-02-14 at 1 35 53 pm

Yarn is at a much more stable point now than it was six months ago, so just going with the `yarn` version already installed seems pretty safe.

## Approach
Let's just stick with the version of `yarn` that comes by default on the Travis boxes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,6 @@ before_install:
   - export CHROME_BIN=google-chrome-stable
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
-  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.1.0
-  - export PATH=$HOME/.yarn/bin:$PATH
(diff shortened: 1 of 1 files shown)


sadatay:
interesting for a few reasons.  i wonder why `https://yarnpkg.com/install.sh` is out of step with the most recent version of yarn - it'd be one thing if it we were getting it from the package repos for the given linux distro but since we're getting it directly this seems weird to me.  i wonder if the travis box is using a bleeding edge version and that url just uses the most recent stable version?

either way, if travis has `yarn` baked in, no reason to re-install it here.  also, aren't we getting rid of travis entirely soon?

cherewaty:
@sadatay we were explicitly asking for `--version 1.1.0` with the install script.

sadatay:
ah, i missed that.  indeed we are.

PR #206, opened 2018-02-14
Fix flakey tests

wwilsman:
## Purpose
Some of our tests can be flakey due to varying reasons. This attempts to fix some of that flakiness.

## Approach
- Fix customer-resource test where we were incorrectly testing for `false`.
- Remove coverage display test because there will never be a situation in which the API returns invalid dates.
- Use it.still for checking if states are not equal

#### TODOS and Open Questions
There are probably other flakey tests as well, but this is a good start to fixing some of them.

## Learning
- When writing tests for things that do not change, or for states that are not equal, we must use `it.still`. This is because when using the convergence `it`, it might pass before the DOM is updated to display the correct information. `it.still` will make sure the assertion passes for at least 50ms before considering it to actually be passing.

  [The readme for `@bigtest/mocha` explains this a little further](https://github.com/thefrontside/bigtest/tree/master/packages/mocha#writing-tests)

### DIFF
--- a/tests/backend-configuration-test.js
+++ b/tests/backend-configuration-test.js
@@ -89,11 +89,11 @@ describeApplication('With unconfigured backend', {
           });
         });
 
-        it('does not display an error state for customer id', () => {
+        it.still('does not display an error state for customer id', () => {
           expect(SettingsPage.customerIdFieldIsInvalid).to.be.false;
         });
 
-        it('does not display an error state for api key', () => {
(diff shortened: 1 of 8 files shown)


sadatay:
makes sense and looks good :shipit: .  a little scary to think of where else this could be occurring in the suite, but this at least gets us past the practical issue of flakey tests in the current build

PR #207, opened 2018-02-14
Add selected title search filter

wwilsman:
## Purpose
Adds the selection status filter to title searches

## Approach
Adds another accordion component with radio buttons for the selection status.

#### TODOS and Open Questions
- We should really return the available filters from the API. Any suggestions on how that would look?


cherewaty:
Re: getting the filters from the API: I guess every search type (provider/package/title) would have a `fields` endpoint that returns name, type, etc.?

PR #209, opened 2018-02-19
Implement Targeted Search Options on Title List

carolegodfrey:
## Purpose
Implement Targeted Search Options for Title Searches.  This allows searching for titles by
- title (ie name)
- issn/isbn
- publisher
- subject
https://issues.folio.org/browse/UIEH-120

## Approach
- Search form is updated to use searchfield component ( (combines a select and text box) for Title searches. Continue to use simple search text box for Provider and Package searches
- eHoldings URL continues to use **q** parameter and adds a new **searchfield** parameter for Title search options, For example?q=education&searchfield=subject
- Supporting requests to mod-kb-ebsco are made using a JSON API filter syntax (for example, filter[subject]=education)
- Updates to mirage configuration to support new search filters
- When switching between search options (for example from publisher to subject), the search box maintains the search term

## Learning
- Used searchfield component as is used in codex search application
https://github.com/folio-org/stripes-components/blob/master/lib/structures/SearchField/readme.md
- https://www.npmjs.com/package/qs

## Screenshots
targeted-search-revised

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -9,18 +9,18 @@ import { Response } from 'mirage-server';
  * @returns {Function} route handler for a search route of this
  * resource type
  */
-function searchRouteFor(resourceType, filter = () => true) {
+function searchRouteFor(resourceType, filter = (model, req) => {
+  let query = req.queryParams.q.toLowerCase();
+  return model.name && model.name.toLowerCase().includes(query);
+}) {
   return function (schema, req) { // eslint-disable-line func-names
-    let query = req.queryParams.q.toLowerCase();
     let page = Math.max(parseInt(req.queryParams.page || 1, 10), 1);
     let count = parseInt(req.queryParams.count || 25, 10);
     let offset = (page - 1) * count;
 
     let collection = schema[resourceType];
     let json = this.serialize(collection.all().filter((model) => {
-      return model.name
-        && model.name.toLowerCase().includes(query)
-        && filter(model, req);
+      return filter(model, req);
     }));
 
     json.meta = { totalResults: json.data.length };
@@ -176,6 +176,10 @@ export default function configure() {
   // Package resources
   this.get('/packages', searchRouteFor('packages'));
 
+  this.get('/titles/:id', ({ titles }, request) => {
+    return titles.find(request.params.id);
+  });
+
   this.get('/packages/:id', ({ packages }, request) => {
     let pkg = packages.find(request.params.id);
 
@@ -214,7 +218,22 @@ export default function configure() {
     let params = req.queryParams;
     let type = params['filter[type]'];
     let selected = params['filter[selected]'];
+    let name = params['filter[name]'];
+    let isxn = params['filter[isxn]'];
+    let subject = params['filter[subject]'];
+    let publisher = params['filter[publisher]'];
+    let queryFiltered = true;
     let filtered = true;
+
+    if (name) {
+      queryFiltered = title.name && title.name.toLowerCase().includes(name.toLowerCase());
+    } else if (isxn) {
+      queryFiltered = title.identifiers && title.identifiers.some(i => i.id.toLowerCase().includes(isxn.toLowerCase()));
+    } else if (subject) {
+      queryFiltered = title.subjects && title.subjects.some(s => s.subject.toLowerCase().includes(subject.toLowerCase()));
+    } else if (publisher) {
+      queryFiltered = title.publisherName && title.publisherName.toLowerCase().includes(publisher.toLowerCase());
+    }
 
     if (type && type !== 'all') {
       filtered = title.publicationType.toLowerCase() === type;
@@ -226,7 +245,7 @@ export default function configure() {
       });
     }
 
-    return filtered;
+    return filtered && queryFiltered;
   }));
 
   this.get('/titles/:id', ({ titles }, request) => {
--- a/src/routes/search.js
+++ b/src/routes/search.js
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-
 import { qs } from '../components/utilities';
 import { createResolver } from '../redux';
 import Provider from '../redux/provider';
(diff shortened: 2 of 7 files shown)


mirage/config.js
```diff
@@ -9,18 +9,18 @@ import { Response } from 'mirage-server';
  * @returns {Function} route handler for a search route of this
  * resource type
  */
-function searchRouteFor(resourceType, filter = () => true) {
+function searchRouteFor(resourceType, filter = (title, req) => {
+  let query = req.queryParams.q.toLowerCase();
+  return title.name && title.name.toLowerCase().includes(query.toLowerCase());
```
wwilsman: To avoid confusion, we should use `model` here instead of `title` since this applies to packages and providers.

Also `let query` already has `toLowerCase` so the second `query.toLowerCase()` isn't necessary.

mirage/config.js
```diff
@@ -226,9 +247,14 @@ export default function configure() {
...
 
-    return filtered;
+    return filtered && queryFiltered;
   }));
 
+  this.get('/titles/:id', ({ titles }, request) => {
+    return titles.find(request.params.id);
+  });
+
+
```
wwilsman: Looks like this was duplicated by accident

src/routes/search.js
```diff
@@ -75,6 +74,21 @@ class SearchRoute extends Component {
     this.setState({ searchType, params });
   }
 
+  /**
+   * Transforms UI params into search params
+   * @param {*} parms - query param object
```
wwilsman: This should be `@param {Object} params`

src/routes/search.js
```diff
@@ -83,12 +97,13 @@ class SearchRoute extends Component {
   getResults() {
     let { resolver } = this.props;
     let { searchType, params } = this.state;
-    let { offset = 0, ...queryParams } = params;
+    let searchParams = this.transformQueryParams(params);
+    let { offset = 0, ...pagedParams } = searchParams;
```
wwilsman: `offset` comes from the URL params, so it shouldn't come from the tranformed search params. You could leave the old destructuring and just transform `queryParams`

```
let { offset = 0, ...queryParams } = params;
let searchParams = this.transformQueryParams(queryParams);
```

src/routes/search.js
```diff
@@ -83,12 +97,13 @@ class SearchRoute extends Component {
...
     let page = Math.floor(offset / 25) + 1;
 
     return resolver.query(searchType, {
       filter: undefined,
-      ...queryParams,
+      ...pagedParams,
```
wwilsman: This can be `searchParams` once we change that transform. `pagedParams` is sort of a misnomer since it doesn't actually include any page parameter.

tests/package-search-test.js
```diff
@@ -148,7 +148,7 @@ describeApplication('PackageSearch', () => {
       });
 
       it('displays an empty search', () => {
-        expect(PackageSearchPage.$searchField).to.have.value('');
+        expect(PackageSearchPage.$titleSearchField).to.have.value('');
```
wwilsman: Since titles are resources in our app `titleSearchField` feels a little confusing. Maybe we should go with `nameSearchField` since all of our resources have `name` attributes

tests/pages/title-search.js
```diff
@@ -27,9 +27,11 @@ export default {
   },
 
   get $searchField() {
+    return $('[data-test-title-search-field]').find('input[name="search"]');
+  },
+  get $otherSearchTypeSearchField() {
```
wwilsman: In `pages/package-search` we left this as `$searchField` and called the other one `$titleSearchField`. We should keep these consistent and decide which way we want them to be.

tests/title-search-test.js
```diff
@@ -9,16 +9,42 @@ describeApplication('TitleSearch', () => {
   let titles;
 
   beforeEach(function () {
-    titles = this.server.createList('title', 3, 'withPackages', {
+    titles = this.server.createList('title', 3, 'withPackages', 'withSubjects', 'withIdentifiers', {
       name: i => `Title${i + 1}`,
-      publicationType: i => (i % 2 ? 'book' : 'journal')
+      publicationType: i => (i % 2 ? 'book' : 'journal'),
```
wwilsman: We use this `i % 2` statement multiple times now, we should add a comment explaining that only odd indexed items are given these alternate attributes.

mirage/config.js
```diff
@@ -214,8 +218,25 @@ export default function configure() {
     let params = req.queryParams;
     let type = params['filter[type]'];
     let selected = params['filter[selected]'];
+    let name = params['filter[name]'];
+    let isxn = params['filter[isxn]'];
+    let subject = params['filter[subject]'];
+    let publisher = params['filter[publisher]'];
+    let queryFiltered = false;
```
wwilsman: We should default this to `true` and remove the last `else` below. This keeps it consistent with `filtered` and says that we should return true unless a more specific filter is provided.

carolegodfrey: Thanks for all the comments @wwilsman - will refine as suggested. As an added note - reviewed functionality with Khalilah (when switching between SearchOptions) and we will be maintaining the search term rather than clearing it out

mirage/config.js
carolegodfrey: Yes - probably an issue during a rebase

tests/package-search-test.js
carolegodfrey: Only thing here is that we are referring to the title search element from the package search. It seems like name would apply to package and provider in addition to title. We could call them titleSearchField, packageSearchField and providerSearchField distinctly in all pages provider-search.js, title-search.js and package-search.js

tests/pages/title-search.js
carolegodfrey: could add a packageSearchField and a providerSearchField (these would be the same element)

tests/package-search-test.js
wwilsman: Ah, I see. I think what confused me is in the gif we're calling the title's name field `title` so I had assumed `$titleSearchField` was meant to be the name search field.

carolegodfrey:
- Revised code to address requested code changes
- Corrected bug where an incorrect search filter (filter[name],filter[isxn],filter[publisher] or filter[subject])  was being propagated on url when switching search types (if both a search field and another filter were selected on a title search)
- Added unit tests to check bug fix 
- Re-Recorded gif to reflect requested  functionality - search box does no longer clear when changing search options

PR #210, opened 2018-02-19
 Add editing for customer resource custom coverage

cherewaty:
## Purpose
Most of https://issues.folio.org/browse/UIEH-33.

As an electronic resource librarian
I want to add/edit custom coverage at the title-package level
So that I can accurately detail my library's access to that title in the package as our coverage differs from the provided managed coverage.

## Approach
This gets the bare minimum implementation out for editing custom coverage on a package title. There's a lot more business logic from the EBSCO KB RM API that can be worked out in separate pull requests.

### Notable behavior
- I figured out an `isPending` pattern that seems to work nicely, where the form doesn't disappear until the change applies. It should be applied to package coverage and package title embargo editing.
- I used a `fieldset`, and the entire `fieldset` (including the `legend`) gets a blue background when in edit mode.

### Why not use the `RepeatableField` from `stripes-components`?
- I'm uncomfortable with how coupled it is with Redux Form. I'd much prefer a component that simply communicates through Data Down, Actions Up. Components relying on Redux Form have a lot of opaque magic going on, and it's hard to tell the difference between Redux Form API and component-specific logic.
- The focus is automatically managed in `RepeatableField` - so when you add a new row, the first input of that row gets immediately focused. That's important for accessibility, but really jarring with a `Datepicker`. We need to figure out a good strategy there.
- The first row of the `RepeatableField` is not removable.
- The add row button positioning could use some polish.
- I prefer the circle X icon we used over the trash can for the repeatable field use case.
- I didn't dig into the testability, and opted for speed of implementation over reusability.

TL;DR: We could spend another week or two shaking out the `RepeatableField` with a design everyone agrees on and functionality eHoldings needs, or move forward now with this.

## Next Steps
https://issues.folio.org/browse/UIEH-160
- This does not validate that the customer resource's custom coverage ranges are within the package's range.
- This does not attempt to sort the coverage ranges. `mod-kb-ebsco` should probably do that sorting, not the front end.
- This does not check overlapping of coverage ranges.
- Managed coverage and custom coverage need to integrated into the same widget, so users can understand the relationship between them.
- The `RepeatableField` component has a good solution for labels where they aren't repeated on every row. We should mimic that if we aren't going to use the `RepeatableField` straight up.
- When edited, the datepickers jump around in the current layout (same as package custom coverage).
- We need to figure out how to surface server errors (toast notifications?).
- We probably want to refactor package custom coverage editing to match the behaviors here (should be quick).

## Screenshots
rkojc5bsbm

### DIFF
--- a/src/components/customer-resource-show.js
+++ b/src/components/customer-resource-show.js
@@ -17,13 +17,15 @@ import { isBookPublicationType, isValidCoverageList } from './utilities';
 import Modal from './modal';
 import styles from './styles.css';
 import CustomEmbargoForm from './custom-embargo';
+import CoverageForm from './coverage-form';
 
 export default class CustomerResourceShow extends Component {
   static propTypes = {
     model: PropTypes.object.isRequired,
     toggleSelected: PropTypes.func.isRequired,
     toggleHidden: PropTypes.func.isRequired,
-    customEmbargoSubmitted: PropTypes.func.isRequired
+    customEmbargoSubmitted: PropTypes.func.isRequired,
+    coverageSubmitted: PropTypes.func.isRequired
   };
 
   static contextTypes = {
@@ -68,7 +70,7 @@ export default class CustomerResourceShow extends Component {
   };
 
   render() {
-    let { model, customEmbargoSubmitted } = this.props;
+    let { model, customEmbargoSubmitted, coverageSubmitted } = this.props;
     let { router, queryParams } = this.context;
     let { showSelectionModal, resourceSelected, resourceHidden } = this.state;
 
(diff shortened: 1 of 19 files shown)


src/components/customer-resource-show.js
```diff
@@ -84,6 +86,14 @@ export default class CustomerResourceShow extends Component {
     let customEmbargoValue = model.customEmbargoPeriod && model.customEmbargoPeriod.embargoValue;
     let customEmbargoUnit = model.customEmbargoPeriod && model.customEmbargoPeriod.embargoUnit;
 
+    let customCoverages = model.customCoverages;
```
sadatay: this might be worthy of a comment as to what behavior this is here to support

sadatay:
looks good to me other than the minor nit, which is not a blocker :shipit:

PR #212, opened 2018-02-19
Get date format string from moment

cherewaty:
## Purpose
Part of https://issues.folio.org/browse/UIEH-33

Previously, we were showing the error message "Enter Date in 02/19/2018 format."

## Approach
Instead we can peel off the format string using `moment`. This is how the `stripes-components` `Datepicker` determines its placeholder.

New: "Enter Date in MM/DD/YYYY format."

Applied to package and package-title custom coverage editing.

## Screenshots
screen shot 2018-02-19 at 3 13 21 pm


cherewaty:
Yup, the context inconsistency is not great. In this case,`validate()` didn't have an obvious way to get the context, so I relied on a prop. We can address the inconsistency when we go back to refactor package custom coverage.

PR #214, opened 2018-02-19
Add package search filters for content type and selection status

wwilsman:
## Purpose
Adds package search filters for content type and selection status using the filters `filter[type]` and `filter[selected]` respectively.

## Approach
For the most part, copy and pasted the filter component for titles and just changed the relevant pieces to reflect the package filters. Same copy-paste method for the tests as well.

## Screenshots
gif

### DIFF
--- a/src/components/package-search-filters.js
+++ b/src/components/package-search-filters.js
@@ -30,12 +30,12 @@ export default function PackageSearchFilters(props) {
         defaultValue: 'all',
         options: [
           { label: 'All', value: 'all' },
-          { label: 'Aggregated Full Text', value: 'aggregatedfulltext' },
-          { label: 'Abstract and Index', value: 'abstractandindex' },
-          { label: 'E-Book', value: 'ebook' },
-          { label: 'E-Journal', value: 'ejournal' },
+          { label: 'Aggregated full text', value: 'aggregatedfulltext' },
+          { label: 'Abstract and index', value: 'abstractandindex' },
+          { label: 'E-book', value: 'ebook' },
+          { label: 'E-journal', value: 'ejournal' },
           { label: 'Print', value: 'print' },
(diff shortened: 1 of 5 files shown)


src/components/package-search-filters.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Accordion, FilterAccordionHeader } from '@folio/stripes-components/lib/Accordion';
+import RadioButton from '@folio/stripes-components/lib/RadioButton';
+
+const selectedFilters = [
```
cherewaty: Let's make all these filter labels Sentence case.

cherewaty:
Before we repeat this pattern again, it's probably a good time to match the `SearchAndSort` style: remove the `<hr>`s and indent the lists:

screen shot 2018-02-19 at 4 32 05 pm

wwilsman:
@cherewaty good point about matching styles. I'll open another PR tomorrow to match the styles in title filters, and just rebase & re-copy the component for this PR.

wwilsman:
Rebased with the common search-filters component. 

While fixing the mirage packages filter, I realized that we do not need to continue filtering if the model has already been filtered out of the results. So I applied this same logic to the title filters and removed the extra boolean variable.

src/components/package-search-filters.js
```diff
@@ -0,0 +1,45 @@
...
+          { label: 'Not selected', value: 'false' },
+          { label: 'Selected by EBSCO', value: 'ebsco' }
+        ]
+      }, {
+        name: 'type',
+        label: 'Content type',
+        defaultValue: 'all',
+        options: [
+          { label: 'All', value: 'all' },
+          { label: 'Aggregated Full Text', value: 'aggregatedfulltext' },
```
cherewaty: Sentence case

PR #215, opened 2018-02-20
Fix custom coverage datepicker row layout

cherewaty:
## Purpose
On both the package and the customer resource custom coverage forms, entering a value into either a start date or end date would cause the row's layout to recalculate. The datepicker inputs would then change size.

## Approach
The flex children that wrap the date pickers needed a `flex-basis` set (the third parameter of the `flex` shortcut).

## Learning
https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis

### Before
before

### After
after

### DIFF
--- a/src/components/coverage-form/coverage-form.css
+++ b/src/components/coverage-form/coverage-form.css
@@ -29,7 +29,7 @@
 
 .coverage-form-datepicker {
   align-self: flex-start;
-  flex: 1 1 auto;
+  flex: 1 1 200px;
   margin-right: 1em;
 }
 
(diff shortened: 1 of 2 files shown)


sadatay:
have you checked to see how this reacts to validation?  curious how it will jive in particular with the longer validation strings that will be coming soon

cherewaty:
@sadatay long validation messages wrap to multiple lines.
screen shot 2018-02-20 at 9 59 28 am

sadatay:
👍 noice

PR #221, opened 2018-02-21
Implement allow KB to add titles functionality

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-54, implement functionality for allowing knowledge bases to add titles at package level. For now, the only KB ui-eholdings interacts with is Ebsco KB.

## Approach
- Leverage the toggle-switch component for this functionality
- Toggle switch reflects the value of "allowKbToAddTitles" that we receive from mod-kb-ebsco
- User has the ability to toggle the switch and appropriate update is saved to "allowKbToAddTitles" in mod-kb-ebsco
- When user selects a package, then "allowKbToAddTitles" toggle should be set to true by default
- When user deselects a package, then "allowKbToAddTitles" should be reset to false
- Add associated unit tests for this functionality

#### TODOS and Open Questions
- [ ] When user refreshes the page, although the state of "allowKbToAddTitles" is true, I see an effect of this toggle go from false to true. At this point, seeing the response from backend, I do not see a reason why this should happen except that in redux, we are setting the default of "allowKbToAddTitles" to false. Elrick mentioned that this could be a VM thing too and that he would try to see if this happens on his machine using my branch. 
- Update to previous comment: Investigated to find out that RM API does not send "allowEbscoToAddTitles" in package list, so turns out that in mod-kb-ebsco we send that as null to ui-eholdings and redux state transitions from null to true which is why we see the transition of the toggle. 
To make that effect less prominent, we are rendering the toggle component only if the value is not null
- Once while testing mnaually, I was able to see that the toggle got stuck loading without updating its state with the response from the latest request, Jeffrey could not reproduce it and I could not reproduce it except that once. 

## Screenshots
allowtitles

### DIFF
--- a/src/components/package-show.js
+++ b/src/components/package-show.js
@@ -5,6 +5,7 @@ import IconButton from '@folio/stripes-components/lib/IconButton';
 import PaneMenu from '@folio/stripes-components/lib/PaneMenu';
 import Button from '@folio/stripes-components/lib/Button';
 import Layout from '@folio/stripes-components/lib/Layout';
+import Icon from '@folio/stripes-components/lib/Icon';
 
 import DetailsView from './details-view';
 import QueryList from './query-list';
@@ -187,7 +188,7 @@ export default class PackageShow extends Component {
                         />
                       </div>
                       ) : (
-                        <p>Loading...</p>
+                        <Icon icon="spinner-ellipsis" />
                       )}
                   </label>
 
--- a/src/components/package-show.js
+++ b/src/components/package-show.js
@@ -22,7 +22,8 @@ export default class PackageShow extends Component {
     fetchPackageTitles: PropTypes.func.isRequired,
     toggleSelected: PropTypes.func.isRequired,
     toggleHidden: PropTypes.func.isRequired,
-    customCoverageSubmitted: PropTypes.func.isRequired
+    customCoverageSubmitted: PropTypes.func.isRequired,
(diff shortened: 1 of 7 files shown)


src/components/package-show.js
```diff
@@ -162,6 +166,25 @@ export default class PackageShow extends Component {
 
                   <hr />
 
+                  <label
+                    data-test-eholdings-package-details-allow-add-new-titles
+                    htmlFor="package-details-toggle-allow-add-new-titles-switch"
+                  >
+                    <h4>
+                      {packageAllowedToAddTitles
```
cherewaty: This label isn't updating when I toggle. 
qjdzmnozlo

 I think it's because the label is driven off the internal component state. The other toggle switches rely on the model received in `props`. Switching this line to `model.allowKbToAddTitles` might work.

sduvvuri-ebsco: mm, interesting. Have not seen that before. Will test it out.

cherewaty:
I added a few commits:
- `yarn.lock` was a little out of date - still resolving to some old Reacts
- replaced the "Loading..." message with a animated icon
- added `allowKbToAddTitles` to the package mirage factory - should make future development on the package show a little easier

@sduvvuri-ebsco I wasn't able to reproduce the Redux issues you messaged me about - this seems to be working as expected for me with both production and mirage data.

sduvvuri-ebsco:
Ok, thanks for checking Jeffrey...

PR #222, opened 2018-02-21
Make common-component for search-filters

wwilsman:
## Purpose
Styles the search filters to match the stripes search-and-sort filter styles. 

## Approach
Since these styles will be used in other search filters, a common component was made

#### TODOS
- The current title-search-filters will be unnecessary once the API supports returning available filters via an endpoint. Instead, we can have the search-form render the search-filters component directly with the data returned from the API.

## Screenshots
*Before*
image

*After*
image

### DIFF
--- a/src/components/title-search-filters.js
+++ b/src/components/title-search-filters.js
@@ -16,32 +16,32 @@ export default function TitleSearchFilters(props) {
       searchType="titles"
       availableFilters={[{
         name: 'selected',
-        label: 'Selection Status',
+        label: 'Selection status',
         defaultValue: 'all',
         options: [
           { label: 'All', value: 'all' },
           { label: 'Selected', value: 'true' },
-          { label: 'Not Selected', value: 'false' },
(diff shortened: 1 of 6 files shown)


src/components/title-search-filters.js
```diff
@@ -1,92 +1,52 @@
...
+          { label: 'Not Selected', value: 'false' },
+          { label: 'Selected By EBSCO', value: 'ebsco' }
+        ]
+      }, {
+        name: 'type',
+        label: 'Publication Type',
+        defaultValue: 'all',
+        options: [
+          { label: 'All', value: 'all' },
+          { label: 'Audio Book', value: 'audiobook' },
```
cherewaty: Since we're in here, sentence case 😎

wwilsman: Totally forgot. Thanks!

wwilsman: Are we sentence casing the section headings `"Selection status"` and `"Publication type"` too? Or leave these title-cased?

PR #226, opened 2018-02-21
Fix eslint no-restricted-globals errors

cherewaty:
## Purpose
We had disabled linting on a few lines due to the `no-restricted-globals` rule.

## Learning
We can follow the rule just by changing `isNaN()` to `Number.isNaN()`.
https://github.com/airbnb/javascript#standard-library--isnan


skomorokh:
N.B. that is not always the case: `isNaN()` tests NaN-ness and coerces to number, `Number.isNaN()` only tests. To be drop-in equivalent you need `Number.isNaN(Number());`

...not relevant in this case as there's already an explicit coercion.

PR #227, opened 2018-02-22
Remove uglify-es resolution

cherewaty:
## Purpose
`uglify-es` has a new release; we don't have to resolve `yarn` to an un-buggy earlier release anymore.

## Approach
I also deleted `yarn.lock` and re-generated it from scratch to remove some duplicate package warnings. That brings in new `stripes-core` and `stripes-components`.

## Screenshots
ICONS!!! 🎉 

localhost_3000_eholdings_searchtype providers iphone 7

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,9 +44,6 @@
     "isomorphic-fetch": "^2.2.1",
     "redux-actions": "^2.2.1"
   },
-  "resolutions": {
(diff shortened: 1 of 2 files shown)


cherewaty:
Replaced with https://github.com/folio-org/ui-eholdings/pull/229 (will do the same upgrades, with the addition of an upgraded Stripes CLI).

PR #231, opened 2018-02-23
 Refactor customer-resource-show test to use BigTest

cherewaty:
## Purpose
We'd like to use [BigTest](https://github.com/thefrontside/bigtest) tools in `ui-eholdings` to improve the speed and reliability of the test suite.

## Approach
I refactored the `CustomerResourceShow` test to use a BigTest page object and assertions. When we've refactored all the page objects, we can move them back out of the `tests/pages/bigtest` directory to `/tests/pages`.

In the process, I discovered a regression where the subject list wasn't actually being printed out to the customer resource pages. I tried to lean more on comparing to explicit strings instead of generated mirage data (how that regression slipped in there in the first place).

#### TODOS and Open Questions
- [x] Need to merge `stripes-core` PR first that upgrades `babel-eslint`: https://github.com/folio-org/stripes-core/pull/225

## Learning
[Babel Legacy Decorator plugin](https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy) makes it so we can use the `@page` decorator. Should be moved into `stripes-core`.

### DIFF
--- a/tests/customer-resource-show-test.js
+++ b/tests/customer-resource-show-test.js
@@ -1,9 +1,8 @@
-/* global describe, beforeEach */
 import { expect } from 'chai';
-import it from './it-will';
+import { describe, beforeEach, it } from '@bigtest/mocha';
 
 import { describeApplication } from './helpers';
-import ResourcePage from './pages/customer-resource-show';
+import ResourcePage from './pages/bigtest/customer-resource-show';
 
 describeApplication('CustomerResourceShow', () => {
   let provider,
@@ -23,7 +22,9 @@ describeApplication('CustomerResourceShow', () => {
     });
 
     let title = this.server.create('title', {
-      publicationType: 'Streaming Video'
+      name: 'Best Title Ever',
+      publicationType: 'Streaming Video',
+      publisherName: 'Amazing Publisher'
     });
 
     title.identifiers = [
@@ -40,12 +41,18 @@ describeApplication('CustomerResourceShow', () => {
       this.server.create('contributor', { type: 'illustrator', contributor: 'Artist, John' })
     ].map(m => m.toJSON());
 
+    title.subjects = [
+      this.server.create('subject', { type: 'TLI', subject: 'Writing' }),
+      this.server.create('subject', { type: 'TLI', subject: 'Words' }),
+    ].map(m => m.toJSON());
+
     title.save();
 
     resource = this.server.create('customer-resource', {
       package: providerPackage,
       isSelected: false,
-      title
+      title,
+      url: 'frontside.io'
     });
   });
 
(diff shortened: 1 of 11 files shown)


tests/customer-resource-show-test.js
```diff
@@ -40,74 +41,78 @@ describeApplication('CustomerResourceShow', () => {
...
+      title,
+      url: 'frontside.io'
     });
   });
 
   describe('visiting the customer resource page', () => {
     beforeEach(function () {
       return this.visit(`/eholdings/customer-resources/${resource.titleId}`, () => {
-        expect(ResourcePage.$root).to.exist;
+        expect(ResourcePage).to.exist;
```
wwilsman: I now realize that there isn't any docs around this on the `@bigtest/interaction` repo, but `ResourcePage.$root` is still the correct way to verify that the root element of the resource page exists.

Also `.to.exist` does not work as you may think here. `chai-jquery` [overrides the built-in chai assertion](https://github.com/chaijs/chai-jquery#exist), and if it isn't a jQuery element, it will call the original implementation instead.

The original implementation of `.to.exist` should work for the bigtest page object because it will throw an error when `.$root` doesn't exist. But for other jQuery matchers, we'll want to use something like [chai-dom](https://github.com/nathanboktae/chai-dom) instead.

tests/customer-resource-show-test.js
```diff
@@ -166,19 +172,19 @@ describeApplication('CustomerResourceShow', () => {
...
     });
 
     it('displays the title name', () => {
-      expect(ResourcePage.titleName).to.equal(resource.title.name);
+      expect(ResourcePage.titleName).to.equal('Best Title Ever');
     });
 
-    it.still('does not display a content type', () => {
-      expect(ResourcePage.contentType).to.equal('');
+    it('does not display a content type', () => {
```
wwilsman: We'll want `it.always` now instead of `it.still`.

Basically, the page will be rendered without the content type element because the request is still out. So this test will pass immediately while there's a loading indicator on the page.

`it.always` will make sure that there won't be a content type for at least 100ms

tests/customer-resource-show-test.js
```diff
@@ -166,19 +172,19 @@ describeApplication('CustomerResourceShow', () => {
...
+      expect(ResourcePage.hasContentType).to.be.false;
     });
 
-    it.still('does not display a publication type', () => {
+    it('does not display a publication type', () => {
```
wwilsman: Same thing here with `it.always` although maybe it doesn't apply in this case because it ensures that there _is_ a publication type element in the DOM, but it is empty. My concern would be if the publication type element renders at least once before any actual text content is rendered inside. In that case it would pass before the text content gets rendered.

tests/customer-resource-show-test.js
cherewaty: Cool, went back to checking for `ResourcePage.$root` to exist.

tests/customer-resource-show-test.js
cherewaty: Using `it.always` now

tests/customer-resource-show-test.js
cherewaty: Changed the logic in the view to conditionally show the Publication type div, then switched this to an `it.always` that checks for `hasPublicationType` (using `isPresent`).

cherewaty:
Getting flaky timeouts with `it.always()`: https://travis-ci.org/folio-org/ui-eholdings/builds/346465220

tests/customer-resource-show-test.js
```diff
@@ -237,4 +244,4 @@ describeApplication('CustomerResourceShow', () => {
       expect(ResourcePage.hasErrors).to.be.true;
     });
   });
-});
+}, describe);
```
cherewaty: @wwilsman is this right/necessary? We want to use `describe` from `@bigtest/mocha` instead of `window.describe`?

wwilsman:
I suspect an issue we should take upstream to `@bigtest/mocha`.

We're relying on mocha's `.timeout()` to set the convergence timeout, that's why we see `Timeout of 100ms exceeded`. The convergence _should_ still finish or throw before then, but there may be some timing issue where the convergence doesn't start right away. We should maybe not rely on mocha's timeout for `always` convergences.

I'll open issues upstream, but we can skip these for now with `it.always.skip`. I'll try to reproduce this in `@bigtest/mocha`'s own test suite tonight 🤞

tests/customer-resource-show-test.js
wwilsman: `describe` from `@bigtest/mocha` is forwarded straight from mocha. Which is actually `window.describe` in this case, so this shouldn't be necessary.

PR #232, opened 2018-02-26
Get global styles to appear in karma runs

cherewaty:
## Purpose
Karma runs haven't been loading in the `global.css` from `stripes-components`. It wasn't that big of a deal until I realized an absolutely positioned `Paneset` was inside of a relatively-positioned div with no height set. That causes the parent to have a calculated height of 0. The contents of the `Paneset` (the eHoldings settings form) weren't visible in Karma runs, even though the contents were in the DOM.
https://issues.folio.org/browse/UIEH-111

## Approach
This feels hacky (a superior strategy would involve changes in the `stripes` ecosystem), but it solves the immediate problem that the eHoldings settings form wasn't visible.

This still doesn't load Source Sans Pro, so the font in karma falls back to the system sans serif.

## Screenshot
localhost_9876__id 26575445

### DIFF
--- a/src/routes/application/application.css
+++ b/src/routes/application/application.css
@@ -1,3 +1,6 @@
+/* Supposed to be coming from webpack entry, but we explicitly import
+   here so karma-webpack will pick it up.
+ */
 @import '@folio/stripes-components/lib/global';
 
 .eholdings-application {
--- a/src/routes/application/application.css
+++ b/src/routes/application/application.css
@@ -1,3 +1,5 @@
+@import '@folio/stripes-components/lib/global';
+
 .eholdings-application {
   width: 100%;
 }
--- a/tests/helpers.js
+++ b/tests/helpers.js
@@ -36,6 +36,7 @@ export function describeApplication(name, setup, describe = window.describe) {
     beforeEach(function () {
       rootElement = document.createElement('div');
       rootElement.id = 'react-testing';
+      rootElement.style = 'height: 100%;';
       document.body.appendChild(rootElement);
 
       this.server = startMirage(setup.scenarios);


cherewaty:
cc @ttamj if you have any thoughts on a better way to do this, I'm all ears.

ttamj:
The discrepancy may be in the way karma-webpack works.  We define the font and global.css as entry points in `webpack.config.base.js`:
```
entry: [
  'typeface-source-sans-pro',
  '@folio/stripes-components/lib/global.css',
  path.join(__dirname, 'src', 'index'),
],
```
I can see these values, along with the rest of the webpack configuration, are getting passed to karma.  However, karma-weback's usage notes indicate the following: 
```
webpack: {
  // karma watches the test entry points
  // (you don't need to specify the entry option)
  // webpack watches dependencies

  // webpack configuration
},
```
If karma-webpack is relying only on the Karma test's entry points rather then the webpack entry points, then perhaps the import for the typeface and global.css belong in either `/tests/harness.js` or `/tests/helpers.js` where we are wrapping/prepping the Stripes application.

src/routes/application/application.css
```diff
@@ -1,3 +1,8 @@
+/* Supposed to be coming from webpack entry, but we explicitly import
+   here so karma-webpack will pick it up.
+ */
+@import '@folio/stripes-components/lib/global';
```
wwilsman: I'm curious... I think maybe in production this would duplicate the styles? They're included once by webpack, but if they're imported here, I don't think postcss will scrub them out properly (because it doesn't know they're being added by webpack).

I agree with @ttamj that this could probably go in `/tests/harness.js`. We have a css loader, so `import '@folio/stripes-components/lib/global.css'` _should_ work.

wwilsman: I assume we could probably do the same with `import 'typeface-source-sans-pro'`

cherewaty:
Replaced by https://github.com/folio-org/ui-eholdings/pull/262

PR #235, opened 2018-02-26
Package title custom coverage validation

sadatay:
## Purpose
Package/Titles require more intensive custom coverage validation than
Package-level custom coverage.  Since Package/Titles can have multiple
custom coverage ranges, additional care must be taken to prevent
overlap among these ranges.  Additionally, Package/Titles must respect
the custom coverage range (if any) of their parent package.  If a
package has a custom coverage range set, all titles within that
package are relegated to custom coverage within the package's range.

## Approach
These new validations, being significantly more complex than those on
packages, crowded the `validate` method, and so all validations were
pulled out into functions and called in sequence during validation.

#### TODOS and Open Questions
- [ ] Kill `moment` deprecation warnings (again)
- [ ] Add screenshots

## Screenshots
_Coming soon_


sadatay:
replaced by https://github.com/folio-org/ui-eholdings/pull/237

PR #237, opened 2018-02-27
Package/Title-Specific Custom Coverage Validations

sadatay:
## Purpose
Package/Titles require more intensive custom coverage validation than
Package-level custom coverage.  Since Package/Titles can have multiple
custom coverage ranges, additional care must be taken to prevent
overlap among these ranges.  Additionally, Package/Titles must respect
the custom coverage range (if any) of their parent package.  If a
package has a custom coverage range set, all titles within that
package are relegated to custom coverage within the package's range.

## Approach
These new validations, being significantly more complex than those on
packages, crowded the `validate` method, and so all validations were
pulled out into functions and called in sequence during validation.

**NOTE**
_This replaces https://github.com/folio-org/ui-eholdings/pull/235.  The functionality is the same, but I manually patched it back onto master.  There were some weird issues with the remote tracking in the original branch which was making it impossible to rebase / squash commits._

### Preventing Range Overlap
overlap

### Restricting to Package Range
package-coverage

### DIFF
--- a/src/components/customer-resource-coverage/customer-resource-coverage.js
+++ b/src/components/customer-resource-coverage/customer-resource-coverage.js
@@ -1,42 +1,46 @@
 import React, { Component } from 'react';
 import { Field, FieldArray, reduxForm } from 'redux-form';
 import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
 import moment from 'moment';
 import isEqual from 'lodash/isEqual';
-import classNames from 'classnames/bind';
 
 import Datepicker from '@folio/stripes-components/lib/Datepicker';
 import Button from '@folio/stripes-components/lib/Button';
 import Icon from '@folio/stripes-components/lib/Icon';
 import IconButton from '@folio/stripes-components/lib/IconButton';
-import CoverageDateList from '../coverage-date-list';
 import KeyValueLabel from '../key-value-label';
+
+import CoverageDateList from '../coverage-date-list';
 import styles from './customer-resource-coverage.css';
+import { formatISODateWithoutTime } from '../utilities';
 
 const cx = classNames.bind(styles);
 
 class CustomerResourceCoverage extends Component {
   static propTypes = {
     initialValues: PropTypes.shape({
-      customCoverages: PropTypes.array,
+      customCoverages: PropTypes.array
     }).isRequired,
+    packageCoverage: PropTypes.object, // eslint-disable-line react/no-unused-prop-types
     onSubmit: PropTypes.func.isRequired,
     handleSubmit: PropTypes.func,
     pristine: PropTypes.bool,
     isPending: PropTypes.bool,
     initialize: PropTypes.func,
-    locale: PropTypes.string // eslint-disable-line react/no-unused-prop-types
+    locale: PropTypes.string, // eslint-disable-line react/no-unused-prop-types
+    intl: PropTypes.object // eslint-disable-line react/no-unused-prop-types
   };
 
   state = {
-    isEditing: false,
+    isEditing: false
   };
 
   componentWillReceiveProps(nextProps) {
     let wasPending = this.props.isPending && !nextProps.isPending;
-    let needsUpdate = !isEqual(this.props.initialValues, nextProps.initialValues);
+    let needsUpdate = !isEqual(this.props.initialValues.customCoverages, nextProps.initialValues.customCoverages);
 
-    if (wasPending || needsUpdate) {
+    if (wasPending && needsUpdate) {
       this.setState({ isEditing: false });
     }
   }
@@ -54,12 +58,13 @@ class CustomerResourceCoverage extends Component {
     this.props.initialize(this.props.initialValues);
   }
 
-  renderDatepicker = ({ input, label, meta }) => {
+  renderDatepicker = ({ input, label, meta, id }) => {
     return (
       <Datepicker
         label={label}
         input={input}
         meta={meta}
+        id={id}
       />
     );
   }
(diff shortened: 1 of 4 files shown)


src/components/customer-resource-coverage/customer-resource-coverage.js
```diff
@@ -1,42 +1,46 @@
...
+    isEditing: false
   };
 
   componentWillReceiveProps(nextProps) {
     let wasPending = this.props.isPending && !nextProps.isPending;
-    let needsUpdate = !isEqual(this.props.initialValues, nextProps.initialValues);
+    let needsUpdate = !isEqual(this.props.initialValues.customCoverages, nextProps.initialValues.customCoverages);
 
-    if (wasPending || needsUpdate) {
+    if (wasPending && needsUpdate) {
```
cherewaty: Is this the right logic now? It's still an `or` the couple of other times this pops up.

src/components/customer-resource-coverage/customer-resource-coverage.js
```diff
@@ -56,11 +60,13 @@ class CustomerResourceCoverage extends Component {
 
   renderDatepicker = ({ input, label, meta }) => {
     return (
-      <Datepicker
-        label={label}
-        input={input}
-        meta={meta}
-      />
+      <div>
```
cherewaty: Shouldn't need this `<div>`.

src/components/customer-resource-coverage/customer-resource-coverage.js
```diff
@@ -226,21 +232,147 @@ class CustomerResourceCoverage extends Component {
...
+/**
+ * Validator to check that no date ranges overlap or are identical
+ * @param {} dateRange - coverage date range to validate
+ * @param {} customCoverages - all custom coverage ranges present in edit form
+ * @param {} index - index in the field array indicating which coverage range is
+ * presently being considered
+ * @param {} intl - object containing locale-specific data & formatting
+ * @returns {} - an error object if errors are found, or `false` otherwise
+ */
+const validateNoRangeOverlaps = (dateRange, customCoverages, index, intl) => {
```
cherewaty: A+ function comments 👏

tests/pages/customer-resource-coverage.js
```diff
@@ -56,8 +72,8 @@ function createRowObject(element) {
     },
 
     clearBeginDate() {
-      let $clearButton = $('[data-test-eholdings-package-details-custom-begin-coverage]')
-        .find('button[id^=datepicker-clear-button]');
+      let $clearButton = $('[data-test-eholdings-coverage-form-date-range-begin]')
+        .find('button[id^=datepicker-clear-button-undefined]');
```
cherewaty: Is `datepicker-clear-button-undefined` really the id? 😨

src/components/customer-resource-coverage/customer-resource-coverage.js
sadatay: i'll investigate the logic here as compared to the other places we use this pattern, but iirc one of our pairing sessions included changing this to `||`.  i'll try to suss out why

tests/pages/customer-resource-coverage.js
sadatay: i think this is why we do a fuzzy search instead of a direct search for id - i added the `undefined` back while debugging tests and forgot to go back and yank it out.  furthermore, it's possible this changed after the rebase.  i'll do some digging

src/components/customer-resource-coverage/customer-resource-coverage.js
sadatay: verified: `||` causes that behavior where clicking any button on the form immediately closes it.  `&&` is correct

src/components/customer-resource-coverage/customer-resource-coverage.js
```diff
@@ -226,21 +233,147 @@ class CustomerResourceCoverage extends Component {
...
+ * Validator to ensure all coverage ranges are within the parent package's
+ * custom coverage range if one is present
+ * @param {} dateRange - coverage date range to validate
+ * @param {} packageCoverage - parent package's custom coverage range
+ * @param {} intl - object containing locale-specific data & formatting
+ * @returns {} - an error object if errors are found, or `false` otherwise
+ */
+const validateWithinPackageRange = (dateRange, packageCoverage, intl) => {
+  // javascript/moment has no mechanism for "infinite", so we
+  // use an absurd future date to represent the concept of "present"
```
elrickvm: Is this descriptive enough for someone new to code based? Like will they understand that 'present' is Folio terms is an open-ended end-date that is rolling? Is there documentation in Folio where you read up in the concept of 'present' per-say? Just a thought.

sadatay: i think it's descriptive enough for the code it's referencing or, as you put it, for someone new to the codebase.  for someone new to FOLIO in general, probably not.  but i'd argue this isn't the right place to delineate what "present" means in this context - we'd probably want to do that in documentation, in user-facing information prompts or (if done as a code comment) somewhere more general than this validation method.

that's just my feeling on it though - since it's a comment i think more information wouldn't be likely to hurt anything.  if you wanna reword it just let me know - you wrote the line anyway so you should probably get first dibs at describing it :) . looks like i got some linting errors to fix before merging anyway

src/components/customer-resource-coverage/customer-resource-coverage.js
sadatay: `C-c C-r i d` 🐇 🎩

src/components/customer-resource-coverage/customer-resource-coverage.js
elrickvm: This is fine with me.

PR #238, opened 2018-02-27
Refactor settings test to use BigTest

cherewaty:
Moves application back end error checking and settings form tests to BigTest page objects and assertions.

Gets rid of the custom eHoldings `fillIn()` helper in favor of the BigTest `fillable()`.


tests/pages/bigtest/settings.js
```diff
@@ -0,0 +1,38 @@
...
+  property,
+  text,
+  value
+} from '@bigtest/interaction';
+
+let hasClassBeginningWith = (className, selector) => {
+  return computed(function () {
+    return this.$(selector).className.includes(className);
+  });
+};
```
wwilsman: 🔥

tests/backend-configuration-test.js
```diff
@@ -112,19 +109,19 @@ describeApplication('With valid backend configuration', () => {
...
     });
 
     it('field for the ebsco RM API key appears with text as password hidden', () => {
-      expect(SettingsPage.$apiKeyField).to.have.prop('type', 'password');
+      expect(SettingsPage.apiKeyInputType).to.equal('password');
     });
 
-    it.still('does not have visible form action buttons', () => {
-      expect(SettingsPage.$actions).to.not.exist;
+    it('does not have visible form action buttons', () => {
```
wwilsman: This should be `always`, but that flakiness we found before means it might have to be skipped 😞

tests/backend-configuration-test.js
```diff
@@ -190,40 +184,41 @@ describeApplication('With valid backend configuration', () => {
...
           }, 500);
 
-          SettingsPage.save();
+          return SettingsPage.save();
         });
 
-        it.still('still shows the form action buttons', () => {
-          expect(SettingsPage.$actions).to.exist;
+        it('still shows the form action buttons', () => {
+          expect(SettingsPage.hasVisibleActions).to.be.true;
```
wwilsman: This was previously `true` in another test outside of this describe, so we should assert that this one is `always` true, just in case it disappears milliseconds after saving changes.  ... but again with the flakiness of `always` currently, we probably need to skip it 😞

cherewaty: `it.always.skip` added

tests/pages/bigtest/settings.js
cherewaty: This was easier than the options discussed in https://github.com/thefrontside/bigtest/issues/52, since the page object doesn't have to pass through the webpack `style-loader`.

PR #239, opened 2018-02-28
Remove detail pane upon conducting new search

elrickvm:
## Purpose
After conducting a search users are presented with a list of results and detail view associated with clicking on an item from the results list. When a user conducts a new search removing the detail view that contains the old  information from the last clicked item makes it a more pleasing user experience

## Approach
Remove the third pane after a new search it conducted in any Provider, Package, or Title search

## Jira
UIEH-104

### DIFF
--- a/src/routes/search.js
+++ b/src/routes/search.js
@@ -23,7 +23,8 @@ class SearchRoute extends Component {
     }).isRequired,
     history: PropTypes.shape({
       push: PropTypes.func.isRequired,
-      replace: PropTypes.func.isRequired
+      replace: PropTypes.func.isRequired,
+      location: PropTypes.object
     }).isRequired,
     searchProviders: PropTypes.func.isRequired,
     searchPackages: PropTypes.func.isRequired,
@@ -150,6 +151,11 @@ class SearchRoute extends Component {
     // if the new query is different from our location, update the location
     if (qs.stringify(params) !== qs.stringify(this.state.params)) {
       let url = this.buildSearchUrl(location.pathname, params);
+
+      // if new query is different from current query this will close detail pane
+      if (params.q !== this.state.params.q) {
+        url = this.buildSearchUrl('/eholdings', params);
+      }
       // if only the filters have changed, just replace the current location
       if (params.q === this.state.params.q) {
         history.replace(url);
--- a/tests/package-search-test.js
+++ b/tests/package-search-test.js
@@ -76,6 +76,23 @@ describeApplication('PackageSearch', () => {
         expect(PackageSearchPage.$backButton).to.not.exist;
       });
 
+      describe('conducting a new search', () => {
+        beforeEach(() => {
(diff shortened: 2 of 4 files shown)


src/components/search-paneset/search-paneset.js
```diff
@@ -37,17 +38,34 @@ export default class SearchPaneset extends React.Component {
...
     }
   }
 
+  componentDidUpdate() {
+    this.shouldClosePane();
+  }
+
+  shouldClosePane = () => {
+    if (this.state.closePane) {
+      this.closePreview();
```
wwilsman: I'm curious about the browser history at this point. If I understand correctly the flow is as follows: a new search happens and the URL is updated via a push, this component receives a new search string and calls this method, `closePreview` then pushes to the browser history thus closing the preview pane.

It seems like there would be two new entries in the browser history. So if you click the back button after a new search, we will get the same search but with the preview pane visible. If we click back again, then we get the old search with the same preview pane visible.

Also, what about when there isn't a preview pane open? If we create a new search without clicking on anything to preview I think those two browser history entries would still be there. Essentially requiring you to click back twice to actually go back to the old search.

I could be completely wrong about this, but that's what seems like is happening following through the code. Could we verify that this isn't the case?

elrickvm: Great insight here Wil. I will check out all this areas that you pointed out to inspect what is happening.

src/routes/search.js
```diff
@@ -265,6 +271,7 @@ class SearchRoute extends Component {
             detailsView={!hideDetails && children}
             totalResults={results.length}
             isLoading={!results.hasLoaded}
+            searchString={params.q}
```
sadatay: i don't think we need this prop anymore

src/routes/search.js
```diff
@@ -150,6 +151,11 @@ class SearchRoute extends Component {
     // if the new query is different from our location, update the location
     if (qs.stringify(params) !== qs.stringify(this.state.params)) {
       let url = this.buildSearchUrl(location.pathname, params);
+
+      // if new query is different from current query this will close detail pane
+      if (params.q !== this.state.params.q) {
+        url = this.buildSearchUrl('/eholdings', params);
+      }
```
sadatay: 🎵 _Can it be that it was all so simple then?_ 🎵

src/components/search-paneset/search-paneset.js
```diff
@@ -61,6 +61,7 @@ export default class SearchPaneset extends React.Component {
     });
   };
 
+
```
sadatay: yoink

src/routes/search.js
elrickvm: Correct!

src/routes/search.js
elrickvm: ...But we rather make it hard. But writing code is like a battle and we all end up with scars...

PR #240, opened 2018-02-28
Highlight selected provider/package/title

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-172, when user conducts a search and selects a result in the provider/package/title result list pane, the selected result should be highlighted within the result list pane. 

## Approach
- Modify index.js to pass 'type' and 'id' of the provider/package/title being selected in search route. 
- Within search list, compare the 'type' and 'id' to identify which item has been selected
- If an item has been selected, set its corresponding "active" prop type to true; else false
- Modify css of the selected item to have a gray background color, white text and the hover color to match the background color
- Also, add associated unit tests

## Learning
- CSS Selectors and Combinators
- react-cx

## Screenshots
highlightselection

### DIFF
--- a/src/components/package-list-item/package-list-item.js
+++ b/src/components/package-list-item/package-list-item.js
@@ -10,6 +10,7 @@ const cx = classNames.bind(styles);
 export default function PackageListItem({
   item,
   link,
+  active,
   showTitleCount,
   showProviderName,
   packageName
@@ -24,7 +25,13 @@ export default function PackageListItem({
       })}
     />
   ) : (
-    <Link to={link} className={styles.item}>
+    <Link
+      data-test-eholdings-package-list-item-active
+      to={link}
+      className={cx('item', {
+        'is-selected': active
+      })}
+    >
       <h5 data-test-eholdings-package-list-item-name>
         {packageName || item.name}
       </h5>
@@ -78,6 +85,7 @@ PackageListItem.propTypes = {
     PropTypes.string,
     PropTypes.object
   ]),
+  active: PropTypes.bool,
   showTitleCount: PropTypes.bool,
   showProviderName: PropTypes.bool,
   packageName: PropTypes.string
--- a/src/components/package-search-list.js
+++ b/src/components/package-search-list.js
@@ -10,7 +10,10 @@ export default function PackageSearchList({
   collection,
   fetch,
   onUpdateOffset
-}) {
+}, { router }) {
+  let { params: routeParams } = router.route.match;
+  let isPackagePage = routeParams.type === 'packages';
+
   return (
     <QueryList
       type="packages"
@@ -28,6 +31,7 @@ export default function PackageSearchList({
             pathname: `/eholdings/packages/${item.content.id}`,
             search: location.search
           }}
+          active={item.content && isPackagePage && routeParams.id === item.content.id}
         />
       )}
     />
@@ -40,4 +44,8 @@ PackageSearchList.propTypes = {
   collection: PropTypes.object.isRequired,
   fetch: PropTypes.func.isRequired,
   onUpdateOffset: PropTypes.func.isRequired
+};
+
+PackageSearchList.contextTypes = {
+  router: PropTypes.object
 };
--- a/src/components/provider-list-item/provider-list-item.js
+++ b/src/components/provider-list-item/provider-list-item.js
@@ -1,14 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
 
 import styles from './provider-list-item.css';
 import Link from '../link';
 
-export default function ProviderListItem({ item, link }, { intl }) {
+const cx = classNames.bind(styles);
+
+export default function ProviderListItem({ item, link, active }, { intl }) {
   return !item ? (
     <div className={styles.skeleton} />
   ) : (
-    <Link to={link} className={styles.item}>
+    <Link
+      data-test-eholdings-provider-list-item-active
+      to={link}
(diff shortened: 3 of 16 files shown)


src/components/provider-search-list.js
```diff
@@ -27,6 +27,7 @@ export default function ProviderSearchList({
             pathname: `/eholdings/providers/${item.content.id}`,
             search: location.search
           }}
+          active={(router.route.match.params.type === 'providers') && (router.route.match.params.id === (item && item.content && item.content.id))}
```
wwilsman: Same comment here about the length

src/components/title-list-item/title-list-item.js
```diff
@@ -21,7 +22,12 @@ export default function TitleListItem({
       })}
     />
   ) : (
-    <Link to={link} className={styles.item}>
+    <Link
+      to={link}
+      className={cx('item', {
+      'is-selected': active
+    })}
```
wwilsman: Odd formatting here too

src/components/title-search-list.js
```diff
@@ -28,6 +28,7 @@ export default function TitleSearchList({
             pathname: `/eholdings/titles/${item.content.id}`,
             search: location.search
           }}
+          active={(router.route.match.params.type === 'titles') && (router.route.match.params.id === (item && item.content && item.content.id))}
```
wwilsman: Length here as well

src/components/package-search-list.js
```diff
@@ -28,6 +28,7 @@ export default function PackageSearchList({
             pathname: `/eholdings/packages/${item.content.id}`,
             search: location.search
           }}
+          active={(router.route.match.params.type === 'packages') && (router.route.match.params.id === (item && item.content && item.content.id))}
```
wwilsman: This is a little long, could we pull this (or part of it) into a variable above the `return`?

Something like
```js
let { params: routeParams } = router.route.match;
let isPackagePage = routeParams.type === 'packages';

...

active={(isPackagePage && routeParams.id == (item.content && item.content.id)}
```

Also, `item` will always be provided, so we just need to verify `item.content` exists, no need to verify `item`.

src/components/package-list-item/package-list-item.js
```diff
@@ -24,7 +25,12 @@ export default function PackageListItem({
       })}
     />
   ) : (
-    <Link to={link} className={styles.item}>
+    <Link
+      to={link}
+      className={cx('item', {
+      'is-selected': active
+    })}
```
wwilsman: Looks like there was some odd formatting that happened here

src/components/provider-list-item/provider-list-item.js
```diff
@@ -1,14 +1,22 @@
...
+export default function ProviderListItem({ item, link, active }, { intl }) {
   return !item ? (
     <div className={styles.skeleton} />
   ) : (
-    <Link to={link} className={styles.item}>
+    <Link
+      to={link}
+      className={cx('item', {
+      'is-selected': active
+    })}
```
wwilsman: Formatting

wwilsman:
To answer your test question: Yes we should definitely write some tests for this 🙂

Could be as simple as 
```
it('has an active state', () => {
  expect(ProvidersSearchPage.providerList[0].isActive).to.be.true
});
```

and the `isActive` could be added to the `createProviderObject` in `tests/pages/provider-search.js`

sduvvuri-ebsco:
Thanks for getting me started on the tests Wil.. I had no idea how to test this functionality...

src/components/title-list-item/title-list-item.js
sduvvuri-ebsco: I thought this format is more readable...I can change it back to a one-liner

wwilsman: I'm referring to the spacing.

```
      className={cx('item', {
      'is-selected': active
    })}
```

vs



```
      className={cx('item', {
        'is-selected': active
      })}
```

wwilsman: My editor does this to me all the time and I have to manually align them with spaces

src/components/package-search-list.js
```diff
@@ -28,6 +31,7 @@ export default function PackageSearchList({
             pathname: `/eholdings/packages/${item.content.id}`,
             search: location.search
           }}
+          active={isPackagePage && routeParams.id === (item.content && item.content.id)}
```
wwilsman: One last thing I just thought of:

When we aren't viewing a preview pane, `routeParams.id` might be `undefined`. And if `item.content` is also `undefined` that would cause this to be `true`.

I think moving `item.content &&` to the beginning of this prop should suffice:

```
item.content && isPackagePage && routerParams.id === item.content.id
```

tests/pages/package-search.js
```diff
@@ -25,6 +25,10 @@ function createPackageObject(element) {
 
     get isSelected() {
       return $scope.find('[data-test-eholdings-package-list-item-selected]').text().toLowerCase() === 'selected';
+    },
+
+    get isActive() {
+      return $('[data-test-eholdings-package-list-item-active]').find('a').attr('class').indexOf('is-selected---') !== -1;
```
wwilsman: I don't think we need this additional element. It looks like when this object is created it uses `$searchResultsItems` below, which selects `[data-test-query-list="packages"] li a` elements. I think you could remove this extra div and just use `$scope.attr('class').indexOf('is-selected--')`.

sduvvuri-ebsco:
Thanks for all of your suggestions Wil. made those changes..

PR #241, opened 2018-02-28
Add confirmation when leaving a customer-resource with unsaved changes

wwilsman:
## Purpose
When editing a customer-resource's embargo or coverage the user should be prompted to confirm navigation when attempting to navigate away from the page.

## Approach
React Router comes with a `Prompt` component which will block navigation internally with `history.block(promptMessage)`. By default, it uses `window.prompt`. When provided with a function, it will evaluate that function to generate a prompt message. When `promptMessage` is `false`, navigation will still be blocked but there will be no `window.prompt`. By using a function that always returns `false` we can create a custom component that will display a modal while still blocking navigation.

The `CustomerResourceCoverage` and `CustomEmbargo` components have been updated to support having their editable state controlled from the parent component. If the parent component passes an `onEdit` prop, then that component will be responsible for providing an `isEditable` property as well (which defaults to false).

## Learning
- [Blocking history transitions](https://github.com/ReactTraining/history#blocking-transitions)

## Screenshots
blocking-transitions

### DIFF
--- a/src/components/customer-resource-show.js
+++ b/src/components/customer-resource-show.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import { Prompt } from 'react-router-dom';
 
 import Button from '@folio/stripes-components/lib/Button';
 import Layout from '@folio/stripes-components/lib/Layout';
@@ -28,15 +29,14 @@ export default class CustomerResourceShow extends Component {
 
   static contextTypes = {
     locale: PropTypes.string,
-    intl: PropTypes.object,
-    router: PropTypes.object,
-    queryParams: PropTypes.object
+    intl: PropTypes.object
   };
 
   state = {
     showSelectionModal: false,
     resourceSelected: this.props.model.isSelected,
-    resourceHidden: this.props.model.visibilityData.isHidden
+    resourceHidden: this.props.model.visibilityData.isHidden,
+    isEditing: false
   };
 
   componentWillReceiveProps({ model }) {
@@ -69,10 +69,25 @@ export default class CustomerResourceShow extends Component {
     });
   };
 
+  // when there are no fields being edited this is equal to 0; when
+  // any number of fields are being edited this is greater than 0;
+  // this number is stored on the instance so that the state will only
+  // update when this count is equal to or greater than 0
+  editCount = 0;
+
+  handleEditing = (edit) => {
+    this.editCount = this.editCount + (edit ? 1 : -1);
+    let isEditing = this.editCount > 0;
+
+    if (isEditing !== this.state.isEditing) {
+      this.setState({ isEditing });
+    }
+  };
+
   render() {
     let { model, customEmbargoSubmitted, coverageSubmitted } = this.props;
     let { locale, intl } = this.context;
-    let { showSelectionModal, resourceSelected, resourceHidden } = this.state;
+    let { showSelectionModal, resourceSelected, resourceHidden, isEditing } = this.state;
 
     let hasManagedCoverages = model.managedCoverages.length > 0 &&
       isValidCoverageList(model.managedCoverages);
@@ -240,6 +255,7 @@ export default class CustomerResourceShow extends Component {
                   <CustomerResourceCoverage
                     initialValues={{ customCoverages }}
                     packageCoverage={model.package.customCoverage}
+                    onEdit={this.handleEditing}
                     onSubmit={coverageSubmitted}
                     isPending={model.update.isPending && 'customCoverages' in model.update.changedAttributes}
                     locale={locale}
@@ -250,6 +266,7 @@ export default class CustomerResourceShow extends Component {
 
                   <CustomEmbargoForm
                     initialValues={{ customEmbargoValue, customEmbargoUnit }}
+                    onEdit={this.handleEditing}
                     onSubmit={customEmbargoSubmitted}
                     isPending={model.update.isPending && 'customEmbargoPeriod' in model.update.changedAttributes}
                   />
@@ -311,6 +328,14 @@ export default class CustomerResourceShow extends Component {
             )
           }
         </Modal>
+
+        <Prompt
+          when={isEditing}
+          message={
+            'There are currently unsaved changes on this page.\n\n' +
+            'These changes will be lost, continue?'
+          }
+        />
       </div>
     );
   }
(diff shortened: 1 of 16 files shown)


tests/customer-resource-embargo-test.js
```diff
@@ -80,6 +80,44 @@ describeApplication('CustomerResourceEmbargo', () => {
         expect(CustomerResourceShowPage.$customEmbargoForm).to.exist;
       });
 
+      describe('then trying to navigate away', () => {
```
sadatay: since this impacts every editable field, and since this will eventually also apply to packages (and even custom resources/packages), i might like to see these tests pulled out into a separate test file instead of dropped into each field-specific test file.  the setup and execution would essentially be the same, but this way it's all in one easy-to-find place.  plus that would allow us to segregate the `sinon` stuff out of the rest of the suite.

src/components/customer-resource-show.js
```diff
@@ -70,10 +72,25 @@ export default class CustomerResourceShow extends Component {
...
   };
 
+  // when there are no fields being edited this is equal to 0; when
+  // any number of fields are being edited this is greater than 0;
+  // this number is stored on the instance so that the state will only
+  // update when this count is equal to or greater than 0
+  editCount = 0;
+
+  handleEditing = (edit) => {
+    this.editCount = this.editCount + (edit ? 1 : -1);
```
sadatay: `this.editCount += (edit ? 1 : -1)`

src/components/customer-resource-show.js
```diff
@@ -70,10 +72,25 @@ export default class CustomerResourceShow extends Component {
...
+    let isEditing = this.editCount > 0;
+
+    if (isEditing !== this.state.isEditing) {
```
sadatay: Why `this.editCount > 0` and not just `!!this.editCount`?.  Is there a case where `editCount` could be negative?  

This sent me into some research on javascript's lack of a logical `xor`, because i thought it might be able to make the logic here more concise.  Long story short, you can use bitwise `xor` to make this a one-liner, but what you have is way clearer.  For entertainment's sake, here it is:

```
this.setState({ isEditing: !!(editCount ^= this.state.isEditing) })
```

src/components/customer-resource-show.js
```diff
@@ -70,10 +72,25 @@ export default class CustomerResourceShow extends Component {
     });
   };
 
+  // when there are no fields being edited this is equal to 0; when
+  // any number of fields are being edited this is greater than 0;
+  // this number is stored on the instance so that the state will only
+  // update when this count is equal to or greater than 0
+  editCount = 0;
+
+  handleEditing = (edit) => {
```
sadatay: thinking about the naming of this method - how would you feel about `trackEditing`?  `handle` seems to imply an atomic action, like a form submission or a click event.  here we have `editCount` being mutated and persisting at this level, so it's like an ongoing thing.  similarly `onEdit` might be better named `registerEdit` or `signalEdit`...

it's a weird pattern you've got going here, passing state back up the chain.  i can't think of any other way to do it, though.  definitely gonna do some research on it - i'm sure other people have run into this type of thing before.

sadatay:
did some reading on this last night.  

how would you feel if we yanked all of the state from the field components up into the CustomerResourceShow component and then passed it back down into the field components as a prop?  i started messing around with this last night and it feels a bit better: the individual fields don't need to track their editing state internally, it feels like that's more the responsibility of the page containing the form(s).  all we use `isEditing` for in the fields is display logic, and that feels more `prop` territory.  

If we kept a structure on the `*Show` components for all editable fields and their editing state we wouldn't have to do any of that counter stuff. we could pass a handler _bound to the page component_ into the field components to report the data back up the chain and then reflect it back down through props.

here's an [SO thread](https://stackoverflow.com/questions/36086806/react-accessing-updated-values-from-nested-components) that explains it succinctly, but i found this pattern referenced elsewhere in my reading so it seems to be a fairly accepted approach

src/components/customer-resource-show.js
wwilsman: WRT the weird pattern: your comment this morning is spot on. The only reason I didn't do this initially was because I didn't want to have a very large footprint for this PR. `window.prompt` is not so great, so I knew the logic around this functionality would change in the near future.

That near future is today though. I've been made aware that `stripes-form` supports blocking navigation, but they do not seem to use the method `react-router` suggests. I will be looking into their implementation and try to reproduce it in this PR (if it isn't too hacky).

src/components/customer-resource-show.js
wwilsman: Thanks, this is left over from a refactor away from 

```
this.setState(({ isEditing }) => ({
  isEditing: isEditing + (edit ? 1 : -1)
}))
```

I like your short-circuiting `(edit || -1)` to maybe add `1`, but I think using `?` makes it more clear that `edit` is a boolean and not a number. If, for whatever reason, `edit = 2` then that would throw the count off.

That won't be necessary though, because I like your suggestion of turning the uncontrolled fields into controlled ones. 👍

src/components/customer-resource-show.js
wwilsman: I was just being concise. It was `!!this.editCount` but I updated it after writing the comment above.

Nice use of a bitwise operator! I'll have to add that to my toolpouch. Another good one is:

```
if (~array.indexOf(value)) { ... }
```

`~` is falsy for `-1` but truthy for all other values

tests/customer-resource-embargo-test.js
wwilsman: No more `sinon` stuff. 🙂

I decided to leave in basic tests for each field that ensure editing that specific field and trying to navigate away doesn't go anywhere. But the tests around interacting with the navigation modal have been pulled into the `customer-resource-show` tests. I think we should repeat those tests for packages as well since that might have a different message, or maybe the message will change depending on what's being edited, etc.

wwilsman:
Updated to use a modal similar to what's available in `stripes-form`. Still going to make those editable fields controllable instead of sending state back up

wwilsman:
The `CustomerResourceCoverage` and `CustomEmbargo` components have been updated to support having their editable state controlled from the parent component. If the parent component passes an `onEdit` prop, then that component will be responsible for providing an `isEditable` property as well (which defaults to false).

If these components are used elsewhere without the `onEdit` prop, `isEditable` is used as the initial value of the `isEditing` state.

PR #243, opened 2018-03-01
Add a pane title to detail views

keathleydavidj:
## Purpose
Reference: https://issues.folio.org/browse/UIEH-187
In order to match the rest of FOLIO, this adds titles to preview pane headers.

## Approach
We moved the back button and the close pane logic into a single place.

## Learning
We learned that the `PaneHeader` paneTitle accepts a node as well as a string.

## Screenshots
screen shot 2018-02-28 at 6 02 18 pm


cherewaty:
:shipit:

PR #244, opened 2018-03-01
Implement Provider Sort Options

carolegodfrey:
## Purpose
Adds functionality to sort providers by either provider name or relevance as described https://issues.folio.org/browse/UIEH-84

## Approach
- Followed same logic as used for filters on other search types to add sort filters
- Updates to Search Form to support a url of `sort=name`. Internal to component passes a filter object with sort onto child filters component.
- Updates to Search form to automatically search when the sort filter is selected. Change is made specific to provider sort filter. Another ticket - https://issues.folio.org/browse/UIEH-157 addresses behavior of filters for other search types
- Updated Mirage Config to support A-Z sorting and a simulated Relevance Sort
- A-Z uses string localeCompare so the sort can be done as case insensitive and can sort numbers naturally ( 1 < 2 < 10)
- Relevance sort is simulated based on behavior of RM API - results containing the largest number of words from a search query appear higher in sort order, otherwise fallback is to an A-Z sort
- Required updates to search logic for multiple word queries - considers a match if results contain any word in the query. This is also how A-Z searches function in RM API.
- There are other user stories https://issues.folio.org/browse/UIEH-119, https://issues.folio.org/browse/UIEH-121 address sorting for packages and titles

## Learning
https://medium.freecodecamp.org/functional-setstate-is-the-future-of-react-374f30401b6b
https://github.com/uberVU/react-guide/blob/master/props-vs-state.md

## Screenshots
provider-sort-options

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -1,56 +1,6 @@
 import { okapi } from 'stripes-config'; // eslint-disable-line import/no-unresolved
 import { Response } from '@bigtest/mirage';
-
-/**
- * Helper for creating a search route for a resource type.
- * Currently includes pagination and searching by name.
- * @param {String} resourceType - resource type's model name
- * @param {Function} [filter] - optional filter function
- * @returns {Function} route handler for a search route of this
- * resource type
- */
-function searchRouteFor(resourceType, filter = (model, req) => {
-  let query = req.queryParams.q.toLowerCase();
-  return model.name && model.name.toLowerCase().includes(query);
-}) {
-  return function (schema, req) { // eslint-disable-line func-names
-    let page = Math.max(parseInt(req.queryParams.page || 1, 10), 1);
-    let count = parseInt(req.queryParams.count || 25, 10);
-    let offset = (page - 1) * count;
-
-    let collection = schema[resourceType];
-    let json = this.serialize(collection.all().filter((model) => {
-      return filter(model, req);
-    }));
-
-    json.meta = { totalResults: json.data.length };
-    json.data = json.data.slice(offset, offset + count);
-    return json;
-  };
-}
-
-/**
- * Helper for creating a nested resource route. Currently only accepts
- * page and count params.
- * @param {String} foreignKey - parent resource foreign key prefix
- * @param {String} resourceType - nested resource's model name
- * @returns {Function} route handler for a nested resource
- */
-function nestedResourceRouteFor(foreignKey, resourceType) {
-  return function (schema, req) { // eslint-disable-line func-names
-    let page = Math.max(parseInt(req.queryParams.page || 1, 10), 1);
-    let count = parseInt(req.queryParams.count || 25, 10);
-    let offset = (page - 1) * count;
-
-    let json = this.serialize(schema[resourceType].where({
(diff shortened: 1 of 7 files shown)


tests/provider-search-test.js
```diff
@@ -156,6 +156,159 @@ describeApplication('ProviderSearch', () => {
...
+              it('displays the last results', () => {
+                expect(ProviderSearchPage.$searchResultsItems).to.have.lengthOf(3);
+              });
+
+              it('reflects the sort=name in the URL query params', function () {
+                expect(this.app.history.location.search).to.include('sort=name');
+              });
+            });
+          });
+          describe('then visiting the page with an existing sort', () => {
```
wwilsman: I think this should be pulled out of this describe bock, and put directly within the `"sorting providers"` describe block. That way there isn't already a sort happening in one of the other `beforeEach` hooks.

mirage/config.js
```diff
@@ -179,7 +262,7 @@ export default function configure() {
     let params = req.queryParams;
     let type = params['filter[type]'];
     let selected = params['filter[selected]'];
-    let filtered = pkg.name && pkg.name.toLowerCase().includes(query);
+    let filtered = pkg.name && query.split(' ').some(w => pkg.name.toLowerCase().includes(w));
```
wwilsman: Maybe not important for this PR, but I think `titles` should also have this behavior. It just might get a little hairy with how we can search by different fields for titles. 

Could probably break this out into yet another helper that looks something like `includesWords(string, words)` where `string` could be `pkg.name`, `title.publisherName`, etc., and `words` could be the actual search string

carolegodfrey: Helper function makes sense - I'd like to defer that until the changes are made for other search types. Title being most tricky with multiple search fields. I opted to hold off on the changes for title search since there are 2 more tickets (UIEH-119 and UIEH-121) to address package and title sort. Changing title search (with multiple words)  will most likely impact associated unit tests and I wanted to close out the provider functionality in this PR first

tests/provider-search-test.js
carolegodfrey: Thanks for the recommendation! Just made the change.

mirage/config.js
```diff
@@ -1,6 +1,80 @@
 import { okapi } from 'stripes-config'; // eslint-disable-line import/no-unresolved
 import { Response } from '@bigtest/mirage';
 
+/**
```
sadatay: maybe we should yank these out into a `mirage/helpers.js` file?

carolegodfrey: @sadatay - do you mean the nameCompare, relevanceCompare and getQuery functions? Also eventually an includesWords?

wwilsman: Including `searchRouteFor` and `nestedResourceRouteFor` these helpers are getting up there in numbers. We should move those over to a helpers file too with the rest. So besides imports, the first thing in this file would be `export default function configure()`

carolegodfrey: Added helpers file, moved over noted functions and added new includesWords function as suggested. Thanks for suggestions!

PR #245, opened 2018-03-01
Fix coverage reporting

cherewaty:
## Purpose
Code coverage reporting is broken. It reports 100% for everything and doesn't generate any files.

## Approach
I broke it by improperly adding `babel-plugin-transform-decorators-legacy` in https://github.com/folio-org/ui-eholdings/pull/231. I overwrote the `babel-loader`'s plugins instead of just adding to them.

I rearranged the `plugins` additions to work properly.

### DIFF
--- a/.stripesclirc.js
+++ b/.stripesclirc.js
@@ -76,18 +76,22 @@ const testPlugin = {
         return rule.loader === 'babel-loader';
       });
 
-      if (options.coverage || options.karma.coverage) {
-        // Brittle way of injecting babel-plugin-istanbul into the webpack config.
-        // Should probably be moved to stripes-core when it has more test infrastructure.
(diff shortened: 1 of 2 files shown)


.stripesclirc.js
```diff
@@ -76,19 +76,19 @@ const testPlugin = {
         return rule.loader === 'babel-loader';
       });
 
+      // Make decorators possible
+      config.module.rules[babelLoaderConfigIndex].options.plugins = [
```
wwilsman: Won't this one also override the babel loader plugins?

cherewaty: You're right. `stripes-core` didn't have any `babel-loader` plugins, but I just bulletproofed it in case it does someday.

PR #247, opened 2018-03-01
Fix new isNaN warning

cherewaty:
## Purpose
https://github.com/folio-org/ui-eholdings/pull/226 introduced a new warning when running tests:
```ERROR: 'Warning: Received NaN for the `value` attribute. If this is expected, cast the value to a string.```

## Approach
I needed to take the approach Jason mentioned:
https://github.com/folio-org/ui-eholdings/pull/226#issuecomment-367519623

### DIFF
--- a/src/components/custom-embargo/custom-embargo.js
+++ b/src/components/custom-embargo/custom-embargo.js
@@ -102,7 +102,7 @@ class CustomEmbargoForm extends Component {
                   <Field
                     name='customEmbargoValue'
                     type="number"
-                    parse={value => (!value ? null : (Number.isNaN(value) ? '' : Number(value)))}
+                    parse={value => (!value ? null : (Number.isNaN(Number(value)) ? '' : Number(value)))}
                     component={this.renderTextField}
                   />
                 </div>

### CODE REVIEW

File: src/components/custom-embargo/custom-embargo.js [hunk 1]
```diff
@@ -102,7 +102,7 @@ class CustomEmbargoForm extends Component {
                   <Field
                     name='customEmbargoValue'
                     type="number"
-                    parse={value => (!value ? null : (Number.isNaN(value) ? '' : Number(value)))}
+                    parse={value => (!value ? null : (Number.isNaN(Number(value)) ? '' : Number(value)))}
```
sadatay: i'll 👍 this because it fixes the bug, but the nested ternaries give me pause.  i'm also confused as to why a lack of `value` needs to explicitly result in `null`, and _further_ confused by non-numbers resulting in `''`.  does the API accept both?  if so we should condense this into a single ternary.  seems strange that it would accept an empty string as an embargo value, though...  🤔

cherewaty: I agree that's strange, but I didn't want to invest the time into investigating. We'll need to touch this code again when we tackle full-page vs. inline editing, so we can revisit then.

sduvvuri-ebsco: when I got this warning, I tried to fix it in the same way and it broke the functionality, please refer: https://stackoverflow.com/questions/25176459/is-number-isnan-more-broken-than-isnan

the intention of having that condition there: khalilah wanted something like "if the user enters any other character but a number, then they should not be able to", so if the value entered is not a number, we just enter an empty string, else we parse it into a number. we can make the default '0' instead of an empty string but the user would always have to backspace it to be able to enter a real value which I felt was kinda annoying, so an empty string if it is not a number.

sduvvuri-ebsco: we also have appropriate validations if the user enters an empty string and tries to submit the form, so that should be taken care of in the client-side validations.

PR #248, opened 2018-03-06
Update Readme to include info to run the app on Windows

sduvvuri-ebsco:
## Purpose
Update README to include information to run ui-eholdings on Windows environment

## Approach
- By setting an environment variable and editing package.json file

## Learning
https://yarnpkg.com/en/docs/usage
https://stackoverflow.com/questions/11928013/node-env-is-not-recognized-as-an-internal-or-external-command-operable-comman
https://stackoverflow.com/questions/40034690/node-env-is-not-recognised-as-an-internal-or-external-command/40036979

## Screenshots
_Let's see those sweet visuals!_

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,14 @@ The `ui-eholdings` module adds the ability to manage electronic holdings in FOLI
 * `yarn start`
 * Visit your app at [http://localhost:8080](http://localhost:8080).
 
+On Windows: 
+If you encounter problems trying to run your app on Windows, specifically, with NODE_ENV:
+
+* Set environment variable NODE_ENV=development
+  `SET NODE_ENV=development` if running from Windows command prompt
+  `export NODE_ENV=development` if running from Cygwin on Windows
+* Edit ui-eholdings/package.json to remove `NODE_ENV=development` from scripts object
+
 By default, this will use the backend OKAPI cluster at
 https://okapi.frontside.io However, if you want to run the application
 against the mirage server contained within the browser, you can turn


README.md
```diff
@@ -27,6 +27,14 @@ The `ui-eholdings` module adds the ability to manage electronic holdings in FOLI
 * `yarn start`
 * Visit your app at [http://localhost:8080](http://localhost:8080).
 
+On Windows: 
+If you encounter problems trying to run your app on Windows, specifically, with NODE_ENV:
+
+* Set environment variable NODE_ENV=development
+  `SET NODE_ENV=development` if running from Windows command prompt
+  `export NODE_ENV=development` if running from Cygwin on Windows
+* Edit ui-eholdings/package.json to remove `NODE_ENV=development` from scripts object
```
cherewaty: If it's causing trouble on Windows, should we remove `NODE_ENV=development` from the `start` script completely? My copy on macOS seems to work fine without it.

ttamj: The CLI defaults NODE_ENV to "development" for the `serve` command.
https://github.com/folio-org/stripes-cli/blob/0b939ac8951a3146b940b05b7d591edc39eefc30/lib/commands/serve.js#L11-L15

sduvvuri-ebsco: Do you have the environment variable of NODE_ENV set and just got rid of it from package.json?
I see that NODE_ENV is being used in .stripesclirc.js

cherewaty: I don't seem to be setting `NODE_ENV` anywhere. Maybe it's "development" by default? I opened a PR here for us to test it out: https://github.com/folio-org/ui-eholdings/pull/249

sduvvuri-ebsco: tested your PR on both Windows and Linux and it works just fine. However, point to note, is that after we get that change, folks have to force install their dependencies again, either yarn or yarn install. Else its giving problems finding stripes in Windows. 
Another point is would we have any reason to set NODE_ENV to something other than "development"? If that is the case, should we mention in README that folks can set that env variable?

cherewaty:
@sduvvuri-ebsco should we just close this PR now? Did https://github.com/folio-org/ui-eholdings/pull/249 make the README additions unnecessary?

sduvvuri-ebsco:
yes, we can close this out..

PR #251, opened 2018-03-06
Implement Package Sort Options

carolegodfrey:
## Purpose
Adds sort options as filter to package search (as documented at https://issues.folio.org/browse/UIEH-119)

## Approach
- Added Sort Options as filter (similar to what was done for Providers)
- Updated filter behavior for all search types so that search is conducted upon selection of filter (not requiring a user to click the search button) - This functionality is required for https://issues.folio.org/browse/UIEH-157
- Updated filter tests so that search is not clicked on filter selection
- Updated Mirage search configuration so that title searches are handled in the same way as package and provider searches (results containing any word in a query are returned)

## Screenshots
package-sort-options

### DIFF
--- a/tests/package-search-test.js
+++ b/tests/package-search-test.js
@@ -149,8 +149,6 @@ describeApplication('PackageSearch', () => {
           expect(PackageSearchPage.$searchResultsItems).to.have.lengthOf(3);
         }).then(() => (
           PackageSearchPage.clickFilter('type', 'ebook')
-        )).then(() => (
-          PackageSearchPage.search('Package')
         ));
       });
 
@@ -168,8 +166,6 @@ describeApplication('PackageSearch', () => {
             expect(PackageSearchPage.$searchResultsItems).to.have.lengthOf(1);
           }).then(() => (
             PackageSearchPage.clearFilter('type')
-          )).then(() => (
-            PackageSearchPage.search('Package')
           ));
         });
 
@@ -205,8 +201,6 @@ describeApplication('PackageSearch', () => {
           expect(PackageSearchPage.$searchResultsItems).to.have.lengthOf(3);
         }).then(() => (
           PackageSearchPage.clickFilter('selected', 'true')
-        )).then(() => (
-          PackageSearchPage.search('Package')
         ));
       });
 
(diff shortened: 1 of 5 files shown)


tests/package-search-test.js
```diff
@@ -306,6 +298,160 @@ describeApplication('PackageSearch', () => {
...
+
+      it('has search filters', () => {
+        expect(PackageSearchPage.$searchFilters).to.exist;
+      });
+
+      it('shows the default sort filter of relevance in the search form', () => {
+        expect(PackageSearchPage.getFilter('sort')).to.equal('relevance');
+      });
+
+      it("displays provider entries related to 'health analytics'", () => {
```
sduvvuri-ebsco: Do we want to modify the test to say package instead of provider and also have the correct package name?

tests/package-search-test.js
```diff
@@ -306,6 +298,160 @@ describeApplication('PackageSearch', () => {
...
+      describe('then filtering by sort options', () => {
+        beforeEach(() => {
+          return convergeOn(() => {
+            expect(PackageSearchPage.packageList.length).to.be.gt(0);
+          }).then(() => (
+            PackageSearchPage.clickFilter('sort', 'name')
+          ));
+        });
+
+        it('displays the providers sorted by provider name', () => {
```
sduvvuri-ebsco: same here..package instead of provider

tests/package-search-test.js
```diff
@@ -306,6 +298,160 @@ describeApplication('PackageSearch', () => {
...
+        describe('then searching for other packages', () => {
+          beforeEach(() => {
+            PackageSearchPage.search('search');
+          });
+
+          it('keeps the sort filter of name in the search form', () => {
+            expect(PackageSearchPage.getFilter('sort')).to.equal('name');
+          });
+
+          it('displays the providers sorted by package name', () => {
```
sduvvuri-ebsco: should this be "displays the packages"?

tests/package-search-test.js
carolegodfrey: Yes, thanks for catching those @sduvvuri-ebsco

PR #252, opened 2018-03-06
Improve DetailsView pane headers

cherewaty:
## Purpose
We want panes to have titles/subtitles in their headers so there's some wayfinding if a user scrolls... but we'd also like to have a large title that can display the full length of the title without getting cut off.

## Approach
We can borrow the scrolling strategy used by many iOS apps - initially show a large title, then shrink it down past a certain scroll threshold.

## Next Steps
It'd be awesome to implement this as an option on `stripes-components` `<Pane>`s. Just dropping a `<Pane>` in the current eHoldings search layout doesn't work, and unfortunately we can't just use a `<PaneHeader>` component anymore (since the scrollable part needs to contain the title/subtitle). Action item: use a `<Paneset>` in the eHoldings search layout.

## Screenshots
bohokwp0xc


cherewaty:
Need to fix tests to find the titles/subtitles in the new spot.

cherewaty:
On hold until we have a full-page editing piece built out.

PR #253, opened 2018-03-06
Refactor package-add-new-titles-test to use BigTest

keathleydavidj:
## Purpose
Change `package-add-new-titles-test` to use BigTest page objects and assertions.

## Learning
Ran into an issue around clicking confirmation modal buttons. The confirmation modal is being rendered directly into the `body` rather than as a child of the `package-show` component. Since BigTest page objects are all scoped to `$root` this meant adding a separate page object for the modal. Once that was added, simply appending the two interactions together works as a nice clean fix.

### DIFF
--- a/tests/package-add-new-titles-test.js
+++ b/tests/package-add-new-titles-test.js
@@ -1,8 +1,8 @@
-import { beforeEach, afterEach, describe, it } from '@bigtest/mocha';
 import { expect } from 'chai';
+import { beforeEach, describe, it } from '@bigtest/mocha';
 
 import { describeApplication } from './helpers';
-import PackageShowPage from './pages/package-show';
+import PackageShowPage from './pages/bigtest/package-show';
 
 describeApplication('PackageShowAllowKbToAddTitles', () => {
   let provider,
@@ -69,15 +69,15 @@ describeApplication('PackageShowAllowKbToAddTitles', () => {
     });
 
     it('sets the state of allow KB to add titles to false', () => {
-      expect(PackageShowPage.allowKbToAddTitles).to.be.false;
+      expect(PackageShowPage.hasAllowKbToAddTitles).to.be.false;
     });
 
     it.always('does not display the allow KB to add titles toggle switch', () => {
-      expect(PackageShowPage.allowKbToAddTitlesToggle).to.not.exist;
+      expect(PackageShowPage.hasAllowKbToAddTitlesToggle).to.be.false;
     });
   });
 
-  describe('visiting the package show page with a package that is not selected and selecting a package', () => {
+  describe('visiting the package show page with a package that is not selected', () => {
     beforeEach(function () {
       pkg = this.server.create('package', {
         provider,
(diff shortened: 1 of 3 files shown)


tests/package-add-new-titles-test.js
```diff
@@ -1,10 +1,10 @@
...
 import { expect } from 'chai';
-import it from './it-will';
+import { beforeEach, describe, it } from '@bigtest/mocha';
+
 import { describeApplication } from './helpers';
-import PackageShowPage from './pages/package-show';
+import PackageShowPage, { PackageShowModal } from './pages/bigtest/package-show';
 
-describeApplication('PackageShowAllowKbToAddTitles', () => {
+describeApplication.only('PackageShowAllowKbToAddTitles', () => {
```
keathleydavidj: oops

tests/pages/bigtest/package-show.js
```diff
@@ -0,0 +1,23 @@
...
+@page class PackageShowPage {
+  allowKbToAddTitles = property('checked', '[data-test-eholdings-package-details-allow-add-new-titles] input');
+  hasAllowKbToAddTitles = isPresent('[data-test-eholdings-package-details-toggle-allow-add-new-titles] input');
+  hasAllowKbToAddTitlesToggle = isPresent('[package-details-toggle-allow-add-new-titles-switch]');
+  isSelected = property('checked', '[data-test-eholdings-package-details-selected] input');
+  toggleAllowKbToAddTitles = clickable('[data-test-eholdings-package-details-allow-add-new-titles] input');
+  toggleIsSelected = clickable('[data-test-eholdings-package-details-selected] input');
+}
+
+@page class PackageShowModalPage {
```
wwilsman: I actually just thought of something.

Instead of exporting a named page object for the modal, we could add a property to `PackageShowPage` that is the modal.

```
@page class PackageShowPage {
  modal = new PackageShowModal('#eholdings-package-confirmation-modal')
}
```

And in the test, we can do

```
PackageShowPage.toggleIsSelected()
  .append(PackageShowPage.modal.confirmSelection())
```

keathleydavidj: That's a good idea, let me give that  a shot.

tests/pages/bigtest/package-show.js
```diff
@@ -0,0 +1,23 @@
+import {
+  clickable,
+  isPresent,
+  page,
+  property,
+} from '@bigtest/interaction';
+
+@page class PackageShowModalPage {
```
wwilsman: I think we can safely call this `PackageShowModal` now that we don't have to have a named export.

tests/pages/bigtest/package-show.js
```diff
@@ -0,0 +1,23 @@
...
+  confirmDeselection = clickable('[data-test-eholdings-package-deselection-confirmation-modal-yes]');
+  cancelDeselection = clickable('[data-test-eholdings-package-deselection-confirmation-modal-no]');
+}
+
+@page class PackageShowPage {
+  allowKbToAddTitles = property('checked', '[data-test-eholdings-package-details-allow-add-new-titles] input');
+  hasAllowKbToAddTitles = isPresent('[data-test-eholdings-package-details-toggle-allow-add-new-titles] input');
+  hasAllowKbToAddTitlesToggle = isPresent('[package-details-toggle-allow-add-new-titles-switch]');
+  isSelected = property('checked', '[data-test-eholdings-package-details-selected] input');
+  modal = new PackageShowModal('#eholdings-package-confirmation-modal');
```
cherewaty: Can we access the modal with a `data-test` attribute instead of relying on an id?

cherewaty: We can revisit later.

tests/package-add-new-titles-test.js
```diff
@@ -69,11 +69,11 @@ describeApplication('PackageShowAllowKbToAddTitles', () => {
     });
 
     it('sets the state of allow KB to add titles to false', () => {
-      expect(PackageShowPage.allowKbToAddTitles).to.be.false;
+      expect(PackageShowPage.hasAllowKbToAddTitles).to.be.false;
     });
 
-    it.still('does not display the allow KB to add titles toggle switch', () => {
-      expect(PackageShowPage.allowKbToAddTitlesToggle).to.not.exist;
+    it.always.skip('does not display the allow KB to add titles toggle switch', () => {
```
wwilsman: After a rebase with the new `@bigtest/mocha` upgrade, you should be able to unskip these

keathleydavidj: Alright, done. Let's see what happens.

keathleydavidj: Well, what happened was a new test failure. I have no idea why it is failing though.

wwilsman: I see... But on the bright side, it looks like it's reliably failing 😅

PR #255, opened 2018-03-06
Update title case to sentence case

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-168, Folio uses sentence case, where as ui-eholdings uses Title case. Update title case where used to sentence case.

## Approach
- Go through each file, figure out where static text is in title case and update it to sentence case.
- Update associated tests as well that check for specific error messages.
- Update certain mirage files as well for consistency.

#### TODOS and Open Questions
- [ ] Need to make changes to mod-kb-ebsco to return certain customized error messages in sentence case. Open question is how pervasive we want to be with this change in mod-kb-ebsco, do we just want to modify the custom messages or also modify data returned by RM API, like package, provider, title names etc. to be in sentence case?

## Screenshots
sentence_case

### DIFF
--- a/src/components/customer-resource-coverage/customer-resource-coverage.js
+++ b/src/components/customer-resource-coverage/customer-resource-coverage.js
@@ -291,7 +291,7 @@ const validateWithinPackageRange = (dateRange, packageCoverage, intl) => {
     let packageEndCoverageDate = packageEndCoverage ? moment(packageEndCoverage) : moment();
     let packageRange = moment.range(packageBeginCoverageDate, packageEndCoverageDate);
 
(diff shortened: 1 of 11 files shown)


cherewaty:
I think we want error messages from `mod-kb-ebsco` to be in sentence case. Package, provider, and title names should remain in whatever form RM API returned them (which would usually be Title Case, but not always, like https://folio.frontside.io/eholdings/titles/2393923).

sduvvuri-ebsco:
Ah, ok..interesting, thought that RM API always stored values in Title Case, thanks for your feedback!

PR #261, opened 2018-03-08
Run stylelint in Travis

cherewaty:
## Purpose
I've been noticing my editor catching some stylelint errors recently on files I hadn't been working on. I realized we put stylelint in the `ui-eholdings` dependencies, but weren't actually running it in CI.

## Approach
Run stylelint whenever we run `yarn lint`.

## Next Steps
Long-term, we probably want to run stylelint separate from ESLInt - should make for easier to parse CI runs. We should also explore exposing stylelint errors at runtime (https://www.npmjs.com/package/stylelint-webpack-plugin?) and getting stylelint into the `stripes-cli` blueprint.

### DIFF
--- a/src/components/package-list-item/package-list-item.css
+++ b/src/components/package-list-item/package-list-item.css
@@ -2,13 +2,15 @@
   height: 100%;
 
   &.is-selected {
-    background-color: rgba(0,0,0,.5);
-    color: #FFFFFF;
+    background-color: rgba(0, 0, 0, 0.5);
+    color: #fff;
+
     &:hover {
-      background-color: rgba(0,0,0,.5);
+      background-color: rgba(0, 0, 0, 0.5);
     }
+
     & h5 {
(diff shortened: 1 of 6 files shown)


src/components/package-list-item/package-list-item.css
```diff
@@ -2,13 +2,15 @@
...
+    color: #fff;
+
     &:hover {
-      background-color: rgba(0,0,0,.5);
+      background-color: rgba(0, 0, 0, 0.5);
     }
+
     & h5 {
-      color: #FFFFFF;
+      color: #fff;
```
Robdel12: hm, won't this provide less accuracy than the full hex value?

cherewaty: In this case, no. I think the stylelint rule only wants the three-character version for colors that have an accurate three-character translation. That makes for a few less characters of css.

PR #263, opened 2018-03-08
Convert `package-custom-coverage-test` to use BigTest

Robdel12:
## Purpose
This PR converts `package-custom-coverage-test` to use BigTest. This one went pretty smoothly.


tests/package-custom-coverage-test.js
```diff
@@ -140,7 +140,7 @@ describeApplication('PackageCustomCoverage', () => {
 
       describe('entering valid coverage', () => {
         beforeEach(() => {
-          return convergeOn(() => {
+          return new Convergence(() => {
```
wwilsman: The API for `Convergence` is `new Convergence(timeout)`. However `@bigtest/mocha` will automatically add this timeout. Instead, these should be:

```
return new Convergence().once(() => {
  // ... expectations
});

File: tests/package-custom-coverage-test.js [hunk 2]
```diff
@@ -234,9 +234,9 @@ describeApplication('PackageCustomCoverage', () => {
           });
 
           it('rejects invalid begin date', () => {
-            convergeOn(() => {
+            return new Convergence(() => {
               expect(PackageShowPage.validationError).to.exist;
-            }).then(() => {
+            }).do(() => {
```
wwilsman: `.do()` is used to run an arbitrary function _after_ a convergence. This should be `.once()` so we can converge on it if it fails. We can also group expectations in a single `.once()`, so we shouldn't have to do `.once().once()` unless you're testing for a specific order of events to happen.

wwilsman: Also, `expect(PackageShowPage.validationError).to.exist;` will always pass because `validationError` will always be `true` or `false`. So probably not necessary.

Robdel12: Yeah, while debugging this test it doesn't seem like anything ever runs.. Nothing logs out. I'll come back to it on Monday

tests/package-custom-coverage-test.js
```diff
@@ -234,9 +234,8 @@ describeApplication('PackageCustomCoverage', () => {
           });
 
           it('rejects invalid begin date', () => {
-            convergeOn(() => {
+            return new Convergence().once(() => {
```
wwilsman: _How did I not notice this before?_

The previous `it` that was here is a false positive. It never returns the promise it creates so it's essentially the same thing as `it('passes without a body', () => {})`.

We should just remove the convergence here because `it` is already convergent.

wwilsman:
To fully move these tests over to bigtest, we should use a bigtest page object and adjust our tests accordingly.

Robdel12:
@wwilsman I'll do that tomorrow as a small break from the toast / error work I'm doing

Robdel12:
@wwilsman I need your rereview when you get a chance

cherewaty:
Rebased to resolve conflicts with https://github.com/folio-org/ui-eholdings/pull/281.

Also added commit to remove the now-unused `advancedFillIn()` and `pressEnter()` helpers.

tests/package-custom-coverage-test.js
```diff
@@ -117,55 +112,49 @@ describeApplication('PackageCustomCoverage', () => {
...
-            expect(PackageShowPage.$endDateField).to.exist;
-          });
-        });
-
         describe('with begin date and end date', () => {
           beforeEach(() => {
-            PackageShowPage.inputBeginDate('12/16/2018');
-            PackageShowPage.inputEndDate('12/24/2018');
+            return PackageShowPage.dateFields.fillBeginDate('12/16/2018')
+              .append(PackageShowPage.dateFields.fillEndDate('12/24/2018'));
```
wwilsman: Shouldn't need `.append` here. Chaining from `dateFields` to other `dateFields` methods should work.

```
PackageShowPage.dateFields
  .fillBeginDate('12/16/2018')
  .fillEndDate('12/24/2018');
```

tests/pages/bigtest/package-show.js
```diff
@@ -32,6 +93,26 @@ import {
...
+  customCoverage = text('[data-test-eholdings-package-details-custom-coverage-display]')
+  hasCustomCoverageAddButton = isPresent('[data-test-eholdings-package-details-custom-coverage-button] button')
+  clickCustomCoverageAddButton = clickable('[data-test-eholdings-package-details-custom-coverage-button] button')
+  clickCustomCoverageCancelButton = clickable('[data-test-eholdings-package-details-cancel-custom-coverage-button] button')
+  clickCustomCoverageEditButton = clickable('[data-test-eholdings-package-details-edit-custom-coverage-button] button')
+  clickCustomCoverageSaveButton = clickable('[data-test-eholdings-package-details-save-custom-coverage-button] button')
+  isCustomCoverageDisabled = property('disabled', '[data-test-eholdings-package-details-save-custom-coverage-button] button')
+  validationError = text('[data-test-eholdings-custom-coverage-date-range-begin] [class^="feedbackError"]')
+
+  dateFields = new PackageDatePicker()
```
wwilsman: Should this be scoped to something? 

i.e. `new PackageDatePicker('[data-test-eholdings-package-details-custom-coverage]')`

cherewaty: Yes it should. I also made it plural to make it clear it's for both date pickers. It could use another refactoring pass in the future to make a page object for a generic datepicker. That would DRY this up a lot.

PR #264, opened 2018-03-08
Only run linting on Travis, not Jenkins

cherewaty:
❌  Jenkins only leaves comments in the PR when the lint fails, but does not fail the build.

✅  Travis fails the build with linting errors. 

No need to run lint on Jenkins.


Robdel12:
image

PR #267, opened 2018-03-08
Get rid of early BigTest prototype

cherewaty:
## Purpose
We can now go all-in with BigTest `it()`s.

## Approach
Instead of the early BigTest prototype that was in this repo, we can import directly from the actual BigTest package. The former contents of `tests/it-will.js` are gone.

Two tests were falsely reporting before. It's pretty cool that those false positives were caught in this switch.

## Next Steps
- Refactor all page objects to use BigTest.
- Remove uses of `convergeOn()`, since we shouldn't have to use it directly.

## Screenshots
screen shot 2018-03-08 at 5 23 22 pm


tests/customer-resource-custom-coverage-test.js
```diff
@@ -1,6 +1,6 @@
-/* global describe, beforeEach */
+import { beforeEach, describe, it } from '@bigtest/mocha';
 import { expect } from 'chai';
-import it, { convergeOn } from './it-will';
+import { convergeOn } from '@bigtest/convergence';
```
wwilsman: One thing to note, `convergeOn` is [_technically_ internal](https://github.com/thefrontside/bigtest/tree/master/packages/convergence#convergeon) even though we export it.

If want to enforce this, we should probably convert these to `new Convergence().once()`. We frequently do `convergeOn(...).then()` which will need to be changed to `.do(...)`. If the `.then()` returns a promise, `.do()` supports that as well (though not officially).

wwilsman: Didn't quite read the whole PR before jumping the gun. 😆I was already reviewing before it got opened.

cherewaty: I'll make another PR to change over to that pattern.

cherewaty: https://github.com/folio-org/ui-eholdings/pull/268

PR #268, opened 2018-03-09
[WIP] Replace uses of convergeOn()

cherewaty:
## Purpose
`convergeOn()` is an [intimate API](https://github.com/thefrontside/bigtest/tree/master/packages/convergence#advanced) of bigtest/convergence. We should instead be creating `Convergence` objects.

## Next Steps
- Using BigTest page objects makes most of this refactor irrelevant. Very few of these direct uses of `Convergence` will need to remain.
- I skipped a couple of failing assertions re: custom embargo editing. I'll revisit those when refactoring that test to use BigTest page objects.

## Learning
More on bigtest/convergence: https://github.com/thefrontside/bigtest/tree/master/packages/convergence

PR #269, opened 2018-03-09
Use BigTest for custom embargo test

cherewaty:
## Purpose
BigTest page objects are nice. Let's use them.

## Approach
Now the custom embargo test doesn't need to use bigtest/convergence directly.

Removes the need for some chicanery in https://github.com/folio-org/ui-eholdings/pull/268

## Learning
Copied the modal-in-page-object pattern from https://github.com/folio-org/ui-eholdings/pull/253

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 dist/*
 .vscode
+artifacts
 coverage
 .DS_STORE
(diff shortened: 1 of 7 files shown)


cherewaty:
Rebased to include changes from https://github.com/folio-org/ui-eholdings/pull/241

PR #271, opened 2018-03-13
UIEH-199 Persist Relationship Data in Store

sadatay:
## Purpose
As we navigate through the eholdings app, data from resolved requests is persisted in the redux store.  When we're loading a search route with a selected detail record, one request is fired off for the list of search records and one is fired off for the singleton detail record.  If the search request resolves after the detail request, the detail page's related records are overwritten by the "newer" request, which contains no relationship data.  

The result in the UI is the list of related records on a detail page flickering and disappearing (see [UIEH-199](https://issues.folio.org/browse/UIEH-199)).  This was particularly evident in the Title Details pane, because Title searches take longer than any other RMAPI resource to resolve, and frequently resolved after the detail pane request.  Although the bug was discovered here, theoretically it could affect any view where both a `query` and a `find` request are fired at the same time.

## Approach
This fix introduces a helper method to intelligently merge relationship data.  Once a relationship is established in the store, only updates to that data are accepted.  If a newer request returns with no relationship data whatsoever, the older data is preferred.

Big thanks to @wwilsman for `redux/data` wisdom and direction on where to make the fix.

## Screenshots
**Before:**
before

**After:**
after

### DIFF
--- a/src/redux/data.js
+++ b/src/redux/data.js
@@ -365,6 +365,27 @@ const handlers = {
       }
     }));
 
+    // helper function to merge relationship data without losing
+    // information.  if a record has `data` populated in its
+    // `relationships` key, we want to persist that even if
+    // a more recent request resolves without that relationship data
+    // included.
+    let mergeRelationships = (existing, incoming) => {
+      if (!incoming) { return existing; }
+
+      let mergedRelationships = {};
+
+      for (let recordType in incoming) {
+        if (incoming[recordType] && incoming[recordType].data) {
+          mergedRelationships[recordType] = incoming[recordType];
+        } else {
+          mergedRelationships[recordType] = existing[recordType] || {};
+        }
+      }
+
+      return mergedRelationships;
+    };
+
     // then we reduce each record in the set of records
     next = records.reduce((next, record) => { // eslint-disable-line no-shadow
       return reduceData(record.type, next, (store) => {
@@ -376,7 +397,7 @@ const handlers = {
             [record.id]: {
               ...recordState,
               attributes: record.attributes || {},
-              relationships: record.relationships || {},
+              relationships: mergeRelationships(recordState.relationships, record.relationships),
               isSaving: false,
               isLoading: false,
               isLoaded: true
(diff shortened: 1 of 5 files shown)


cherewaty:
Solution looks good. Is there a reasonable way to write a test for this?

wwilsman:
So I looked into how we could test this, and it looks like we can specify route specific timing like this:

```
server.get('/some/route', () => {
  // ... route handler
}, { timing: 3000 })
```

But, we would have to redefine the handler. I can't quickly find a way to get a reference to the previous handler, otherwise we could just redefine the route with the same handler but different timing.

sadatay:
could we test this in the unit tests for redux/model?

elrickvm:
This came out nice. The solution is elegant.

src/redux/data.js
```diff
@@ -365,6 +365,27 @@ const handlers = {
...
 
+    // helper function to merge relationship data without losing
+    // information.  if a record has `data` populated in its
+    // `relationships` key, we want to persist that even if
+    // a more recent request resolves without that relationship data
+    // included.
+    let mergeRelationships = (existing, incoming) => {
+      if (!incoming) { return existing; }
+
+      let mergedRelationships = {};
```
cowboyd: this is a great opportunity for a funcadelic fold:

```js
return foldl((merged, { key, value }) => {
  if (value && value.data) {
    return append(merged, {[key]: value});
  } else {
    return append(merged, {[key]: existing[key]);
  }  
{}, incoming);
```

cowboyd:
Also, can we not lazily compute this based of the record state Instead of doing it eagerly in the reducer?

cowboyd:
That, or keep both versions of the record in order to compute it lazily. In other words, the store holds facts about the record, not its authorative value.

```js
{
 'packages': {
   a: [{/*...*/}, {/*...*/}, {/*...*/}]
 }
}
```

resolver.packages.a //=> computed off of all known versions of `packages.a`

wwilsman:
_**That** would be pretty sweet._ 

But probably not within the scope of this PR. It would be really awesome to work that in somewhere though. That would also solve some other attributes getting overwritten when missing from the second request (`contributorsList` is one of them that isn't a relationship for example).

cowboyd:
Agree, though you're right.... don't want to derail this PR. As for the code change, if not an import of funcadelic (which I think would be very handy) and should happen eventually, then a `Object.keys(reduce)` would be better IMHO.

PR #272, opened 2018-03-14
Refine preview pane design with accordions

cherewaty:
## Purpose
In setting the stage for adding a coverage statement (https://issues.folio.org/browse/UIEH-79) and redesigning the selection interaction (https://issues.folio.org/browse/UIEH-152), I ended up organizing the preview panes with accordions (https://issues.folio.org/browse/UIEH-175). 

## Next Steps
The "Holding status" section looks 👎 . I'll be tackling that very soon in https://issues.folio.org/browse/UIEH-152.

These two tasks will be the most challenging work of https://issues.folio.org/browse/UIEH-175:
- I didn't yet try to get padding on the scrollable lists on preview pane - that caused some serious regressions in the scroll JS. We can do that later.
- It'd also be nice to put the scrollable lists on preview panes inside a closable accordion. That can come later too.

## Screenshots
eq4znxlga7

### DIFF
--- a/src/components/custom-embargo/custom-embargo.css
+++ b/src/components/custom-embargo/custom-embargo.css
@@ -1,10 +1,11 @@
 @import '@folio/stripes-components/lib/variables';
 
 .custom-embargo-form {
-  padding: 0.5em 1em 1em;
+  padding: 0 0 1em;
 
   &.is-editing {
     background: #e6f3ff;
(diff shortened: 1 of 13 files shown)


cherewaty:
Replaced with accessible solution https://github.com/folio-org/ui-eholdings/pull/274

PR #273, opened 2018-03-14
Update logic for search/filter panel visibility on smaller screens +style for disabled search button

carolegodfrey:
## Purpose
Close out remaining functionality for Filter Behavior and result list update as described at:
_[UIEH-157](https://issues.folio.org/browse/UIEH-157)_
Prior pull request - https://github.com/folio-org/ui-eholdings/pull/251 addressed filter selection (without requiring click of search button).
 
The following 2 items are addressed by this PR
- When the search box is empty the search button is currently disabled and remains disabled until text is entered in the search box. There was no visual indication that the button is disabled. Added styling to the button so this is more apparent. (this is done instead of displaying a message about an empty search)
- Updated behavior of search/filter panel visibility on small screens. Rules as follows:
1. Hide Panel if the user enters a new search term
2. Show Panel if the user changes filtering or sort options (but the search term remains the same)

## Approach
Moved logic to detect new search (either new search term or new parameters) from search routes component into search-paneset (based on feedback from Wil). Compared locations as well as details of search in componentWillReceiveProps
~~Similar to caching of queries, keep track of whether a search term has changed or the filters only have changed (in search routes). These statuses are updated when performing a search and read when rendering results.~~

#### TODOS and Open Questions
- [ ] ~~Would like to understand if the way I am storing this stateful information is recommended in  SearchRoute. `this.searchChanged` and `this.filtersOnlyChanged` Making determination based on current search compared to prior search. Is there a better approach?~~
- [ ] Confirm understanding of behavior for when panel is visible/hidden
- [ ] Added majority of tests to title search and some basic tests to provider and package - trying to avoid redundancy

## Learning
- https://developmentarc.gitbooks.io/react-indepth/content/life_cycle/update/component_will_receive_props.html
- https://reactjs.org/blog/2016/01/08/A-implies-B-does-not-imply-B-implies-A.html
Used chrome responsive web designer, also using developer tools in chrome (with smaller screen widths) provides similar behavior.

## Screenshots
filter-behavior-search-button

### DIFF
--- a/src/components/search-form/search-form.css
+++ b/src/components/search-form/search-form.css
@@ -73,6 +73,17 @@
   &:hover {
     background-color: color(var(--primary) shade(8%));
   }
+
+  &[disabled] {
+    background-color: #ccc;
+    border-color: #ccc;
+    color: #888;
+
+    &:hover {
+      cursor: initial;
+      opacity: 1;
+    }
+  }
 }
 
 .search-results-list {
--- a/src/routes/search.js
+++ b/src/routes/search.js
@@ -148,6 +148,7 @@ class SearchRoute extends Component {
    */
   updateURLParams(params) {
     let { location, history } = this.props;
+
     // if the new query is different from our location, update the location
     if (qs.stringify(params) !== qs.stringify(this.state.params)) {
       let url = this.buildSearchUrl(location.pathname, params);
@@ -156,6 +157,7 @@ class SearchRoute extends Component {
       if (params.q !== this.state.params.q) {
         url = this.buildSearchUrl('/eholdings', params);
       }
+
       // if only the filters have changed, just replace the current location
       if (params.q === this.state.params.q) {
         history.replace(url);
--- a/src/components/search-paneset/search-paneset.js
+++ b/src/components/search-paneset/search-paneset.js
@@ -5,12 +5,11 @@ import PaneMenu from '@folio/stripes-components/lib/PaneMenu';
 import IconButton from '@folio/stripes-components/lib/IconButton';
 import Button from '@folio/stripes-components/lib/Button';
 import capitalize from 'lodash/capitalize';
-
+import { qs } from '../utilities';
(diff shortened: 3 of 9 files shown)


src/routes/search.js
```diff
@@ -44,6 +44,11 @@ class SearchRoute extends Component {
     // cache queries so we can restore them with the search type buttons
     this.queries = {};
 
+    // cache indicators if a search or a filter has changed
+    // these impact the display of search panel on a small screen
```
wwilsman: Since these only impact the _display_ of the search panel, could we move this logic into there somehow? This way, our routes are only concerned with data, and our components can handle their own display logic.

carolegodfrey: Thanks @wwilsman - -Makes sense -- but just trying to understand where I could get context within the search-paneset to detect changes from a previous search to the current search. (for example, understanding current search has new search term or just new filters). The search route seemed logical for this, but I'll see what I can do in the paneset itself

src/components/search-form/search-form.css
```diff
@@ -73,6 +73,17 @@
   &:hover {
     background-color: color(var(--primary) shade(8%));
   }
+
+  &[disabled] {
+    background-color: #ccc !important;
+    border-color: #ccc !important;
+    color: #888 !important;
```
cherewaty: I don't think  `!important` is needed on these three lines, since `.search-submit[disabled]` is more specific than the `.search-submit` that originally gave it background/border/text colors.

src/routes/search.js
cherewaty: Great refactor 👍 makes a lot more sense to me to have that logic in the `SearchPaneset` component instead of the route.

src/components/search-form/search-form.css
carolegodfrey: Thanks @cherewaty - just made the correction and tried it out!

PR #276, opened 2018-03-16
Remove unused blur() helper

cherewaty:
As part of moving to BigTest page objects, the old prototype `blur()` helper is no longer in use.


tests/pages/helpers.js
```diff
@@ -40,23 +40,6 @@ export function advancedFillIn(el, value) {
   });
 }
 
-export function blur(el) {
```
keathleydavidj: Woo! Deleting code is the best!

PR #277, opened 2018-03-16
UIEH-196: Refactor customer-resource-custom-coverage test to use BigTest

SamKeathley:
## Purpose
Ref [UIEH-196](https://issues.folio.org/browse/UIEH-196). 

Change `customer-resource-custom-coverage` to use `BigTest` page objects.

### ToDo
- [ ] The page object could use some refactoring to dry up some of the begin/end date properties.

### DIFF
--- a/tests/customer-resource-custom-coverage-test.js
+++ b/tests/customer-resource-custom-coverage-test.js
@@ -1,10 +1,8 @@
 import { beforeEach, describe, it } from '@bigtest/mocha';
 import { expect } from 'chai';
-import { convergeOn } from '@bigtest/convergence';
-
 import { describeApplication } from './helpers';
-import ResourcePage from './pages/customer-resource-show';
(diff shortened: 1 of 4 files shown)


tests/customer-resource-custom-coverage-test.js
```diff
@@ -39,7 +37,7 @@ describeApplication('CustomerResourceCustomCoverage', () => {
     });
 
     it('does not display coverage', () => {
-      expect(CustomerResourceCoverage.$root).to.not.exist;
+      expect(CustomerResourceCoverage.exists).to.be.false;
```
cherewaty: We could just leave this the same, and we wouldn't need the `isRootPresent` helper: `expect(CustomerResourceCoverage.$root).to.not.exist;`

@wwilsman thoughts?

SamKeathley: We had to change this because whenever you use `$root` when it doesn't exist it throws an error.

PR #282, opened 2018-03-16
Remove Travis config

cherewaty:
`ui-eholdings` will now build on CircleCI.


sadatay:
we'll probably also want to disable travis for the repo.  other than that, i approve of this.  sounds like the circle failures we're seeing currently are something with yarn and the alpine container we're using.  are we in a holding pattern on that for something to update on their end?

cherewaty:
I think the alpine container problem is now fixed. The last step to getting passing Circle builds is getting the BrowserStack key in the Circle env.

vbar-ebsco:
Merging in attempt to resolve CI block

PR #283, opened 2018-03-19
Run Firefox tests directly on Circle container

cherewaty:
## Purpose
We're bumping against rate limits on BrowserStack, and Firefox is frequently running very slowly: https://www.browserstack.com/question/670

## Approach
We can run Firefox tests in Linux directly on the Circle container instead of trying to run them on macOS through BrowserStack. That should help cut down the BrowserStack usage and avoid its Firefox problems.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,7 +47,7 @@ jobs:
           at: ~/
       - run:
           name: Run tests
-          command: yarn test --karma.singleRun --coverage
+          command: yarn test --karma.singleRun --karma.browsers=Chrome --coverage
       - store_test_results:
           path: ./artifacts/junit
       - store_artifacts:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -74,13 +74,22 @@ jobs:
 
   test-firefox:
     docker:
-      - image: circleci/node:latest
+      - image: circleci/node:latest-browsers
     steps:
       - attach_workspace:
           at: ~/
+      - run:
+          name: Install the latest Firefox
```
wwilsman: If we need to install the latest Firefox, do we need to do the same with Chrome?

cherewaty:
The `circleci/node:latest-browsers` image still has Firefox 47, and that didn't go so well: https://circleci.com/gh/folio-org/ui-eholdings/204

cherewaty:
Chrome is always on latest.

PR #285, opened 2018-03-19
Preserve search and filter options when closing preview pane

sadatay:
## Purpose
_Addresses [UIEH-209](https://issues.folio.org/browse/UIEH-209)_

In the three-pane layout, the preview pane can be hidden via the 'x'
icon in the pane header.  This was previously a hard link to
`/eholdings`, which would blow away the search results as well as the
preview pane.  

## Approach
This work uses `router.history.goBack()` instead of a
hard link.  Also added some tests specific to the 'x' button, as our
previous tests only exercised the '<-' button on a full detail record.

## Screenshots

**Before**
before

**After**
after


wwilsman:
What about the scenario where we navigate directly to a URL that shows a search & details pane?

https://folio.frontside.io/eholdings/packages/19-1125?searchType=packages&q=ebsco

Navigating back would feel weird here.

The `<Link>` component accepts a `location` object as a value of the `to` prop. We could strip the ID of the current resource, but preserve all search params.

Something like:

``` javascript
// `location.search` is the entire query param string "?searchType=&q=&etc"
<Link to={{ pathname: '/eholdings', search: currentLocation.search }}>
  <IconButton icon="closeX"/>
</Link>
```

cherewaty:
A fresh rebase and force push should remove the Travis status check from this PR.

cherewaty:
Rebased and added commit.

cherewaty:
Includes commits from https://github.com/folio-org/ui-eholdings/pull/313. That should be merged first.

PR #286, opened 2018-03-19
UIEH-193: Prevent toggling switch when clicking in label's white space

SamKeathley:
## Purpose
Ref [UIEH-193](https://issues.folio.org/browse/UIEH-193)
This change confines the elements inside the label to the length of the label itself, preventing functionality applied to the label from spanning the length of the container.

## Approach
By placing the `ToggleSwitch` and `<h4>` within the label element and styling them within a confined space, we are now only able to click the switch itself to change it's state. As a note, the only way I was able to make the `<h4>` and `ToggleSwitch` remain stacked was by adding a `<br />` between them. 

## Screenshots
# Before

# After

### DIFF
--- a/src/components/customer-resource-show.js
+++ b/src/components/customer-resource-show.js
@@ -125,6 +125,7 @@ export default class CustomerResourceShow extends Component {
                   htmlFor="customer-resource-show-toggle-switch"
                 >
                   <h4>{resourceSelected ? 'Selected' : 'Not selected'}</h4>
+                  <br />
                   <ToggleSwitch
                     onChange={this.handleSelectionToggle}
                     checked={resourceSelected}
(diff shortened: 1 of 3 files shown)


cherewaty:
Are you sure we need the `<br />`s?

cherewaty:
Does this also take care of the same interaction on `/src/components/customer-resource-show.js`?

SamKeathley:
@cherewaty I personally hate that I had to add the `<br />`s but I could not find a different way to take away the click-ability of both the white space by the toggle AND the white space by the `<h4>`s without them. 

Additionally, I didn't previously add the code to `src/components/customer-resource-show.js` but have just pushed that up with the `<br />`s to make sure it's consistent visually and in functionality.

PR #287, opened 2018-03-19
Minimize BrowserStack usage

cherewaty:
## Purpose
`ui-eholdings` has been rate-limited on BrowserStack lately, causing CI builds to take sometimes 10x longer than they should.

## Approach
- Firefox, now being tested directly on the Linux container, now needs to pass on Circle.
- Travis will now test Firefox directly on its containers instead of BrowserStack.
- Stop running Edge and IE 11 tests. IE can't even begin to run the test suite, and we can revisit Edge with https://issues.folio.org/browse/UIEH-182

## Next Steps
- Only run Travis or Circle, not both.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,9 +146,3 @@ workflows:
       - test-firefox:
           requires:
             - checkout-and-install
-      - test-edge:
-          requires:
-            - checkout-and-install
-      - test-ie:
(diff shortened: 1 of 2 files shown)


cherewaty:
Modified to keep Firefox tests running on BrowserStack, since directly on the container tests were flaking.

PR #288, opened 2018-03-20
[WIP] UIEH-221 Create full-page view for editing custom embargo

cherewaty:
## Purpose
`ui-eholdings` has established a pattern of editing sections of data individually on package and package-title records. Other FOLIO apps, for the near future, will have a pattern where an entire record gets edited at once. Ideally we'd be able to have both patterns in eHoldings.

Setup for https://issues.folio.org/browse/UIEH-79

Resolves https://issues.folio.org/browse/UIEH-221

## Approach
Now the `redux-form` fields and validations for editing a custom embargo can be shared by an inline-style form and and a full-page-edit-style form.

## Next Steps
- Refactor package-title custom coverage dates to work the same way. (https://github.com/folio-org/ui-eholdings/pull/305)
- ~~I broke the dependent field functionality where setting the embargo unit to "None" automatically sets the value to 0 (see skipped test). A more predictable pattern would probably be to have a clear button off to the side, like custom coverage. In the meantime, a user can still input "0" and "None", but the fields can't change each other.~~ - updated to restore this piece, even though it requires awkwardly passing around `redux-form`'s `change`
- All the inline-style forms could use a shared component.


cherewaty:
Changed to WIP, we'll get https://github.com/folio-org/ui-eholdings/pull/297 shipped first.

cherewaty:
Now includes changes from https://github.com/folio-org/ui-eholdings/pull/305 - ship that one first

cherewaty:
Replaced by https://github.com/folio-org/ui-eholdings/pull/317

PR #295, opened 2018-03-21
Use stripes-components Button for search

cherewaty:
There's no longer a compelling reason to use our own button that mimics the upstream styles - we should just use the available `stripes-components` button.

localhost_3000_eholdings_searchtype titles iphone 5_se

### DIFF
--- a/src/components/search-form/search-form.js
+++ b/src/components/search-form/search-form.js
@@ -4,6 +4,7 @@ import Link from 'react-router-dom/Link';
 import capitalize from 'lodash/capitalize';
 import isEqual from 'lodash/isEqual';
 import SearchField from '@folio/stripes-components/lib/structures/SearchField';
+import Button from '@folio/stripes-components/lib/Button';
 import ProviderSearchFilters from '../provider-search-filters';
 import PackageSearchFilters from '../package-search-filters';
 import TitleSearchFilters from '../title-search-filters';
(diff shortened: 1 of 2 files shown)


src/components/search-form/search-form.js
```diff
@@ -4,6 +4,7 @@ import Link from 'react-router-dom/Link';
 import capitalize from 'lodash/capitalize';
 import isEqual from 'lodash/isEqual';
 import SearchField from '@folio/stripes-components/lib/structures/SearchField';
+import Button from '@folio/stripes-components/lib/Button';
```
cowboyd: Why not 

```js
import { Button } from '@folio/stripes-components';
```

cherewaty: We can do that with a Button, but we're not importing anything else from `stripes-components`  directly off the package.

The index isn't kept up to date: https://github.com/folio-org/stripes-components/blob/master/index.js

rasmuswoelk:
Hey @cherewaty 

Since you are switching to a Button-component from stripes-components you might wanna use the SegmentedControl component as well instead of your own custom one? I just added an example and there's a readme for it as well in the link below.

https://github.com/folio-org/stripes-components/tree/master/lib/SegmentedControl

cherewaty:
@rasmuswoelk I actually considered making that part of this PR. I'll revisit next week.

PR #296, opened 2018-03-21
[WIP] UIEH-71 display UI errors

Robdel12:
## Purpose

To be able to see errors from the server in the UI. Currently the errors are swallowed & never displayed to the user.

## Approach

The `Toast` component from stripes core is broken (due to dependency changes) & has CSS styling that doesn't make it very reusable. So along with displaying errors in the UI, this PR also includes a reusable `Toast` component with a reusable page object for testing. 

For rendering the errors my approach is to attack each route and check to see if the models `update` property is marked as `isRejected`. If so, pass those errors along to the `Toast` component to display. In the future I think we should take @wwilsman's suggestion and wire the toasts up to redux so they can respond to actions that fire when an error occurs. 

#### TODOS and Open Questions

- [ ] Get actual design for toasts
- [x] Get actual color for an error, info, & success toast
- [ ] Finish implementing error handling throughout the app (at least unskip the skipped tests re: error handling)
- [x] Create a more robust, complete page object for use in testing
- [ ] Update GIFs

## Screenshots

A single error:

A ton of errors:

From the top:

_(for demo of the toast, but not exactly a valid examples)_
Info:

Success

Accessibility:


cherewaty:
Re: Colors:
- `@import '@folio/stripes-components/lib/variables';` at the top of your css file
- `background-color: var(--error)`
- `background-color: var(--warn)` - I would just use this vocab instead of "info"
- `background-color: var(--success)`

src/components/toaster/style.css
```diff
@@ -0,0 +1,96 @@
...
+.toast {
+  display: flex;
+  justify-content: space-between;
+  width: 300px;
+  margin-left: auto;
+  margin-bottom: 20px;
+  margin-right: auto;
+  color: white;
+  padding: 10px;
+  border-radius: 4px;
```
cherewaty: `border-radius: var(--radius)`

Robdel12:
Closing this PR in favor of a new branch name with the correct JIRA ticket

PR #297, opened 2018-03-22
UIEH-79 Add coverage statement editing

cherewaty:
## Purpose
When a package-title is selected, a user can customize it by setting a coverage statement.

Resolves https://issues.folio.org/browse/UIEH-79

## Approach
This will be the first example of a set of `redux-form` fields being shared in two places, the customer resource edit and show routes. Eventually we'll want embargo and coverage dates to work the same way.

A lot of the page objects and event handling could be done more elegantly, but in the interests of time, we've got a bulky-but-working version.

### DIFF
--- a/src/components/coverage-statement-form/coverage-statement-form.js
+++ b/src/components/coverage-statement-form/coverage-statement-form.js
@@ -1,0 +1,166 @@
+import React, { Component } from 'react';
+import { reduxForm } from 'redux-form';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import isEqual from 'lodash/isEqual';
+
+import Button from '@folio/stripes-components/lib/Button';
+import Icon from '@folio/stripes-components/lib/Icon';
+import IconButton from '@folio/stripes-components/lib/IconButton';
+
+import CoverageStatementFields, { validate as validateCoverageStatement } from '../coverage-statement-fields';
+import styles from './coverage-statement-form.css';
+
+const cx = classNames.bind(styles);
+
+class CoverageStatementForm extends Component {
+  static propTypes = {
+    initialValues: PropTypes.shape({
+      coverageStatement: PropTypes.string
+    }).isRequired,
+    isEditable: PropTypes.bool,
+    onEdit: PropTypes.func,
+    onSubmit: PropTypes.func.isRequired,
+    isPending: PropTypes.bool,
+    handleSubmit: PropTypes.func,
+    initialize: PropTypes.func,
+    pristine: PropTypes.bool
+  };
+
+  state = {
+    isEditing: !!this.props.isEditable
+  };
+
+  componentWillReceiveProps(nextProps) {
+    let wasPending = this.props.isPending && !nextProps.isPending;
+    let needsUpdate = !isEqual(this.props.initialValues, nextProps.initialValues);
+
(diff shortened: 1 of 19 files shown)


src/components/coverage-statement-form/coverage-statement-form.js
```diff
@@ -0,0 +1,166 @@
...
+        <div data-test-eholdings-customer-resource-add-coverage-statement-button>
+          <Button
+            type="button"
+            onClick={this.handleEdit}
+          >
+            Set coverage statement
+          </Button>
+        </div>
+      );
+    }
```
Robdel12: Would it be possible to pull this out into methods that return what they're doing? For example:

(ignore the formatting) 
```js
renderCoverageStatement() {
  return (
     <div className={styles['coverage-statement-display']}>
          <span data-test-eholdings-customer-resource-coverage-statement-display>
            {coverageStatement}
          </span>
          <div data-test-eholdings-customer-resource-edit-coverage-statement-button>
            <IconButton icon="edit" onClick={this.handleEdit} />
          </div>
     </div>
  );
}

renderEditingForm() {
  return (
    <form onSubmit={handleSubmit(onSubmit)}>
          <div
            data-test-eholdings-customer-resource-coverage-statement-form
            className={styles['coverage-statement-form-editing']}
          >
            <CoverageStatementFields />
            <div className={styles['coverage-statement-action-buttons']}>
              <div
                data-test-eholdings-customer-resource-cancel-coverage-statement-button
                className={styles['coverage-statement-action-button']}
              >
                <Button
                  disabled={isPending}
                  type="button"
                  role="button"
                  onClick={this.handleCancel}
                  marginBottom0 // gag
                >
                  Cancel
                </Button>
              </div>
              <div
                data-test-eholdings-customer-resource-save-coverage-statement-button
                className={styles['coverage-statement-action-button']}
              >
                <Button
                  disabled={pristine || isPending}
                  type="submit"
                  role="button"
                  buttonStyle="primary"
                  marginBottom0 // gag
                >
                  {isPending ? 'Saving' : 'Save'}
                </Button>
              </div>
              {isPending && (
              )}
            </div>
          </div>
        </form>
  );
}

renderSetCoverage() {
  return (
      <div data-test-eholdings-customer-resource-add-coverage-statement-button>
          <Button
            type="button"
            onClick={this.handleEdit}
          >
            Set coverage statement
          </Button>
        </div>
  );
}

render() {
    let {
      pristine,
      isPending,
      isEditable = this.state.isEditing,
      // change,
      handleSubmit,
      onSubmit
    } = this.props;
    const {
      coverageStatement
    } = this.props.initialValues;
    let contents;

    if (isEditable) {
      contents = this.renderEditingForm();
    } else if (coverageStatement) {
      contents = renderCoverageStatement();
    } else {
      contents = this.renderSetCoverage();
    }

    return (
      <div
        data-test-eholdings-embargo-form
        className={cx(styles['coverage-statement-form'], {
          'is-editing': isEditable
        })}
      >
        {contents}
      </div>
    );
  }
}
```

That way it cleans up the render method. How do people feel about that pattern? It's usually what I do when I feel like `render` is getting out of hand.

cherewaty: Good idea. Done. I look forward to producing an `InlineForm` component that will make this pattern even easier to use.

PR #301, opened 2018-03-22
UIEH-183 display UI errors from PUTs

Robdel12:
## Purpose

To be able to see errors from the server in the UI. Currently the errors are swallowed & never displayed to the user.

## Approach

The `Toast` component from stripes core is broken (due to dependency changes) & has CSS styling that doesn't make it very reusable. So along with displaying errors in the UI, this PR also includes a reusable `Toast` component with a reusable page object for testing. 

For rendering the errors my approach is to attack each route and check to see if the models `update` property is marked as `isRejected`. If so, pass those errors along to the `Toast` component to display. In the future I think we should take @wwilsman's suggestion and wire the toasts up to redux so they can respond to actions that fire when an error occurs. 

#### TODOS and Open Questions

- [x] Get actual design for toasts
- [x] Get actual color for an error, info, & success toast
- [x] Finish implementing error handling throughout the app (at least unskip the skipped tests re: error handling)
- [x] Create a more robust, complete page object for use in testing
- [x] Update GIFs

## Screenshots

A single error:

A ton of errors:

From the top:

_(for demo of the toast, but not exactly a valid examples)_
warn:

Success

Accessibility:


cherewaty:
Needs a couple things moved around after the file moves in https://github.com/folio-org/ui-eholdings/pull/313

Robdel12:
@wwilsman / @cherewaty giving this a bump for a review

src/components/toaster/toast.js
```diff
@@ -0,0 +1,104 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { CSSTransition } from 'react-transition-group';
+import classNames from 'classnames';
+import Icon from '@folio/stripes-components/lib/Icon';
+
+import style from './style.css';
+
+function captialize(word) {
```
wwilsman: `import capitalize from 'lodash/capitalize';`

src/components/toaster/toast.js
```diff
@@ -0,0 +1,104 @@
...
+
+  static defaultProps = {
+    type: 'warn',
+    timeout: 5000
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
wwilsman: We don't need to redefine the constructor to create a state. Instance initializers are supported in our environment.
```
state = {
  isOpen: this.props.isOpen
}
```

src/components/toaster/toast.js
```diff
@@ -0,0 +1,104 @@
...
+      isOpen: props.isOpen || true
```
wwilsman: If `<Toast isOpen={false} {...}/>`, that will do nothing since this state is always set to `true`.

src/components/toaster/toast.js
```diff
@@ -0,0 +1,104 @@
...
+      this.hideToast();
+    }, this.props.timeout);
+  }
+
+  componentWillUnmount() {
+    window.clearTimeout(this.timer);
+  }
+
+  hideToast = () => {
+    this.setState({ isOpen: false }, () => {
```
wwilsman: Do we need this state at all? If we accept an `isOpen` prop, and always call `onClose`, we could just make this a controlled component. The render method below would just reference `props.isOpen` instead of the state.

src/components/toaster/toast.js
Robdel12: TIL, this works!

src/components/toaster/toast.js
Robdel12: Ohh, didn't know we had lodash in our project.

PR #303, opened 2018-03-22
Refactor provider search test to use BigTest

elrickvm:
## Purpose

By refactoring the tests to take full advantage of the entire BigTest library we can significantly reduce and nearly eliminate the flaky tests cases that were causing failures during Continuous Integration runs.

## Approach

 - Refactor the Provider Search Page Object to uses BigTest interactions and covergences

 - Refactor the Provider Search Tests to use the new Big Test Provider Search Page Objects.

## Learning
https://bigtest.js.org/docs/intro.html


cherewaty:
Oops, I approved before I noticed the failures. I'll take another look when tests are passing.

cherewaty:
:shipit:

PR #305, opened 2018-03-23
UIEH-222 Copy custom coverage editing to full-page view

cherewaty:
## Purpose
Package-title custom coverage dates should be editable both inline and in the full-page edit view.

Resolves https://issues.folio.org/browse/UIEH-222.

## Approach
Abstract out the fields and validations in the `CustomerResourceCoverageFields` component.

## Next Steps
- Copy custom embargo editing the same way (https://issues.folio.org/browse/UIEH-221)
- Create an `InlineForm` component to handle a lot of the shared interactions and styles.

### DIFF
--- a/src/components/customer-resource-coverage-fields/customer-resource-coverage-fields.js
+++ b/src/components/customer-resource-coverage-fields/customer-resource-coverage-fields.js
@@ -1,0 +1,243 @@
+import React, { Component } from 'react';
+import { Field, FieldArray } from 'redux-form';
+import PropTypes from 'prop-types';
+import moment from 'moment';
+
+import Datepicker from '@folio/stripes-components/lib/Datepicker';
+import Button from '@folio/stripes-components/lib/Button';
+import IconButton from '@folio/stripes-components/lib/IconButton';
+
+import styles from './customer-resource-coverage-fields.css';
+import { formatISODateWithoutTime } from '../utilities';
+
+export default class CustomerResourceCoverageFields extends Component {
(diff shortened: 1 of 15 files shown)


src/components/customer-resource-coverage-fields/customer-resource-coverage-fields.js
```diff
@@ -0,0 +1,243 @@
...
+import Datepicker from '@folio/stripes-components/lib/Datepicker';
+import Button from '@folio/stripes-components/lib/Button';
+import IconButton from '@folio/stripes-components/lib/IconButton';
+
+import styles from './customer-resource-coverage-fields.css';
+import { formatISODateWithoutTime } from '../utilities';
+
+export default class CustomerResourceCoverageFields extends Component {
+  static propTypes = {
+    packageCoverage: PropTypes.object, // eslint-disable-line react/no-unused-prop-types
```
sadatay: minor nit, but why not just `eslint-disable` this rule for the block instead of line-by-line?

sadatay:
🚢

src/components/customer-resource-coverage-fields/customer-resource-coverage-fields.js
cherewaty: No particular reason 😎 Actually, after looking at it again, I decided to leave it that way, so if we add any props in the future they're immediately checked by ESLint. I'd probably forget to take future props out of the block ignore.

PR #312, opened 2018-03-23
Upgrade Stripes

cherewaty:
Updates to the latest Stripes CLI/Core/Components.

Biggest changes:
- search field no longer has rounded corners
- profile icon is a circle
- inputs now have a "changed" state (a blue background), so inputs have a state between explicitly validated (green) and explicitly invalidated (red)

localhost_3000_ iphone 6_7_8

### DIFF
--- a/src/components/coverage-statement-fields/coverage-statement-fields.js
+++ b/src/components/coverage-statement-fields/coverage-statement-fields.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { Field } from 'redux-form';
 
-import TextArea from '@folio/stripes-components/lib/TextArea';
+import { TextArea } from '@folio/stripes-components';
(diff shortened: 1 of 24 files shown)


cherewaty:
Update: uses prettier imports enabled by https://github.com/folio-org/stripes-components/pull/298

src/components/coverage-statement-fields/coverage-statement-fields.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { Field } from 'redux-form';
 
-import TextArea from '@folio/stripes-components/lib/TextArea';
+import { TextArea } from '@folio/stripes-components';
```
Robdel12: Sweet breaking API change

PR #315, opened 2018-03-24
UIEH-240 Run Firefox tests on CircleCI

cherewaty:
With all tests now using BigTest page objects, we should see far fewer flaky test situations with Firefox on a Linux container.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,18 +92,12 @@ jobs:
       - attach_workspace:
           at: ~/
       - run:
-          name: Install the latest Firefox ESR
-          command: |
-            sudo apt-get remove firefox-mozilla-build &&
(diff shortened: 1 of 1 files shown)


cherewaty:
This might need some iteration. I'll regularly rebase over the next week or so to see how reliable it will be.

Robdel12:
@cherewaty I rebased this and pushed. Curious to see what 11 days of work does

Robdel12:
@cherewaty these tests passed?!

cherewaty:
@Robdel12 yes they passed, but they took 50% longer than Chrome 🤔

cherewaty:
Circle switched to Firefox 62 in its default images today, so these tests now pass 🎉

PR #318, opened 2018-03-26
UIEH-212 Create settings subnavigation

cherewaty:
## Purpose
Until now, `ui-eholdings` has only had one screen of settings. With root proxy settings about to be built and custom labels coming very soon, we needed a way to have more than one screen of settings.

## Prior art in FOLIO
`stripes-components` has a `Settings` component to do this exact thing. Reasons for not using that:
- It's unusable on narrow screens. The UX pattern is such that if you select a module in the leftmost pane, you'll be taken directly to the first item in that module's subnav, with no way to ever access the subnav. Using the component as-is would mean eHoldings is no longer 100% responsive to screens 320px and wider.
- It introduces a router `Switch` in a component a few levels down in the application hierarchy. In eHoldings, all routes are in `/src/index.js`. Routes shouldn't be spread all over the application - they should be established in a single top-level place. This would be a very large change in the way the `Settings` component works.
- Its logic determining whether a link in the list has an active state makes brittle assumptions about the URL structure. Not a blocker, but needs to be addressed.
- It's coupled to `stripes-connect`. That's probably not blocking eHoldings from using it, but it could have some strange side-effects. eHoldings does not use `stripes-connect` at all - we use `react-redux` directly.

Changes to the `Settings` component to meet a more general use case would definitely mean breaking changes for other consumers of that component. Addressing the issues above would need test coverage and a thoughtful deprecation strategy to have an unchaotic roll-out.

## Approach
This creates a new `Settings` component that contains the navigation, and establishes `SettingsKnowledgeBase` and `SettingsDetailPane` components.

The subnav and detail panes both show back buttons at small screen sizes.

The layout at wider screen sizes is almost identical to the other FOLIO apps' settings, with small tweaks to proportions and a max-width set on the rightmost pane's content.

## Next Steps
The knowledge base credentials form needs some consistency work - either it should start using `redux-form` and `stripes-components` `TextField`s (probably the best option), or it should get its own CSS updated to match.

## Screenshots

localhost_3000_settings_ ipad


ttamj:
@cherewaty This seems like something we would want to address within the original `<settings>` component itself for consistency across the FOLIO community.  Even with a breaking change, we can rely on versioning to keep other apps on their current version until upgraded.

Is part of the trouble with the existing `<settings>` component, the `<paneset>` and `<pane>` within?

skomorokh:
With regards to the routing, `<settings>` is typically at the top level insofar as the settings API is concerned--this is a different entry point than for the app. Indeed, some modules only provide settings. Eventually we hope to make this clear by allowing modules to offer implementations of multiple APIs so that apps with settings panels will explicitly implement both "app" and "settings" (more discussion in JIRA STRIPES-492 which culminated in STCOR-148)

eg. stripes-core matches `/eholdings` but delegates the rest of the path to the component implementing "app" for that module. It matches `/settings/eholdings` and delegates the rest to the component implementing "settings". Right now a) apps are assumed to implement both as there is no way to express otherwise b) these are the same component.

I share Matt's hope that the standard component can be improved through this work so the whole project can benefit.

PR #319, opened 2018-03-27
UIEH-212 Create settings subnavigation (alternate implementation)

cherewaty:
## Purpose
Until now, ui-eholdings has only had one screen of settings. With root proxy settings about to be built and custom labels coming very soon, we needed a way to have more than one screen of settings.

Groundwork for https://issues.folio.org/browse/UIEH-212

## Approach
To compare to https://github.com/folio-org/ui-eholdings/pull/318, this uses the `stripes-components` `Settings` with modifications: https://github.com/folio-org/stripes-components/pull/301

#### TODO
- [x] Remove commit pointing to `stripes-components` fork.

### DIFF
--- a/src/components/settings-detail-pane/settings-detail-pane.js
+++ b/src/components/settings-detail-pane/settings-detail-pane.js
@@ -22,12 +22,11 @@ export default class SettingsDetailPane extends Component {
         defaultWidth="fill"
         paneTitle={paneTitle}
         firstMenu={(
-          <div className={styles['settings-detail-pane-back-button']}>
-            <IconButton
-              icon="left-arrow"
-              href="/settings/eholdings"
-            />
-          </div>
+          <IconButton
+            icon="left-arrow"
+            href="/settings/eholdings"
+            className={styles['settings-detail-pane-back-button']}
+          />
         )}
       >
         {children}
--- a/src/components/settings-detail-pane/settings-detail-pane.js
+++ b/src/components/settings-detail-pane/settings-detail-pane.js
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import Link from 'react-router-dom/Link';
 import Pane from '@folio/stripes-components/lib/Pane';
 import IconButton from '@folio/stripes-components/lib/IconButton';
 
@@ -23,9 +22,12 @@ export default class SettingsDetailPane extends Component {
         defaultWidth="fill"
         paneTitle={paneTitle}
         firstMenu={(
-          <Link to="/settings/eholdings" className={styles['settings-detail-pane-back-button']}>
-            <IconButton icon="left-arrow" />
-          </Link>
+          <div className={styles['settings-detail-pane-back-button']}>
+            <IconButton
+              icon="left-arrow"
+              href="/settings/eholdings"
+            />
+          </div>
         )}
       >
         {children}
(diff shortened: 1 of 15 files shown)


src/index.js
```diff
@@ -53,9 +56,26 @@ export default class EHoldings extends Component {
...
+          },
+          {
+            route: 'root-proxy',
+            label: 'Root proxy',
+            component: SettingsRootProxyRoute
+          }
+        ]}
+        paneTitle="eHoldings"
+        activeLink={router.route.location.pathname}
+      />
```
cherewaty: How do we feel about this pattern?

cherewaty:
Blocked by https://github.com/folio-org/ui-eholdings/pull/312

src/index.js
Robdel12: ehhhhhh, so `Settings` contains a router and handles its own routes? I feel like scattering routes in many places throughout a SPA is a recipe for disaster.

If we can't prevent the bad pattern from entering our app (or contribute upstream a better fix) I think we should document that this is not a pattern that should be replicated in elsewhere in our app. ☣

wwilsman: I'd much rather pass actual `<Route>`s as children to this component.

```
<Route ... component={Settings}>
  <Route name="Knowledge Base" ... component={SettingsKnowledgeBaseRoute}/>
</Route>
```

Settings component can look down at props to get the name.

```
let names = React.Children.map(this.props.children, child => child.props.name)
```

cherewaty: I'm hesitant to make any more changes to `stripes-components`' `Settings` without simultaneously adding test coverage. How about the commit I just added as a compromise? It at least hides away the gook inside the `SettingsRoute`.

src/index.js
```diff
@@ -56,7 +59,10 @@ export default class EHoldings extends Component {
     } = this.props;
 
     return showSettings ? (
-      <Route path={rootPath} component={SettingsRoute} />
+      <Route path={rootPath} component={SettingsRoute}>
+        <Route path="knowledge-base" exact component={SettingsKnowledgeBaseRoute} name="Knowledge base" />
+        <Route path="root-proxy" exact component={SettingsRootProxyRoute} name="Root proxy" />
+      </Route>
```
wwilsman: I like this much better. We should add a comment above the knowledge base route that says something like "these paths are not used by the Route component, but by the settings component later on"

Just because my immediate reaction to this is "these paths look wrong"

wwilsman: Also, since we kinda just throw away these `Route` components after pulling the data off of them in `SettingsRoute`, `exact` is probably unnecessary. Although I could see the argument of keeping it if the upstream `Settings` component ends up supporting nested routes like this.

cherewaty: I left `exact` so it's a little closer to the ideal. Comments added.

src/components/settings-detail-pane/settings-detail-pane.js
```diff
@@ -0,0 +1,35 @@
...
+    } = this.props;
+
+    return (
+      <Pane
+        defaultWidth="fill"
+        paneTitle={paneTitle}
+        firstMenu={(
+          <Link to="/settings/eholdings" className={styles['settings-detail-pane-back-button']}>
+            <IconButton icon="left-arrow" />
+          </Link>
```
wwilsman: `<button>` within an `<a>` here. This is the last thing, looks good otherwise

PR #321, opened 2018-03-27
UIEH-223 Copy coverage date editing to full page package edit route

cherewaty:
## Purpose
https://issues.folio.org/browse/UIEH-223
A package's attributes should be editable from a "full-page" view, not just an "inline" form. 

## Approach
I used the same patterns established in https://github.com/folio-org/ui-eholdings/pull/305 and https://github.com/folio-org/ui-eholdings/pull/317.

I had to avoid using the pane header dropdown because of the lack of z-index management in FOLIO. When there's an active search query, the pane header dropdown isn’t visible, because the z-indices we’re using for eHoldings layout are higher than anything in `stripes-core` or `stripes-components` (but there aren’t enough integers between those used in Stripes and zero for the things we’re doing). Instead I just put a pencil icon in the top right, which is what other FOLIO apps have at the moment anyway. I updated the customer resource edit to work the same way.

## Next Steps
- Create an `InlineForm` component to handle a lot of the shared interactions and styles.

### DIFF
--- a/src/components/customer-resource-show.js
+++ b/src/components/customer-resource-show.js
@@ -3,7 +3,9 @@ import PropTypes from 'prop-types';
 
 import {
   Button,
-  KeyValue
+  IconButton,
+  KeyValue,
+  PaneMenu
 } from '@folio/stripes-components';
 
 import DetailsView from './details-view';
@@ -127,13 +129,6 @@ export default class CustomerResourceShow extends Component {
       }];
     }
 
-    let actionMenuItems = [
-      {
-        label: 'Edit',
-        to: `/eholdings/customer-resources/${model.id}/edit`
-      }
-    ];
-
     return (
       <div>
         <Toaster toasts={errors} position="bottom" />
@@ -143,7 +138,13 @@ export default class CustomerResourceShow extends Component {
           model={model}
           paneTitle={model.name}
           paneSub={model.packageName}
-          actionMenuItems={actionMenuItems}
+          lastMenu={(
+            <PaneMenu>
+              <Link to={`/eholdings/customer-resources/${model.id}/edit`}>
+                <IconButton icon="edit" />
+              </Link>
+            </PaneMenu>
+          )}
           bodyContent={(
             <div>
               <DetailsViewSection label="Holding status">
--- a/src/components/customer-resource-show.js
+++ b/src/components/customer-resource-show.js
@@ -140,9 +140,10 @@ export default class CustomerResourceShow extends Component {
           paneSub={model.packageName}
           lastMenu={(
             <PaneMenu>
-              <Link to={`/eholdings/customer-resources/${model.id}/edit`}>
-                <IconButton icon="edit" />
-              </Link>
+              <IconButton
+                icon="edit"
+                href={`/eholdings/customer-resources/${model.id}/edit`}
+              />
             </PaneMenu>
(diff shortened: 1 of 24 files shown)


src/components/customer-resource-show.js
```diff
@@ -141,7 +138,13 @@ export default class CustomerResourceShow extends Component {
           model={model}
           paneTitle={model.name}
           paneSub={model.packageName}
-          actionMenuItems={actionMenuItems}
+          lastMenu={(
+            <PaneMenu>
+              <Link to={`/eholdings/customer-resources/${model.id}/edit`}>
+                <IconButton icon="edit" />
```
wwilsman: I think this renders an actual `<button>` within the `<a>` (and I think we might be doing this elsewhere?). I just checked and `IconButton` accepts an `href` prop and it will use `Link` behind the scenes. This way we can avoid the accessibility and browser issues with nested interaction elements.

src/components/package-custom-coverage/package-custom-coverage.js
```diff
@@ -133,78 +61,101 @@ class PackageCustomCoverage extends Component {
...
+        <PackageCoverageFields />
+        <div className={styles['custom-coverage-action-buttons']}>
+          <div
+            data-test-eholdings-package-details-cancel-custom-coverage-button
+            className={styles['custom-coverage-action-button']}
+          >
+            <Button
+              disabled={isPending}
+              type="button"
+              role="button"
```
wwilsman: Is `role="button"` necessary here? (it's already a `<button>`)

I know this is what was previously here, but while we're at it we can clean it up a little.

Unless there is a reason this button element needs to explicitly have a button role?

cherewaty: Probably not needed. Removed everywhere we were doing that with a `Button`.

src/components/package-coverage-fields/package-coverage-fields.js
```diff
@@ -0,0 +1,116 @@
...
+            <p data-test-eholdings-coverage-fields-no-rows-left>
+              No date range set. Saving will remove custom coverage.
+            </p>
+            <div
+              className={styles['coverage-fields-add-row-button']}
+              data-test-eholdings-coverage-fields-add-row-button
+            >
+              <Button
+                type="button"
+                role="button"
```
wwilsman: Missed a role

src/components/package-edit/package-edit.js
```diff
@@ -0,0 +1,117 @@
...
+              <PackageCoverageFields />
+            </DetailsViewSection>
+            <div className={styles['package-edit-action-buttons']}>
+              <div
+                data-test-eholdings-package-cancel-button
+              >
+                <Button
+                  disabled={model.update.isPending}
+                  type="button"
+                  role="button"
```
wwilsman: And another

src/components/package-edit/package-edit.js
```diff
@@ -0,0 +1,117 @@
...
+                  Cancel
+                </Button>
+              </div>
+              <div
+                data-test-eholdings-package-save-button
+              >
+                <Button
+                  disabled={pristine || model.update.isPending}
+                  type="submit"
+                  role="button"
```
wwilsman: Last one

PR #322, opened 2018-03-27
 Use KeyValue from stripes-components

cherewaty:
Use the `KeyValue` component provided by `stripes-components`, instead of a custom fork.

Unlocked by https://github.com/folio-org/stripes-components/pull/275


cherewaty:
Needs https://github.com/folio-org/ui-eholdings/pull/312 merged first.

PR #327, opened 2018-04-02
UIEH-71: Display GET errors

SamKeathley:
## Purpose
Re: [UIEH-71](https://issues.folio.org/browse/UIEH-71)
Currently we do not have `GET` errors displaying to the user. This change displays a generic `error` message to users so that they know that the application is listening, but also that something has gone awry.

## Approach
I based my work off of @Robdel12's pull request #301 to add similar functionality to the `GET` errors as the `PUT`s, utilizing the `toast` components written here. To make the code a bit more dry and reusable, I added both the aforementioned `PUT` errors work, as well as my `GET` errors work, into the `utilities` file.

## Screenshots
This is an example of an error that displays with a package that does not exist. Identical to the errors one gets for `PUT`s.
localhost_3000_eholdings_packages_1

### DIFF
--- a/src/components/utilities.js
+++ b/src/components/utilities.js
@@ -65,3 +65,17 @@ export const qs = {
   parse: path => queryString.parse(path, { ignoreQueryPrefix: true }),
   stringify: params => queryString.stringify(params, { encodeValuesOnly: true })
 };
+
+export const processErrors = ({ request, update }) => {
+  let processErrorsSet = ({ errors, timestamp }) => errors.map((error, index) => ({
+    message: error.title,
+    type: 'error',
+    id: `error-${timestamp}-${index}`
+  }));
+
+  let hasErrors = update.isRejected || request.isRejected;
+  let putErrors = hasErrors ? processErrorsSet(update) : [];
+  let getErrors = hasErrors ? processErrorsSet(request) : [];
+
+  return getErrors.concat(putErrors);
+};
--- a/yarn.lock
+++ b/yarn.lock
@@ -327,7 +327,7 @@
     react-router-dom "^4.0.0"
     redux-form "^7.0.3"
 
-"@folio/ui-testing@github:folio-org/ui-testing#framework-only":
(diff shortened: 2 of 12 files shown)


src/components/utilities.js
```diff
@@ -65,3 +65,16 @@ export const qs = {
...
+    message: error.title,
+    type: 'error',
+    id: `error-${timestamp}-${index}`
+  }));
+  const hasPutErrors = update.isRejected;
+  const putErrors = hasPutErrors ? processErrorsSet(update) : [];
+  const hasGetErrors = request.isRejected;
+  const getErrors = hasGetErrors ? processErrorsSet(request) : [];
+  return getErrors.concat(putErrors);
+};
```
Robdel12: Nice! I think this eventually will be replaced with redux dispatching actions when an error occurs but that's a follow up ticket. 

I took a stab at trying to reduce this down but didn't really come up with much 😅 (mostly let > const) 

```js
export const processErrors = ({ request, update }) => {
  let processErrorsSet = ({ errors, timestamp }) => errors.map((error, index) => ({
    message: error.title,
    type: 'error',
    id: `error-${timestamp}-${index}`
  }));

  let hasErrors = update.isRejected || request.isRejected;
  let putErrors = hasErrors ? processErrorsSet(update) : [];
  let getErrors = hasErrors ? processErrorsSet(request) : [];

  return getErrors.concat(putErrors);
};
```

yarn.lock
```diff
@@ -327,7 +327,7 @@
     react-router-dom "^4.0.0"
     redux-form "^7.0.3"
 
-"@folio/ui-testing@github:folio-org/ui-testing#framework-only":
+"@folio/ui-testing@folio-org/ui-testing#framework-only":
```
Robdel12: 🤔 wonder why this is popping up

src/components/utilities.js
SamKeathley: Ah, you're right haha. That does look better. Thanks!

PR #328, opened 2018-04-02
UIEH-223 Create InlineForm component

cherewaty:
## Purpose
Continuation of https://issues.folio.org/browse/UIEH-223. A lot of the interactions were repeated several times. The pattern for editing individual fields needed some DRYing up.

## Approach
Creates an `InlineForm` component that contains the save and cancel buttons for every inline editing interaction.

This DRYing up could probably be taken even farther. It'd be awesome to isolate `redux-form` as much as possible. I'll revisit as time allows this week.


src/components/custom-embargo-form/custom-embargo-form.js
```diff
@@ -53,98 +53,88 @@ class CustomEmbargoForm extends Component {
...
-        <div data-test-eholdings-customer-resource-add-custom-embargo-button>
-          <Button
-            type="button"
-            onClick={this.handleEditCustomEmbargo}
-          >
-            Set custom embargo period
-          </Button>
-        </div>
-      );
+      contents = this.renderSetEmbargo();
```
Robdel12: OHHH 🔥 love this refactor

cherewaty:
Yes, it would make sense to break out an `InlineForm` page object/interactor, but I'm going to go ahead and ship this.

PR #331, opened 2018-04-03
Clean up a11y issues

cherewaty:
Fixes the easiest wins:
- Color contrast of text in results lists
- Add readable test to icon buttons

Bonus: fixes search filters pane sizing behavior.

### DIFF
--- a/src/components/customer-resource-show.js
+++ b/src/components/customer-resource-show.js
@@ -136,7 +136,7 @@ export default class CustomerResourceShow extends Component {
             <PaneMenu>
               <IconButton
                 icon="edit"
-                title={`Edit ${model.name}`}
+                ariaLabel={`Edit ${model.name}`}
                 href={`/eholdings/customer-resources/${model.id}/edit`}
               />
             </PaneMenu>
--- a/src/components/customer-resource-show.js
+++ b/src/components/customer-resource-show.js
@@ -136,6 +136,7 @@ export default class CustomerResourceShow extends Component {
             <PaneMenu>
               <IconButton
                 icon="edit"
+                title={`Edit ${model.name}`}
                 href={`/eholdings/customer-resources/${model.id}/edit`}
               />
             </PaneMenu>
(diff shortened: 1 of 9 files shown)


wwilsman:
RE the chrome failure:

Looks like it can't find this selector `[class^="feedbackError--"]`

There's a possibility that the class wasn't first in it's list. `^=` checks the beginning of the attribute. So it would match `class="feedbackError--gH76k blah"` but not `class="blah feedbackError--gH76k"`

We might want to change that to `[class*="feedbackError--"]` instead. But that could possibly match `class="someother-feedbackError--something"`.

I think the safest bet is to use the generated class names by importing the style like we're doing in the [toast page object](https://github.com/folio-org/ui-eholdings/blob/master/tests/pages/toast.js).

cherewaty:
Flaky test fix attempt: https://github.com/folio-org/ui-eholdings/pull/333

src/components/customer-resource-show.js
```diff
@@ -136,6 +136,7 @@ export default class CustomerResourceShow extends Component {
             <PaneMenu>
               <IconButton
                 icon="edit"
+                title={`Edit ${model.name}`}
```
Robdel12: So I know we have no control over this right now but `title` is not for a11y. Here's a great resource as to why: https://a11yproject.com/posts/title-attributes/

But the TL;DR screen readers treat `title` differently. If you want to label something use `aria-label`

src/components/package-list-item/package-list-item.js
```diff
@@ -32,9 +32,12 @@ export default function PackageListItem({
         'is-selected': active
       })}
     >
-      <h5 data-test-eholdings-package-list-item-name>
+      <div
```
Robdel12: If not an `h5` is this better semantically as a paragraph tag?

cherewaty: I took this commit back out. The headings are more semantically correct, but we need them to be an h3 in one instance and an h4 in another. I'll revisit that later.

src/components/customer-resource-show.js
cherewaty: Lucky for us... `IconButton` has an `ariaLabel` prop. Changed.

Robdel12: Oh good! I'd like to deprecate the `title` prop entirely. It should pretty much never be used.

PR #338, opened 2018-04-05
Refactor KB auth form to use redux-form

cherewaty:
## Purpose
The knowledge base settings form wasn't using `stripes-components` text inputs, or the `redux-form` patterns established in the rest of `ui-eholdings`. For consistency, we should handle this form the same way.

## Approach
Also now uses `Toast`s when an error occurs.

### Before

folio frontside io_ ipad 1

### After

localhost_3000_settings_eholdings_knowledge-base ipad

### DIFF
--- a/tests/backend-configuration-test.js
+++ b/tests/backend-configuration-test.js
@@ -52,7 +52,7 @@ describeApplication('With unconfigured backend', {
       });
     });
 
-    describe('when visiting settings', () => {
+    describe('when visiting the KB auth form', () => {
       beforeEach(function () {
         return this.visit('/settings/eholdings/knowledge-base', () => expect(SettingsPage.$root).to.exist);
       });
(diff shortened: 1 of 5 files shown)


tests/backend-configuration-test.js
```diff
@@ -192,7 +192,7 @@ describeApplication('With valid backend configuration', () => {
         });
 
         it('shows an error message', () => {
-          expect(SettingsPage.errorText).to.equal('an error has occurred');
+          expect(SettingsPage.toast.errorText).to.equal('an error has occurred');
```
Robdel12: awh, does this mean there's no inline form errors?

cherewaty: There are client-side validations to make sure the fields aren't empty, but yes, any server error just shows the toast.

PR #340, opened 2018-04-09
UIEH-259 Edit custom packages

cherewaty:
## Purpose
Users need to be able to edit the attributes of custom packages. Custom packages are currently shown in search results, but they have the exact same UX as managed packages. This starts down the road of diverging managed and custom packages.

Part of https://issues.folio.org/browse/UIEH-259
Part of https://issues.folio.org/browse/UIEH-248
Part of https://issues.folio.org/browse/UIEH-275

## Approach
When hitting `eholdings/packages/:id/edit`, determine from the package's `isCustom` attribute whether to show the managed or custom package edit form.

*Custom package editing currently does not work in production.* This PR does not change that, but sets up everything needed in mirage to keep iterating on the UI.

## Next Steps
- Be able to create a custom package from scratch (`/eholdings/packages/new`).
- Add holding status and visibility toggles


tests/custom-package-edit-test.js
```diff
@@ -0,0 +1,201 @@
...
+        }]
+      }, 500);
+
+      return this.visit(`/eholdings/packages/${providerPackage.id}/edit`, () => {
+        expect(PackageEditPage.$root).to.exist;
+      });
+    });
+
+    it('dies with dignity', () => {
+      expect(PackageEditPage.hasErrors).to.be.true;
```
Robdel12: No toast error message?

cherewaty: Toasts added.

src/components/custom-package-edit/custom-package-edit.js
```diff
@@ -0,0 +1,149 @@
...
+
+    if (queryParams) {
+      actionMenuItems.push({
+        label: 'Full view',
+        to: `/eholdings/packages/${model.id}/edit`
+      });
+    }
+
+    return (
+      <div>
```
Robdel12: It'd be cool to use a fragment here. Try `<> </>` first and if that doesn't work: `React.Fragment`

cherewaty: Oops, merged then saw this. I'll give that a shot on some follow-up work. I definitely like the reduced DOM.

PR #342, opened 2018-04-11
UIEH-283 Add package type label

cherewaty:
## Purpose
We want to be able to tell whether a package is managed or custom.
Resolves https://issues.folio.org/browse/UIEH-283

## Screenshots
localhost_3000_eholdings_ iphone 6_7_8

localhost_3000_eholdings_packages_123355-2720678 iphone 6_7_8


cherewaty:
Requires https://github.com/folio-org/mod-kb-ebsco/pull/95

PR #351, opened 2018-04-13
Reorganize package and resource components

cherewaty:
## Purpose
With our development of custom package and custom title forms, we'll be creating over two dozen new components for individual fields and their inline form variants. The thing is... those `package` and `resource` field components are only used by a very specific set of parents. They don't need to live at the top level of a "common" `components` directory.

## Approach
Nest components related to package and resource editing inside of `_fields` and `_forms` directories for their specific domain.

### Before
screen shot 2018-04-13 at 2 03 22 pm

(Not pictured: `custom-package-edit`, `managed-package-edit`)

### After
screen shot 2018-04-13 at 2 02 57 pm


src/components/resource-edit/resource-edit.js
```diff
@@ -1,143 +0,0 @@
...
-const validate = (values, props) => {
-  return Object.assign({}, validateCoverageDates(values, props), validateCoverageStatement(values), validateEmbargo(values));
-};
-
-export default reduxForm({
-  validate,
-  enableReinitialize: true,
-  form: 'ResourceEdit',
-  destroyOnUnmount: false,
-})(ResourceEdit);
```
wwilsman: Where did this file and the other `resource-edit` files go? I noticed there are `custom-resource-edit` and `managed-resource-edit`, but those were pre-existing.

If I had to assume, this was left behind from another change that introduced the other edit components? Just want to make sure this wasn't an accidental delete

PR #353, opened 2018-04-15
Adjust custom coverage fields behavior

cherewaty:
## Purpose
When a user clicks the "circle X" icon to remove custom coverage from a package or resource, they should get a warning that saving changes will remove the previously set date range(s). The message was being shown too eagerly - any time a user removed the last date range row, they saw the "Saving will remove" message, even if they didn't previously have any custom coverage dates set.

## Approach
Check the `initialValue` of `customCoverages` to determine whether to show the "Saving will remove custom coverage" message on removing all date range rows.

## Next Steps
For a package, `customCoverage` is an object with `beginCoverage` and `endCoverage` keys, vs. a resource, which has a `customCoverages` array. The `PackageCoverageFields` component uses a `redux-form` `FieldArray` to have the same UX as setting coverage dates on a resource. Instead of happening in the route, the transformation between an array and a single object should happen in the `PackageCoverageFields` component, since it's a purely presentational concern.

## Screenshot
localhost_3000_eholdings_resources_1 iphone 6_7_8

### DIFF
--- a/tests/resource-custom-coverage-test.js
+++ b/tests/resource-custom-coverage-test.js
@@ -135,6 +135,10 @@ describeApplication('ResourceCustomCoverage', () => {
           it('removes the new row', () => {
             expect(ResourceCoverage.dateRangeRowList().length).to.equal(1);
           });
+
+          it('does not display the saving will remove message', () => {
+            expect(ResourceCoverage.hasSavingWillRemoveMessage).to.be.false;
+          });
         });
       });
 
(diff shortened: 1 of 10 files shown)


tests/resource-custom-coverage-test.js
```diff
@@ -135,6 +135,10 @@ describeApplication('ResourceCustomCoverage', () => {
           it('removes the new row', () => {
             expect(ResourceCoverage.dateRangeRowList().length).to.equal(1);
           });
+
+          it('does not display the saving will remove message', () => {
```
wwilsman: If this _was not_ previously displayed, we should use `it.always` to ensure it doesn't get displayed after this test gets run.

tests/resource-custom-coverage-test.js
```diff
@@ -334,8 +338,8 @@ describeApplication('ResourceCustomCoverage', () => {
           return ResourceCoverage.dateRangeRowList(0).clickRemoveRowButton();
         });
 
-        it('displays the no rows left message', () => {
-          expect(ResourceCoverage.hasNoRowsLeftMessage).to.be.true;
+        it('displays the saving will remove message', () => {
```
wwilsman: If this _was_ previously displayed, we should use `it.always` to ensure it doesn't disappear after this test gets run.

PR #357, opened 2018-04-17
UIEH-259: Custom Package Creation

wwilsman:
## Purpose

[UIEH-259](https://issues.folio.org/browse/UIEH-259)

Sets up the route and view for creating a custom package, and adds fields for name, content type, and custom coverage to the creation form.

## Approach

While creating custom packages are not yet supported in the backend, we can utilize mirage to iterate on the frontend.

- The route for creating custom packages is `/eholdings/packages/new`.
- A navigation modal is shown if there are changes to the form and a user clicks "cancel"
- When a custom package is successfully created, the user is redirected to the show route for the newly created package and a toast success message is shown.
- The current fields were preexisting in the edit route, but re-utilized for this creation route.
- A `create` action was added to models which will trigger a request to create the record with the specified attributes. Once resolved, the default resolve reducer takes care of adding the new record to the store.

#### TODOs
- Backend needs to support custom package creation.
- Add a "new package" button in the UI to access this new route.

### DIFF
--- a/src/components/details-view-section/details-view-section.js
+++ b/src/components/details-view-section/details-view-section.js
@@ -1,10 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
 import styles from './details-view-section.css';
+
+const cx = classNames.bind(styles);
 
 export default function DetailsViewSection(props) {
   return (
-    <div className={styles['details-view-section']}>
+    <div className={cx('details-view-section', { 'hide-separator': !props.separator })}>
       <h3 className={styles['details-view-section-label']}>{props.label}</h3>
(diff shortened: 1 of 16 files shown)


src/components/details-view-section/details-view-section.js
```diff
@@ -1,10 +1,13 @@
...
 import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
 import styles from './details-view-section.css';
 
+const cx = classNames.bind(styles);
+
 export default function DetailsViewSection(props) {
   return (
-    <div className={styles['details-view-section']}>
+    <div className={cx('details-view-section', { 'hide-separator': !props.separator })}>
```
cherewaty: 👏

PR #358, opened 2018-04-17
UIEH-254: Add description field to custom title

SamKeathley:
## Purpose
Adds editable description to edit page and description field to resource view.

Resolves [UIEH-254](https://issues.folio.org/browse/UIEH-254)

### DIFF
--- a/tests/pages/resource-edit.js
+++ b/tests/pages/resource-edit.js
@@ -26,6 +26,10 @@ import Datepicker from './datepicker';
   isPeerReviewed = property('checked', '[data-test-eholdings-peer-reviewed-field] input[type=checkbox]');
   checkPeerReviewed = clickable('[data-test-eholdings-peer-reviewed-field] input[type=checkbox]');
 
+  descriptionField = value('[data-test-eholdings-description-textarea] textarea');
+  fillDescription = fillable('[data-test-eholdings-description-textarea] textarea');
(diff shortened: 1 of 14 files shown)


tests/pages/resource-edit.js
```diff
@@ -68,6 +72,13 @@ import Datepicker from './datepicker';
   fillPublisher = fillable('[data-test-eholdings-publisher-name-field] input');
   publisherValue = value('[data-test-eholdings-publisher-name-field] input');
   publisherHasError = hasClassBeginningWith('feedbackError--', '[data-test-eholdings-publisher-name-field] input');
+
+  // 1500+ character quote for character limit tests.
+  superLongQuote= `Trees cover up a multitude of sins. If you don't think every day is a good day - try missing a few. You'll see. Put light against light - you have nothing. Put dark against dark - you have nothing. It's the contrast of light and dark that each give the other one meaning. Water's like me. It's laaazy. Boy, it always looks for the easiest way to do things We might as well make some Almighty mountains today as well, what the heck.
```
cherewaty: Since it's test-specific, and doesn't model any of the DOM, I think it makes more sense to define `superLongQuote` in the test itself, not in the page object.

SamKeathley: Yeah, I wasn't sure if it was going to be used later in other tests. I'll go ahead and change that.

PR #359, opened 2018-04-17
Separate custom title editing from custom resource editing

cherewaty:
## Purpose
The EBSCO RM API tightly couples custom titles to custom resources.
 _A custom resource is a relationship between a custom package and a title. That title can be either managed or custom._

Besides being a relationship, a resource - managed or custom - has a few attributes of its own - custom coverage dates, embargo period, holding status, etc.

There is not an RM API endpoint for editing a title. Managed titles don't have any editable attributes. The only way to edit a custom title, which does have editable attributes, is by editing one of the resources that belong to it. Changing a title attribute like `titleName`, `description`, or `publisherName` on a resource will apply the same change to _all_ resources belonging to that custom title. That can cause unintended and unreported side effects.

## Approach
This PR attempts to separate title fields from resource fields in the UI. Now only resource-specific fields can be edited from `eholdings/resources/:id/edit`, and title attributes can be edited at `eholdings/titles/:id/edit`. Only custom titles have a link to `eholdings/titles/:id/edit`.

To get around the lack of `PUT /eholdings/titles/:id`, the front-end applies changes to title fields to the first resource belonging to that title with `PUT /eholdings/resources/:id`. It's a nasty band-aid that should be ripped off ASAP.

## Next Steps
- Build a `PUT /eholdings/titles/:id` in `mod-kb-ebsco` that can move the logic changing a title's first resource away from the front end.
- There doesn't appear to be a way to get around the lack of an RM API endpoint for creating custom titles without also creating a custom resource (`packageId` is required). The UI for creating a custom title will still have to require a package relationship to set up, and essentially will be creating a custom resource.

## Learning
https://developer.ebsco.com/docs

### Custom resource show
localhost_3000_ iphone 6_7_8 1

### Custom resource edit
localhost_3000_ iphone 6_7_8 2

### Custom title show
localhost_3000_ iphone 6_7_8 3

### Custom title edit
localhost_3000_ iphone 6_7_8 4

### Managed title show
localhost_3000_ iphone 6_7_8 5


mirage/serializers/resource.js
```diff
@@ -13,6 +13,8 @@ function mapResourceAttrs(hash, resource) {
   hash.attributes.contributors = resource.title.contributors;
   hash.attributes.identifiers = resource.title.identifiers;
   hash.attributes.subjects = resource.title.subjects;
+  hash.attributes.isPeerReviewed = resource.title.isPeerReviewed;
+  hash.attributes.description = resource.title.description;
```
keathleydavidj: Did not know about this bit. TIL

src/components/title/edit/title-edit.js
```diff
@@ -0,0 +1,153 @@
...
+      }).isRequired
+    }).isRequired
+  };
+
+  componentWillReceiveProps(nextProps) {
+    let wasPending = this.props.model.update.isPending && !nextProps.model.update.isPending;
+    let needsUpdate = !isEqual(this.props.initialValues, nextProps.initialValues);
+
+    if (wasPending && needsUpdate) {
+      this.context.router.history.push(
```
sadatay: couldn't you call `handleCancel` from here instead?  looks identical.

src/routes/title-edit.js
```diff
@@ -0,0 +1,107 @@
...
+      name,
+      isPeerReviewed,
+      publicationType,
+      publisherName,
+      description
+    } = values;
+
+    let resource = model.resources.records[0];
+
+    resource.name = name;
```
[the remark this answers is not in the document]
cherewaty: Yep, that's totally a better way to do it. Commit added.

src/routes/title-edit.js
```diff
@@ -0,0 +1,93 @@
...
+  componentDidUpdate(prevProps) {
+    if (!prevProps.updateRequest.isResolved && this.props.updateRequest.isResolved) {
+      this.context.router.history.push(
+        `/eholdings/titles/${this.props.model.id}`,
+        { eholdings: true }
+      );
+    }
+  }
+
+  titleEditSubmitted = (values) => {
```
carolegodfrey: Nice on making edit against first package in the front end at least temporarily until available in mod-kb-ebsco

PR #360, opened 2018-04-17
UIEH-261 Custom Package holding status

elrickvm:
## Purpose
With a custom-package as a resource librarian I want to be able to remove that custom-package from my holdings. 

## Approach
 - Ensure holding-status work within and not within a context of a Redux-Form
 - Make Toggle Switch accessible. Making sure the toggle is usable from Assistive Devices.
 - Report holding-status toggle / removal errors back to the user.
   - Add destroys Errors to Toast
   - Keeps user on current page to make a decision based on the failure.
 - Add the capability to to adjust holding-status from both inline and full-page edit mode.
 - Allow the custom package holding-status is adjusted from either direct access from pasting into URL or navigating after a search.
 - Redirect the user after a successful removal of a custom-package from either a inline, full-page and from either direct input of URL or navigating after a search.

#### TODOS and Open Questions.
1. Its possible that we may need to add a notification based on a successful removal as well.

#### Jira
https://issues.folio.org/browse/UIEH-261

### Successful removal of custom-package
  - Direct input of URL in address bar

### DIFF
--- a/src/components/toggle-switch/toggle-switch.css
+++ b/src/components/toggle-switch/toggle-switch.css
@@ -7,7 +7,12 @@
   width: 60px;
 
   & input {
-    display: none;
+    /* makes input visually hidden but accessible */
+    position: absolute;
+    clip: rect(0, 0, 0, 0);
+    white-space: nowrap;
+    clip-path: inset(50%);
+    border: 0;
   }
 
   & input:checked + .toggle-switch-slider {
--- a/src/components/toggle-switch/toggle-switch.js
+++ b/src/components/toggle-switch/toggle-switch.js
@@ -1,54 +1,67 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import styles from './toggle-switch.css';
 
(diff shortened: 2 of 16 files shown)


src/components/toggle-switch/toggle-switch.css
```diff
@@ -7,7 +7,12 @@
   width: 60px;
 
   & input {
-    display: none;
+    /* makes input visually hidden but accessible */
```
cherewaty: 👏

src/components/toggle-switch/toggle-switch.js
```diff
@@ -1,54 +1,67 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import styles from './toggle-switch.css';
 
 const cx = classNames.bind(styles);
 
-export default class ToggleSwitch extends Component {
+export default class ToggleSwitch extends React.Component {
```
cherewaty: Any reason for this change? The previous way seems preferable.

elrickvm: That somehow got in there in the Rebase...good catch!

PR #361, opened 2018-04-18
UIEH-308 Add pane header to results pane even if no search yet

cherewaty:
## Purpose
Making eHoldings more consistent with the `<SearchAndSort>` UX used by other FOLIO modules.

We'll need the header there to add "+ New" buttons for custom packages and titles.

## Approach
Add a pane header and mimic the `stripes-smart-components` `<NoResultsMessage>`, which unfortunately is tightly coupled to `<SearchAndSort>`.

I decided against reusing a `<ResultsPane>` for the empty state to avoid the scroll logic there.

## Known drawbacks
The "+ New" button that will soon appear in the header won't be accessible on small screens until conducting a search.

## Screenshots
localhost_3000_eholdings_packages_new

### DIFF
--- a/src/components/package/create/package-create.js
+++ b/src/components/package/create/package-create.js
@@ -2,7 +2,12 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { reduxForm } from 'redux-form';
 
-import { Button, PaneHeader } from '@folio/stripes-components';
+import {
+  Button,
+  IconButton,
+  PaneHeader,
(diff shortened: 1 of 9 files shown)


cherewaty:
Also added "New" button for package creation (https://issues.folio.org/browse/UIEH-308).

cherewaty:
Tests added, and for clarity, pane renamed from `no-results-pane` to `pre-search-pane`.

PR #363, opened 2018-04-19
[WIP] Don’t send payloads for DELETE requests

cherewaty:
## Purpose
`DELETE` requests shouldn't send a body.

Replaces https://github.com/folio-org/ui-eholdings/pull/362.

## Approach
Don't send the payload from the `destroy` action creator.


cherewaty:
Replaced by https://github.com/folio-org/ui-eholdings/pull/367

PR #365, opened 2018-04-22
UIEH-215 Search resource list within title detail view

cherewaty:
## Purpose
Add search icon button and access to filtering a list of resources ("packages") on a title detail view. https://issues.folio.org/browse/UIEH-215

## Approach
Builds on the work in https://github.com/folio-org/ui-eholdings/pull/346.

One minor change from that PR: if the title doesn't have any related packages to begin with, the search functionality doesn't even appear, which theoretically may only be a problem with mirage data. I don't think there's ever a title with production data that's not related to any packages.

#### TODOs
The same TODOs from PR #346 apply here:
- The inner search params are not persisted in the URL. This prevents us from deep linking searches within the details view. The current params are persisted in the route state, and in the future we can make this route use query params instead of it's state without changing any other props or components.
- There is no focus management for modals in stripes. Ideally, we would focus-trap the modal when open and focus back to the list when it is closed.
- There should be some indicator when search filters are active (UIEH-216), and another indicator showing how many packages are returned from the search (UIEH-148)

_Additionally, there's some needed back-end work._

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -146,7 +146,7 @@ export default function configure() {
   };
 
   this.get('/packages', searchRouteFor('packages', packagesFilter));
-  this.get('/providers/:id/packages', nestedResourceRouteFor('provider', 'packages', packagesFilter));
+  this.get('/providers/:id/packages', nestedResourceRouteFor('provider', 'packages', 'name', packagesFilter));
 
   this.get('/packages/:id', ({ packages }, request) => {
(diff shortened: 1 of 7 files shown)


cherewaty:
Closing, not doable for a while: https://issues.folio.org/browse/UIEH-215?focusedCommentId=29403&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-29403

PR #366, opened 2018-04-22
UIEH-214 Search resource list within package detail view

cherewaty:
## Purpose
Add search icon button and access to filtering a list of resources ("titles") on a package detail view. https://issues.folio.org/browse/UIEH-214

## Approach
Builds on the work in https://github.com/folio-org/ui-eholdings/pull/346.

#### TODOs
The same TODOs from PR #346 apply here:
- The inner search params are not persisted in the URL. This prevents us from deep linking searches within the details view. The current params are persisted in the route state, and in the future we can make this route use query params instead of it's state without changing any other props or components.
- There is no focus management for modals in stripes. Ideally, we would focus-trap the modal when open and focus back to the list when it is closed.
- There should be some indicator when search filters are active (UIEH-216), and another indicator showing how many packages are returned from the search (UIEH-148)

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -214,7 +214,38 @@ export default function configure() {
     return {};
   });
 
-  this.get('/packages/:id/resources', nestedResourceRouteFor('package', 'resources'));
+  let packageTitlesFilter = (resource, req) => {
+    let params = req.queryParams;
+    let type = params['filter[type]'];
+    let selected = params['filter[selected]'];
+    let filtered = true;
+
+    if (params.q) {
+      let searchString = params.q.toLowerCase();
+
+      if (params.searchfield === 'title' && resource.title.name) {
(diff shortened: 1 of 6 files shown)


mirage/config.js
```diff
@@ -214,7 +214,38 @@ export default function configure() {
     return {};
   });
 
-  this.get('/packages/:id/resources', nestedResourceRouteFor('package', 'resources'));
+  let packageTitlesFilter = (resource, req) => {
```
wwilsman: I would expect this to look more like the title search route below. For example, `mod-kb` would accept `filter[subject]=somesubject` instead of `searchfield=subject&q=somesubject`.

For the other related PR, I copied the existing package search route filter and used it in both package search and provider package search. And that's exactly what the backend does as well (uses the same filters & params for top-level and related searches)

PR #367, opened 2018-04-23
 Remove delete request payload

elrickvm:
## Purpose
When sending a DELETE request there doesn't need to be a request body sent with the request. 

## Approach
Remove the payload from the request epic.
deleterequestnopayload

### DIFF
--- a/src/redux/data.js
+++ b/src/redux/data.js
@@ -130,21 +130,18 @@ const resolve = (request, body, payload = {}) => {
   let records = [];
   let ids = [];
 
+  // on request where neither a body or payload is sent
+  // such as a delete request we need to pick id off request
+  if (!data && request.params.id) {
+    data = { id: request.params.id };
+  }
+
   if (Array.isArray(data)) {
     records = records.concat(data);
     ids = records.map(({ id }) => id);
   } else if (data) {
     records = [data];
     ids = [data.id];
-  } else {
-    // on request where neither a body or payload is sent
-    // such as a delete request we need to pick id off request
-    records = [
-      {
-        id: request.params.id
-      }
-    ];
-    ids = [request.params.id];
   }
 
   if (body && body.included) {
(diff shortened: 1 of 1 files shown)


wwilsman:
We can include the change from #363 as well and close that one.

elrickvm:
Ok. Will pull that change in and see what happens. That branch was failing. If I remember correctly, removing the payload from the Action creator caused the Epic to fail because it tried to reference a payload in the Epic that was not passed in. But i will verify if that is still the case.

wwilsman:
Yea, it looks like the change you have here is the reference that would fail. Otherwise we only reference `payload` when it is a `POST` or `PUT`

elrickvm:
Sorry I said Epic when it was suppose to be Resolve.

elrickvm:
Its weird without the payload in the in the action creator it affects the form. So, when you are on a package edit route and toggle then save. Everything goes through but now for some reason the form thinks it has unsaved data and pops up the navigationModal ask to confirm unchanged data.

elrickvm:
I think i may have a solution @wwilsman in the last commit

elrickvm:
I was able to craft a solution in the last commit.

cherewaty:
👍 that looks right to me now, nice work!

src/redux/data.js
```diff
@@ -134,9 +133,18 @@ const resolve = (request, body, payload) => {
   if (Array.isArray(data)) {
     records = records.concat(data);
     ids = records.map(({ id }) => id);
-  } else {
+  } else if (data) {
     records = [data];
     ids = [data.id];
+  } else {
```
wwilsman: Instead of this additional `else`, lets remove the previous `if (data)` addtion and make `data` default to `{ id: package.params.id }`.

Something like this above the first `if` statement should work:
```js
if (!data && request.params.id) {
  data = { id: request.params.id };
}
```

wwilsman: On second thought, let's keep `else if (data)` for sanity. But still create the default `data` above the first `if` statement

elrickvm: Nice!

PR #370, opened 2018-04-24
UIEH-341 Smooth out back button logic

cherewaty:
## Purpose
When clicking to edit a package, title, or package-title, there's not reason to leave the three-pane search results layout.

When leaving to a full-page layout, the back button was occasionally not being shown, because we were not passing the `eholdings: true` state through `stripes-components` `<IconButton>`s.

https://issues.folio.org/browse/UIEH-341

## Next Steps
The `IconButton` `href` prop actually accepts a router location object, even though its `PropTypes` throw a warning if the `href` isn't a string:
```
Warning: Failed prop type: Invalid prop `href` of type `object` supplied to `IconButton`, expected `string`.
```


cherewaty:
Tests added. Thanks @wwilsman for not letting me get by without tests 😄

PR #374, opened 2018-04-26
[UIEH-258] Edit / view custom resource URLs

Robdel12:
## Purpose

This PR adds the ability for you to edit & view a URL on a custom resource (which is determined by looking at if it has a custom title). You cannot edit the URL on a managed resource, only on a custom resource. 

## Approach

Implement editing the URL for a custom packages resource ([custom-resource-edit.js](https://github.com/folio-org/ui-eholdings/blob/master/src/components/resource/edit-custom/custom-resource-edit.js))  field.  Given the way the [code set up currently](https://github.com/folio-org/ui-eholdings/blob/master/src/components/resource/edit.js#L7) we should _not_ add the URL field to the `EditManaged` component. 

Label: Custom URL
Field Type: textbox
Field is not required
URL validation: URL must begin with http:// or https://
Character limit: 600

## Learning
Here's some things I learned when implementing this ticket. It may or may not be helpful to you: 

- It would be nice to link to previous examples of this being implemented
- KB —> RM API —> mod-kb-ebsco
	- mod-kb-ebsco only exists to smooth over RM api
		- Could be named `mod-eholdings-ebsco-rm-api`
- Managed title/package: something your knowledge base provides (like kb-ebsco)
- Custom title/package: something that you defined
- A resource links a title to a package. It specifies how I access the content (which is the title)
	- If I buy a MEDLINE package and also have a ProQuest package that both have the “Medicine and Health Rhode Island” title there will be two resources that define how to access “Medicine and Health Rhode Island” for both ProQuest & MEDLINE. 

## Screenshots

image


mirage/config.js
```diff
@@ -258,7 +258,14 @@ export default function configure() {
   this.get('/packages/:id/resources', nestedResourceRouteFor('package', 'resources'));
 
   this.get('/resources/:id', ({ resources }, request) => {
-    return resources.find(request.params.id);
+    let resource = resources.find(request.params.id);
+
+    // if the ID is 1, it'll be a custom title
+    if (request.params.id === '1') {
+      resource.update('isTitleCustom', true);
```
cherewaty: This looks like it'll get it us into trouble. Could this be handled in the `resource` `serializer`'s `mapResourceAttrs()` instead?

Robdel12: I like it. I pushed up a change for that

mirage/serializers/resource.js
```diff
@@ -1,6 +1,11 @@
 import ApplicationSerializer from './application';
 
 function mapResourceAttrs(hash, resource) {
+  // if the ID is 1, it'll be a custom title
+  if (resource.id === '1') {
```
cherewaty: It's still hard-coded to that ID :( 

What if you did `hash.attributes.isTitleCustom = resource.title.isTitleCustom;` here?

Robdel12: Ah, how would we knowingly get to a custom title / resource? Would it be with a query param?

Robdel12: Ah, I'll just remove it then.

src/components/resource/edit-custom/custom-resource-edit.js
```diff
@@ -86,6 +87,11 @@ class CustomResourceEdit extends Component {
           actionMenuItems={actionMenuItems}
           bodyContent={(
             <form onSubmit={handleSubmit(onSubmit)}>
+              <DetailsViewSection
+                label="Resources"
```
cherewaty: Let's label it "Resource information" just like on the show route.

PR #376, opened 2018-04-27
 Upgrade bigtest/interaction to bigtest/interactor

cherewaty:
## Purpose
Upgrade from `bigtest/interaction` to `bigtest/interactor`.

## Approach
`bigtest/interactor` now exposes an `isPresent` on the root of interactors, so the `isRootPresent` helper is no longer necessary.

### Other improvements:
- `.interaction` no longer needed for chaining
- `once()` has been replaced with `when()`
- interaction methods now always have the selector as the first param; I updated `hasClassBeginningWith` to work the same way

## Learning
https://github.com/bigtestjs/interactor/tree/v0.4.1


tests/package-create-test.js
```diff
@@ -61,11 +61,7 @@ describeApplication('PackageCreate', () => {
       return PackageCreatePage
         .fillName('My Package')
         .addCoverage()
-        // Currently, `.dateRangeRowList(0)` immediately throws when
-        // it does not exist. When `@bigtest/interaction` is updated,
-        // this `.once().do()` pattern shouldn't be necessary.
-        .once(() => PackageCreatePage.dateRangeRowList(0))
-        .do(list => list.fillDates('12/16/2018', '12/18/2018'))
+        .append(PackageCreatePage.dateRangeRowList(0).fillDates('12/16/2018', '12/18/2018'))
```
wwilsman: We don't even need `.append` for this anymore 😃 

```js
return PackageCreatePage
  .fillName('My Package')
  .addCoverage()
  .dateRangeRowList(0).fillDates('12/16/2018', '/12/18/2018')
  .save();
```

tests/pages/navigation-modal.js
```diff
@@ -1,24 +1,11 @@
...
-      return false;
-    }
-  });
-};
-
-@page class NavigationModal {
-  exists = isRootPresent();
-  isVisible = isVisible();
+@interactor class NavigationModal {
+  isModalVisible = isVisible();
```
wwilsman: `isVisible` is one of the new default properties on Interactors. We shouldn't need to do this part anymore

tests/pages/package-search.js
```diff
@@ -78,7 +76,7 @@ import { isRootPresent } from './helpers';
     isSelected: computed(function () {
       return this.isSelectedText === 'Selected';
     }),
-    isActive: is('[class*="is-selected"]'),
+    isActive: is(this.$root, '[class*="is-selected"]'),
```
wwilsman: Because of the way instance initializers work, I'm pretty sure `this` refers to the instance of `PackageSearchPage`. So It's weird that this works. The old way should be the right way, since if you omit the selector it defaults to the current instance's `$root`.

If the old way doesn't work and it for-sure is not an actual failure, then this is most definitely a bug in `@bigtest/interactor`.

Same goes for the other few times this happens (I noticed it's always with `is`)

tests/pages/navigation-modal.js
Robdel12: What does that mean?

wwilsman: Interactors have some default properties now. `isPresent`, `isVisible`, `text`, `value`, etc. Basically the property helpers that can be called with no arguments don't have to be.

tests/pages/package-search.js
cherewaty: `isActive: is('[class*="is-selected"]')` fails with `Error: unable to find "[class*="is-selected"]"`

wwilsman: 🐛 

Fix incoming

tests/pages/navigation-modal.js
cherewaty: Fixed this one 👍

tests/custom-package-edit-test.js
```diff
@@ -125,11 +125,11 @@ describeApplication('CustomPackageEdit', () => {
 
     describe('entering invalid data', () => {
       beforeEach(() => {
-        return PackageEditPage.interaction
-          .once(() => PackageEditPage.dateRangeRowList().length > 0)
+        return PackageEditPage
+          .when(() => PackageEditPage.dateRangeRowList().length > 0)
           .do(() => {
```
wwilsman: FWIW, these are the `.when().do()` blocks we should be able to get rid of. Below, `.dateRangeRowList(0)` is lazy, so it will only continue if the `0` indexed item exists.

tests/pages/package-search.js
wwilsman: Fix released. Try `@bigtest/interactor` `0.4.2`

cherewaty: New `interactor` release put in to fix `is()`.

tests/custom-package-edit-test.js
cherewaty: I got rid of as many `.when().do()`s as I could 😁

PR #379, opened 2018-04-30
UIEH-304 Create Custom Titles

wwilsman:
## Purpose

[UIEH-304](https://issues.folio.org/browse/UIEH-304) 

There should be a route where we can create custom titles just like the custom package creation route.

## Approach

The custom title creation route is `/eholdings/titles/new`. It includes a navigation modal that appears when attempting to navigate away with unsaved changes, and an error toast when a server error is encountered. When the custom title is successfully created, the title show page for the new title is shown with a success toast.

The existing fields that can be found when editing a custom title were included in the creation form. Additionally, a package select field was added. This field requires us to get a list of custom packages when the route component is mounted. For now, we can develop and iterate on this page using mirage, but the backend will need support for querying packages matching `filter[custom]=true`.

Two important things to note:

- Custom packages can possibly be paginated. For now, I've set the page size to 1000. But a more future-proof implementation might include a type-ahead field.

- In order to pass the `packageId` with a title, the title's model needs to specify it as an attribute. We can either leave this as is, alter the data-layer to allow non-attributes, or create another model representing a new custom title.

## Screenshots
title-create

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -128,6 +128,7 @@ export default function configure() {
     let params = req.queryParams;
     let type = params['filter[type]'];
     let selected = params['filter[selected]'];
+    let custom = params['filter[custom]'];
     let filtered = true;
 
     if (params.q && pkg.name) {
@@ -140,6 +141,11 @@ export default function configure() {
 
     if (filtered && selected) {
       filtered = pkg.isSelected.toString() === selected;
+    }
+
+    if (filtered && custom) {
+      // packages don't always have `isCustom` defined
+      filtered = pkg.isCustom ? custom === 'true' : custom === 'false';
     }
 
     return filtered;
@@ -254,12 +260,16 @@ export default function configure() {
     return titles.find(request.params.id);
   });
 
-  this.post('/titles', ({ titles }, request) => {
+  this.post('/titles', (schema, request) => {
     let body = JSON.parse(request.requestBody);
-    let title = titles.create(body.data.attributes);
+    let { packageId, ...titleAttrs } = body.data.attributes;
+    let title = schema.titles.create(titleAttrs);
+    let pkg = schema.packages.find(packageId);
 
     title.update('isSelected', true);
-    title.update('isCustom', true);
+    title.update('isTitleCustom', true);
+
+    schema.resources.create({ package: pkg, title });
 
     return title;
   });
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -254,6 +254,16 @@ export default function configure() {
     return titles.find(request.params.id);
   });
 
+  this.post('/titles', ({ titles }, request) => {
+    let body = JSON.parse(request.requestBody);
+    let title = titles.create(body.data.attributes);
+
+    title.update('isSelected', true);
+    title.update('isCustom', true);
+
+    return title;
+  });
+
   // Resources
   this.get('/packages/:id/resources', nestedResourceRouteFor('package', 'resources'));
 
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -262,14 +262,17 @@ export default function configure() {
 
   this.post('/titles', (schema, request) => {
     let body = JSON.parse(request.requestBody);
-    let { packageId, ...titleAttrs } = body.data.attributes;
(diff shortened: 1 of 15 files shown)


mirage/config.js
```diff
@@ -254,6 +260,20 @@ export default function configure() {
     return titles.find(request.params.id);
   });
 
+  this.post('/titles', (schema, request) => {
```
cherewaty: Do we need to `POST /resources` or `POST /titles`? https://github.com/folio-org/mod-kb-ebsco/pull/109

src/components/title/create/title-create.js
```diff
@@ -0,0 +1,131 @@
...
+        />
+
+        <div className={styles['title-create-form-container']}>
+          <form onSubmit={handleSubmit(onSubmit)}>
+            <DetailsViewSection label="Title information" separator={false}>
+              <NameField />
+              <PublisherNameField />
+              <PublicationTypeField />
+              <DescriptionField />
+              <PackageSelectField packages={customPackages} />
```
cherewaty: Let's move the `PackageSelectField` into a new `DetailsViewSection` with label `Package information`.

mirage/config.js
carolegodfrey: We need to POST /resources. Can only create a custom title (resource) in the context of a custom package

wwilsman: I'm a little confused. I thought we were creating that post to RM API from the backend of `/titles`?

Since a new title should post to `/title` on any other API.

Unless this entire PR is meant to be a custom resource and not a custom title?

mirage/scenarios/default.js
```diff
@@ -2,37 +2,34 @@
...
-    name: 'Atlanta A&T Library'
+    name: 'Atlanta A&T Library',
+    packagesTotal: 1
   });
 
   let customPackage = server.create('package', {
+    provider: customProvider,
     name: 'Atlanta A&T Drumming Books',
     contentType: 'AggregatedFullText',
+    isSelected: false,
```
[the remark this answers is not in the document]
wwilsman: You're right. This is just the scenario that was already not working before. I just updated it so things would be linked together properly. It isn't actually used for testing. Only when `--mirage` is passed to `yarn start` without a scenario.

I'll change this though so it isn't confusing when booting with mirage

mirage/config.js
wwilsman: My assumption was that we would create a custom title from the frontend with `POST /titles`. But the backend creates a resource (which creates the title at the same time).

Then when we link additional packages to a title, it creates more resources

wwilsman: I think this is the last blocker for this PR.

If we're artificially splitting titles and resources into different routes, I think `mod-kb-ebsco` needs to do that too. Should it not?

The eventual goal is to have this frontend work with other backends. That's going to be infinitesimally harder if we have to do weird things specific to RM API like "create a new title by posting to resources."

Since `mod-kb-ebsco` is supposed to smooth out this weirdness so the frontend can be sane, shouldn't we have a `POST /titles` for creating _titles_?

Again, I know we're actually creating a resource at RM API's level, but if we couple the frontend _that_ tightly to the weirdness of RM API, that is going to be hell for us when it comes to supporting other knowledge bases in the future (if we're able to at all)

cherewaty: From the whiteboard session we had a few weeks ago, I think the determination was that `PUT /titles` we could do, but `POST /titles` just isn't really possible in a sane way without RM API changes. Because a title can't be created without a package relationship, the object being created is really a resource (with a new title being created simultaneously).

wwilsman: Hence the `packageId` sent with this payload. Couldn't RM API just map `POST /titles` to `POST /resources` behind the scenes?

Besides the `packageId` field, I don't see a difference in POSTing to titles or resource besides one sounds like it should work and the other feels gross.

If we really don't want to have `POST /titles`, I think this should become a new resource page instead of a new title page (`/eholdings/resources/new`).

I'm pretty against having new titles post to resources when we can abstract that away in `mod-kb-ebsco`. We can either call them resources in the frontend too, or add that POST to `/titles`.

PR #381, opened 2018-05-01
UIEH-256 Edit identifiers on a custom title

cherewaty:
## Purpose
Users should be able to edit the identifiers of a custom title.

Finishes https://issues.folio.org/browse/UIEH-256.

## Approach
The server wants separate type and subtype fields; we merge them into one for the UI.

The UI implementation leans heavily on `redux-form`'s `FieldArray`, like we've previously done with custom coverage dates.

### DIFF
--- a/src/components/title/edit/title-edit.js
+++ b/src/components/title/edit/title-edit.js
@@ -13,6 +13,7 @@ import NameField, { validate as validateName } from '../_fields/name';
 import EditionField, { validate as validateEdition } from '../_fields/edition';
 import PublisherNameField, { validate as validatePublisher } from '../_fields/publisher-name';
 import PublicationTypeField from '../_fields/publication-type';
+import IdentifiersFields, { validate as validateIdentifiers } from '../_fields/identifiers';
 import DescriptionField, { validate as validateDescription } from '../_fields/description';
 import ContributorField, { validate as validateContributor } from '../_fields/contributor';
 import PeerReviewedField from '../_fields/peer-reviewed';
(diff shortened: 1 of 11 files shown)


src/components/title/edit/title-edit.js
```diff
@@ -123,12 +124,19 @@ class TitleEdit extends Component {
...
+                />
+
                 <EditionField />
                 <PublisherNameField />
                 <PublicationTypeField />
-                <ContributorField
-                  initialValue={initialValues.contributors}
+
+                <IdentifiersFields
+                  initialValue={initialValues.identifiers}
```
wwilsman: Not a blocker by any means, but just a question: Why do we need `initialValue` here? I thought `redux-form` did that for us when we pass `initialValues` to the form.

cherewaty: @wwilsman it's a weird strategy for this use case:
If I have a previously saved identifier, open the edit route, and clear all the identifier rows, I get a message warning me that saving will remove all the identifiers. There's probably a more elegant way to calculate that state.

wwilsman: Ok, intersting. I thought `redux-form` was mapping the props just like this behind the scenes, but who knows what it actually does. Lol

src/routes/title-edit.js
```diff
@@ -48,15 +48,48 @@ class TitleShowRoute extends Component {
...
+    });
+  }
+
   titleEditSubmitted = (values) => {
     let { model, updateResource } = this.props;
     let resource = model.resources.records[0];
-    updateResource(Object.assign(resource, values));
+    updateResource(Object.assign(resource, {
+      ...values,
+      identifiers: this.expandIdentifiers(values.identifiers)
```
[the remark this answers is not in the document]
cherewaty: There's probably a way with `redux-form`'s `change()` prop, where the inputs derive off of component state and are faux-connected to `redux-form`. But that's... not pretty. I'd rather do it this way for now, even if it means we'll have to do it again in the `title-create` route.

PR #383, opened 2018-05-01
UIEH-253 - Add edition field to custom title edit page

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-253, users should be able to add or edit an edition for a custom title and view the same on title-record as well as resource-record pages.

## Approach
- Created a component called "edition" in _fields in title
- Used that component in title-edit, resource-show and title-show
- Added needed config for mirage
- Added unit tests

#### TODOS and Open Questions
- [ ] When we refresh a title, after editing "edition" or "description" or "peer-reviewed", those values are getting reset in the redux layer; coordinating with @cherewaty for ideas
- [ ] This PR takes care of editing edition from title-edit page, we might want to do something similar on title-create page after Wil's PR is merged.
- [x] Unable to create gif because title search is timing out on me. Will try and create one tomorrow.

## Screenshots

edition_custom_title

### DIFF
--- a/src/components/title/_fields/edition/edition-field.js
+++ b/src/components/title/_fields/edition/edition-field.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import { Field } from 'redux-form';
+
+import { TextField } from '@folio/stripes-components';
+import styles from './edition-field.css';
+
+export default function EditionField() {
+  return (
+    <div
+      data-test-eholdings-edition-field
+      className={styles['edition-field']}
+    >
+      <Field
+        name="edition"
+        component={TextField}
+        label="Edition"
+      />
+    </div>
+  );
+}
+
+export function validate(values) {
+  const errors = {};
+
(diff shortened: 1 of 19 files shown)


src/components/title/_fields/edition/edition-field.js
```diff
@@ -0,0 +1,30 @@
...
+        label="Edition"
+      />
+    </div>
+  );
+}
+
+export function validate(values) {
+  const errors = {};
+
+  if (values.edition && values.edition.length > 250) {
```
wwilsman: Do we want to also validate that they didn't pass an empty string?

`"    "` would pass this validation for example

sduvvuri-ebsco:
Thanks for your feedback @wwilsman, RM API accepts empty values for edition but in any case, added a check as well as test to ensure that empty spaces are not allowed as input.

PR #384, opened 2018-05-02
UIEH-312 Refactor custom embargo UI

cherewaty:
## Purpose
The custom embargo fields were pretty tricky to use before:
- To clear them out, you had to input zero or set the unit to "None."
- When opening a resource edit form without embargo set, you had to touch those fields for the form to not error.

Resolves https://issues.folio.org/browse/UIEH-312

## Approach
Introduces a close X icon button to remove the custom embargo period, just like custom coverage dates.

## Learning
Had to make use of `redux-form`'s `change()` so edits to the embargo fields from the close X icon button are persisted to the redux store. Really strange pattern that makes me uncomfortable. I wish all that was just in component state.

## Next Steps
Package custom coverage should be refactored to adopt the patterns established here, so we no longer have to misuse `redux-form`'s `FieldArray` to get the UX we want.


src/components/resource/_fields/custom-embargo/custom-embargo-fields.js
```diff
@@ -38,20 +58,47 @@ export default class CustomEmbargoFields extends Component {
...
-            }}
+          />
+        </div>
+
+        <div
+          data-test-eholdings-custom-embargo-remove-row-button
+          className={styles['custom-embargo-clear-row']}
+        >
+          <IconButton
+            icon="hollowX"
```
Robdel12: Can we add an aria-label to convey want they're removing?

src/components/resource/_fields/custom-embargo/custom-embargo-fields.js
```diff
@@ -25,8 +46,7 @@ export default class CustomEmbargoFields extends Component {
           <Field
             name="customEmbargoValue"
             component={TextField}
-            type="number"
-            parse={value => (!value ? null : (Number.isNaN(Number(value)) ? '' : Number(value)))}
+            placeholder="Number"
```
Robdel12: Could we add an `autoFocus` to this when it's a brand new field?

cherewaty: Thanks @Robdel12 🙌 `ariaLabel` and `autoFocus` added.

src/components/resource/_fields/custom-embargo/custom-embargo-fields.js
```diff
@@ -25,8 +46,8 @@ export default class CustomEmbargoFields extends Component {
...
+            autoFocus={initialValue.customEmbargoValue === 0}
```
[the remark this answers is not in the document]
cherewaty: Thankfully, no, because the fields won't show up if the `customEmbargoValue` is initially 0.

PR #386, opened 2018-05-03
[UIEH-255] Edit contributor on a custom title

Robdel12:
## Purpose

This PR implements the ability to edit a custom titles contributor. 

## Approach

I took the same route as the custom coverages date pickers and it worked out pretty well. These fields are keyboard accessibile too.


cherewaty:
Looking good. One more behavior addition: with other `FieldArray`s, we have an additional state.

If a user visits `resources/:id/edit` with some existing contributors, then removes all of them, show the message "Nothing set. Saving will remove any previously set contributors."

Example from identifiers: https://github.com/folio-org/ui-eholdings/pull/381/files#diff-3e2b4b2657cce5ea1f53e65709fc9e10R35

Robdel12:
@cherewaty good catch, I added that in

src/components/title/edit/title-edit.js
```diff
@@ -124,6 +126,9 @@ class TitleEdit extends Component {
                 <EditionField />
                 <PublisherNameField />
                 <PublicationTypeField />
+                <ContributerField
```
cherewaty: Definitely a nit, but we should probably fix that typo to prevent future copy-paste confusion. `Contributer` -> `Contributor`

PR #390, opened 2018-05-04
WIP to try and order such that list loads ahead of detail record - UIEH-357

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-357, for both managed and custom titles, we see that the detail-record loads ahead of the list as a result of which in the redux data layer, search list record overrides the detail record data and unfortunately, search list records might not have all fields of a detail-record. Effect is that certain fields are missing on detail record.

## Approach
To alleviate the situation above, try to order the events such that detail record does not load until the search completes rendering. @wwilsman suggested these changes and is a work in progress.

#### TODOS and Open Questions
- [ ] I might have misunderstood your suggestion and might be doing something wrong. While the provider detail is still present in the third pane, this change is not able to load list of packages for a given provider. 
- [ ] When I tried to debug, search is not returning the last search request as we expected.
- [ ] If this solution works for providers, then port it over for packages and titles.

### DIFF
--- a/src/routes/provider-show.js
+++ b/src/routes/provider-show.js
@@ -15,6 +15,7 @@ class ProviderShowRoute extends Component {
       }).isRequired
     }).isRequired,
     model: PropTypes.object.isRequired,
+    search: PropTypes.object,
     resolver: PropTypes.object.isRequired,
     getProvider: PropTypes.func.isRequired,
     getPackages: PropTypes.func.isRequired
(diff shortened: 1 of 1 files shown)


wwilsman:
I don't think this PR is necessary anymore since #409, right?

sduvvuri-ebsco:
yes, you are right. we do not need this anymore, will delete it.

PR #392, opened 2018-05-08
UIEH-322: Alter Modal css rule to enable scrolling of content

keathleydavidj:
## Purpose
When performing a package search from within a provider record, the modal that pops up with the search interface simply cuts off the overflowing filter radio inputs.

Resolves: [UIEH-322](https://issues.folio.org/browse/UIEH-322)

## Approach
I've changed the Modal component content area from `overflow: hidden` to `overflow: auto`.

### DIFF
--- a/src/components/modal/modal.css
+++ b/src/components/modal/modal.css
@@ -17,7 +17,7 @@
   background-color: #fff;
   max-height: calc(80vh);
   width: 100%;
-  overflow: hidden;
+  overflow: auto;
   border-radius: var(--radius, 4px);
   box-shadow: 0 5px 28px 1px rgba(0, 0, 0, 0.2);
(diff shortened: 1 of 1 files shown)


src/components/modal/modal.css
```diff
@@ -25,6 +25,10 @@
   outline: 0;
 }
 
+.scroll {
+  overflow: scroll;
```
wwilsman: This will _always_ show the scrollbar regardless if it's needed or not. `auto` will only show it if necessary.

keathleydavidj: Fair, let me get that fixed up.

cherewaty: So I think we won't even need the `scrollable` prop: we just always have `overflow: auto`.

PR #385, opened 2018-05-02
UIEH-298 Custom and Managed Package Full Page Edit (Visibility)- WIP

carolegodfrey:
[UIEH-298 - Edit Custom Package Detail Record : Display Visibility toggle on Full Page Edit Mode](https://issues.folio.org/browse/UIEH-298) 
[UIEH-324 - Managed Package Detail: Add Additional Fields to Full Page Editing](https://issues.folio.org/browse/UIEH-324)
[UIEH-339 - Edit: unselected Package Detail Records](https://issues.folio.org/browse/UIEH-339)

## Purpose
Updates are required to `Custom Package` Full Page Edit to support the following added field
- Visibility Toggle

Updates are required to `Managed Package` Full Page Edit to support the following added fields
- Visibility Toggle
- Holding Status
- Automatically Select New Titles
When Holding Status is toggled OFF, other fields should appear as READ only

## Approach
All fields added in this PR are Toggles. 
Maintain logic as is in place for Package Show In-Line edits
Used similar logic for Modal for Holding Status (managed packages) as was done for custom packages
Fields can only be edited if the Package isSelected

#### TODOS and Open Questions
- [ ] Currently having an issue with Unit tests for Managed Packages (manage-packaged-edit-test.js), Specifically, when visiting the edit page for a managed packaged that is selected. The test visually appears that the package is NOT  selected. Need to resolve this issue and (once resolved) add more detailed tests
- [ ] Toggle switches are not in their own separate _fields folder/file. Assume this is ok since eventually migrating away from toggle and no specific validations are needed
- [ ] Make Package Name and Content Type READ ONLY on Custom Package Edit when isSelected is false 

## Screenshots
custom-package-fullpage

PR #393, opened 2018-05-08
UIEH-358 - Custom Package Full Page Edit (Name/Content Type Disable when deselected)

carolegodfrey:
[UIEH-358 - Edit: unselect Package Details Records (package name/content type)](https://issues.folio.org/browse/UIEH-358)

## Purpose
When users are on the full page edit form for a custom package, they can toggle the packages Holding Status from Selected to Unselected.  The toggle is changed on the form, but not yet committed to the backend. The commit is only done once the form is submitted via a Save button. In the case of Unselecting a Custom Package and committing via Save, the package is essentially deleted. When the toggle appears as unselected on the full page edit form (and has not yet been submitted), all other editable fields should be disabled. The majority of this work was addressed in the following PR: 
https://github.com/folio-org/ui-eholdings/pull/385 
This PR addresses the disabling of 2 Custom Package Fields (Package Name and Content Type)

## Approach
Updated Name and Content type fields to appear as readonly KeyValue fields when Holding Status is UnSelected.
Similar to Coverage Dates and Visibility, the check is done in the CustomPackageEdit form 
Moved Custom Package Edit Selection tests into a separate test file from Custom Package Edit Tests

#### TODOS and Open Questions
- [ ] Looked into moving isReadonly logic into Name and Content-Type _fields components (as is visible in earlier commits to this branch). I opted to keep the logic in the form - solution seems simpler. But interested in other opinions on this and what are best practices

## Learning
https://github.com/erikras/redux-form
## Screenshots
custom-package-edit-fields-visible
Below scenario - shows edit of inputs, deselect (name and content type revert to original values) and save of custom package (package is deleted)
edit-and-save

### DIFF
--- a/src/components/package/edit-custom/custom-package-edit.js
+++ b/src/components/package/edit-custom/custom-package-edit.js
@@ -6,6 +6,7 @@ import isEqual from 'lodash/isEqual';
 import {
   Button,
   Icon,
+  KeyValue
 } from '@folio/stripes-components';
 import { processErrors } from '../../utilities';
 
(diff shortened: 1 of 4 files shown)


carolegodfrey:
Thanks for your help @elrickvm - squashed commit and removed the merge commit

PR #396, opened 2018-05-09
UIEH-355 Hide editable areas when managed-title in managed-package is unselected

elrickvm:
## Purpose

Managed-Titles and Managed-Packages are capable of being 'unselected' from a users electronic holdings and not not viewable to a patron but still present and able to be interacted with by a eHoldings application user. However, the level of interaction a user can have with a title differs from when the title is 'selected' vs 'unselected'. Until a title is 'selected'  the user should not be able to make any customization to that title because titles that are not selected are essentially not added to the user eHoldings and giving them permission to make customizations. 
 
## Approach
 - Hide / Disable editable sections until the title is "selected".

## Jira
 - https://issues.folio.org/browse/UIEH-355

### DIFF
--- a/src/components/resource/edit-managed-title/resource-edit-managed-title.js
+++ b/src/components/resource/edit-managed-title/resource-edit-managed-title.js
@@ -150,6 +150,8 @@ class ResourceEditManagedTitle extends Component {
                   data-test-eholdings-resource-holding-status
                   htmlFor="managed-resource-holding-toggle-switch"
                 >
+                  <h4>{managedResourceSelected ? 'Selected' : 'Not selected'}</h4>
+                  <br />
                   <Field
                     name="isSelected"
                     component={ToggleSwitch}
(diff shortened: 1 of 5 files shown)


src/components/resource/edit-managed-title/resource-edit-managed-title.js
```diff
@@ -162,26 +162,38 @@ class ResourceEditManagedTitle extends Component {
...
               >
-                <CustomCoverageFields
-                  initialValue={initialValues.customCoverages}
-                />
+                {managedResourceSelected ? (
+                  <CustomCoverageFields
+                    initialValue={initialValues.customCoverages}
+                  />
+                ) : (
+                  <p>Add the package to holdings to set custom coverage dates.</p>
```
cherewaty: Text for all of these should be `Add the resource to holdings` instead of `package`.

elrickvm: So right!

src/components/resource/edit-managed-title/resource-edit-managed-title.js
```diff
@@ -162,26 +162,38 @@ class ResourceEditManagedTitle extends Component {
               <DetailsViewSection
```
carolegodfrey: On the full page edit page - we should Selected/Hidden below Holding Status and Above Toggle - should we have it here as well to be consistent?

tests/resource-edit-managed-title-test.js
```diff
@@ -38,6 +38,37 @@ describeApplication('ResourceEditManagedTitleInManagedPackage', () => {
...
+      beforeEach(() => {
+        return ResourceEditPage.toggleIsSelected();
+      });
+
+      it('should not display the coverage button', () => {
+        expect(ResourceEditPage.hasAddCustomCoverageButton).to.be.false;
+      });
+
+      it('displays a message to add title to holdings', () => {
+        expect('');
```
carolegodfrey: Wasn't sure on the intent for this

carolegodfrey:
Changes look good - this is most likely a backend issue unrelated to this PR - but when I commit changes for select toggle I am seeing Invalid ContributorList error. ( See this in production as well so not new ). 
Also will there be a separate PR for resource-edit-custom-tile as well?

elrickvm:
Thats a great question @carolegodfrey. I am not sure. Can you ever get to a custom-title in a state of "unselected". For consistency it can be added to make custom-title behave the same. So when you "unselect" it hides the sections.

tests/resource-edit-managed-title-test.js
elrickvm: We were going to add a tests there then changed our mind.

carolegodfrey:
Once you unselect and submit, the custom resource gets removed from the package - so you are right in that it is not a permanent state to have a custom resource that is unselected. It would basically be for the case when you are on the full page edit page and happen to toggled back and forth between selected and unselected. Just a thought to keep them consistent and this is what I had done at the package level

elrickvm:
Yeah that is true. Consistency is King @carolegodfrey . I'll add it.

PR #397, opened 2018-05-09
Remove package view logic from route

elrickvm:
## Purpose
A Routes responsibility is to pass along data and just return a View not determine which kind of View to return. View branching logic belongs in the View. The View receives data and then looks at that data to determine what kind of View it should return to the Route. 

## Approach
 - Move View Logic from the Route to a intermediate View that will contain the branching logic to determine what kind of View to return.

### DIFF
--- a/src/routes/package-edit.js
+++ b/src/routes/package-edit.js
@@ -7,8 +7,7 @@ import { createResolver } from '../redux';
 import Package from '../redux/package';
 import Resource from '../redux/resource';
 
-import ManagedPackageEdit from '../components/package/edit-managed/managed-package-edit';
-import CustomPackageEdit from '../components/package/edit-custom/custom-package-edit';
+import View from '../components/package/edit';
 
 class PackageEditRoute extends Component {
   static propTypes = {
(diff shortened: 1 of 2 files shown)


elrickvm:
Now that i am thinking of it. Would these be better to be name PackageEditView? And then also change the Resource ones as well?

src/components/package/package-view/package-view.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
cherewaty: Let's put this branching logic for the different `View` and `initialValues` in `src/components/package/edit.js` instead of creating this new `package-view` component. That was the intention of `src/components/package/edit.js` - I'm not sure at what point we diverged from that.

src/routes/package-edit.js
```diff
@@ -7,8 +7,7 @@ import { createResolver } from '../redux';
 import Package from '../redux/package';
 import Resource from '../redux/resource';
 
-import ManagedPackageEdit from '../components/package/edit-managed/managed-package-edit';
-import CustomPackageEdit from '../components/package/edit-custom/custom-package-edit';
+import PackageView from '../components/package/package-view';
```
cherewaty: Then here: `import View from '../components/package/edit`;

src/components/package/package-view/package-view.js
elrickvm: Sounds good.

PR #398, opened 2018-05-09
Refactor Resource View logic

elrickvm:
## Purpose
Refactor the Resource View Logic and Move View Logic to a consistent location as other places in the application.

## Approach
 - Use the edit component to contain the branching logic for which view is returned.


src/routes/resource-edit.js
```diff
@@ -6,7 +6,9 @@ import moment from 'moment';
 import { createResolver } from '../redux';
 import Resource from '../redux/resource';
 
-import ResourceView from '../components/resource/resource-view';
+// import ResourceView from '../components/resource/resource-view';
```
cherewaty: Commented line can go bye bye.

elrickvm: Sharp eye!

PR #400, opened 2018-05-09
UIEH-216: Show amount of filters applied to search

SamKeathley:
This resolves [UIEH-216](https://issues.folio.org/browse/UIEH-216)

## Purpose
Show the number of filters applied when  searching within a list on a detail record.

## Approach
Added the `Badge` component from `stripes-components` next to the search icon. The filter count updates live with the rest of the filter list. Does not include `sort option` in count.


src/components/details-view/details-view.js
```diff
@@ -198,6 +199,12 @@ export default class DetailsView extends Component {
 
     let historyState = router.history.location.state;
 
+    const countFilters = (params) => {
+      const { q, filter = {} } = params;
+      const appliedFilters = [q, ...Object.values(filter)].filter(val => !!val);
```
wwilsman: We should use `let` for these. In the rest of the app we use `const` for top-level constants and `let` for anything else that's scoped to a block.

tests/provider-show-package-search-test.js
```diff
@@ -47,6 +47,10 @@ describeApplication('ProviderShow package search', () => {
     it('shows the package search modal', () => {
       expect(ProviderShowPage.searchModal.isPresent).to.be.true;
     });
+
+    it('displays 0 filter count', () => {
```
wwilsman: Instead of displaying `0`, I think we should just not display a badge at all

wwilsman:
This looks awesome! Just a couple minor things 👍

src/components/details-view/details-view.js
SamKeathley: Man, I always forget this haha. You're right.

src/components/details-view/details-view.js
```diff
@@ -198,6 +199,12 @@ export default class DetailsView extends Component {
 
     let historyState = router.history.location.state;
 
+    let countFilters = (params) => {
+      let { q, filter = {} } = params;
+      let appliedFilters = [q, ...Object.values(filter)].filter(val => !!val);
```
wwilsman: Little know trick here: instead of `.filter(val => !!val)`, we can just do `.filter(Boolean)` to filter out any falsey values. Either way works, I just wanted to share 😄

src/components/details-view/details-view.js
```diff
@@ -198,6 +199,12 @@ export default class DetailsView extends Component {
 
     let historyState = router.history.location.state;
 
+    let countFilters = (params) => {
```
wwilsman: Now that I'm looking at this again, I don't think this needs to be a function at all since it's called immediately below, and it's always with `searchParams`.

We could probably just move the logic into a single assignment:

```js
let filterCount = [searchParams.q]
  .concat(Object.values(searchParams.filter))
  .filter(Boolean).length;
```

PR #401, opened 2018-05-14
UIEH-148: Display total number of package search results for provider-show package search list.

keathleydavidj:
## Purpose
As per [UIEH-148](https://issues.folio.org/browse/UIEH-148) this PR displays the number of results for a package search from within the provider record detail page.

## Approach
In the `details-view` I've had to pass in an additional property containing the actual number to be displayed as this information does not seem accessible on the `model` or anywhere else already available to that component. Please let me know if that isn't the case as it seems silly to add a property simply to display a number.

#### TODOS and Open Questions
- [x] Is there a set of localization strings that I should be adding this "${#} Records found" string to?
I can see this might be a concern in the future on [UIEH-351](https://issues.folio.org/browse/UIEH-351)

## Screenshots
screen shot 2018-05-16 at 12 05 02 pm


cherewaty:
Let's try moving the "records found" message underneath the "Packages" title so it's farther away from the number of filters indicated in https://github.com/folio-org/ui-eholdings/pull/400

src/components/details-view/details-view.js
```diff
@@ -272,7 +274,14 @@ export default class DetailsView extends Component {
               data-test-eholdings-details-view-list={type}
             >
               <div className={styles['list-header']}>
-                <h3>{capitalize(listType)}</h3>
+                <div>
+                  <h3>{capitalize(listType)}</h3>
+                  {resultsLength && (
+                    <div data-test-eholdings-details-view-results-count>
+                      <p><small>{resultsLength} records found</small></p>
```
cherewaty: I realized we'll have to deal with the singular version here if only 1 is found, but we can deal with that when we start to look at localizing strings.

PR #404, opened 2018-05-15
UIEH-372 - Implement GET handling of root-proxies on settings

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-372, on the settings/root-proxy page, user should be able to see a dropdown with list of available root proxy options to choose from.

## Approach
- Create a component called root-proxy-select-field which can get a collection of root-proxies hitting the /eholdings/root-proxies endpoint in mod-kb
- Use this component to populate view of settings-root-proxy
- Establish route for handling the GET
- Associated unit tests

#### TODOS and Open Questions
- [ ] Had an issue trying to implement the PUT for the user to be able to update root-proxy selection. A new ticket has been created to address the same - https://issues.folio.org/browse/UIEH-212

## Learning
- A lot about routes and redux data layer as part of debugging

## Screenshots
root_proxy_get


src/components/settings-root-proxy/settings-root-proxy.js
```diff
@@ -0,0 +1,55 @@
...
+      rootProxies
+    } = this.props;
+
+    return (
+      <SettingsDetailPane
+        paneTitle="Root proxy"
+      >
+        <h3>Root Proxy Setting</h3>
+
+        {rootProxies.isLoading ? (
```
carolegodfrey: Nice to have this effect here!

PR #405, opened 2018-05-15
[UIEH-359] Add visibility toggle to custom resource

Robdel12:
## Purpose
Add the ability to toggle the visibility on a custom resource and a custom package. 

From the looks of it custom package full page edit already has the visibility toggle

## Approach

This follows the same pattern as toggling a selected package.


cherewaty:
This covers one of the two cases we need a visibility toggle: we also need one on `src/components/resource/edit-custom-title/resource-edit-managed-title.js`

I double-checked UIEH-359 to make sure that's in scope, and it is.

PR #410, opened 2018-05-24
UIEH-365 Remove inline edit of coverage statement

carolegodfrey:
[UIEH-365 -Remove Inline Editing - Title+Package Show - Coverage Statement Field ONLY](https://issues.folio.org/browse/UIEH-365) 

## Purpose
This PR partially implements removal of inline editing for title packages (aka resources) by removing the inline edit of coverage statement. 
## Approach
- Removed coverage-statement form and associated css. This functionality is instead available in full page edit (resource-edit-custom-title and resource-edit-managed-title
- Implemented logic for coverage statement as follows:

1.  Displays coverage statement if one has been set for the resource
2.  Displays 'No coverage statement has been set.' if not set
3.  Displays 'Add the resource to holdings to set a coverage statement.' if isSelected = false

- Updated Visibility text message to be plain text
- Updated existing tests, removing inline edit functionality
- Confirmed tests for full page editing

#### TODOS and Open Questions
- [ ] Not part of PR - Seems like some redundancy in the following 2 test files
(managed-resource-edit-test.js and resource-edit-managed-title-test.js). Can consolidate as part of follow on PR?


cherewaty:
Good catch on the redundant tests - I think it must've gotten renamed at some point at the old version accidentally got reintroduced.

PR #411, opened 2018-05-24
[UIEH-354] Replace title management toggle with radios for managed packages

Robdel12:
## Purpose

This is the first step in replacing our toggles with radio buttons. This is because the toggle switches weren't obvious what state they were in.

## Approach
Remove the toggle switch and use `RadioButtonGroup` with redux form. This removed the weird complexity we had with the toggles. Now it's falling more inline with the way the rest of the form works. 

## TODO
- [ ] Unskip last test. For some reason when you reselect the package `add to kb` isn't true.

### DIFF
--- a/src/components/package/edit-managed/managed-package-edit.js
+++ b/src/components/package/edit-managed/managed-package-edit.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { reduxForm, Field } from 'redux-form';
 import isEqual from 'lodash/isEqual';
@@ -117,12 +117,6 @@ class ManagedPackageEdit extends Component {
       packageVisible: e.target.checked
     });
   }
-
-  handleAllowKbToAddTitlesToggle = (e) => {
-    this.setState({
-      packageAllowedToAddTitles: e.target.checked
-    });
-  };
 
   handleOnSubmit = (values) => {
     if (this.state.allowFormToSubmit === false && values.isSelected === false) {
@@ -245,9 +239,9 @@ class ManagedPackageEdit extends Component {
               </DetailsViewSection>
               <DetailsViewSection label="Title management">
                 {packageSelected ? (
-                  <div>
+                  <div className={styles['title-management-radios']}>
                     {packageAllowedToAddTitles != null ? (
-                      <React.Fragment>
+                      <Fragment>
                         <Field
                           label="Automatically select new titles"
                           name="allowKbToAddTitles"
@@ -265,7 +259,7 @@ class ManagedPackageEdit extends Component {
                             data-test-eholdings-allow-kb-to-add-titles-radio-no
                           />
                         </Field>
-                      </React.Fragment>
+                      </Fragment>
                     ) : (
                       <label
                         data-test-eholdings-package-details-allow-add-new-titles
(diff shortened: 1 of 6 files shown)


tests/managed-package-edit-add-new-titles-test.js
```diff
@@ -214,8 +215,10 @@ describeApplication('ManagedPackageEditAllowKbToAddTitles', () => {
         expect(PackageEditPage.isSelected).to.be.true;
       });
 
-      it('displays an ON Toggle for allow KB to add titles', () => {
-        expect(PackageEditPage.allowKbToAddTitles).to.be.true;
+      // TODO: Fix before merge
+      // Why is this not true? Something weird is going on.
+      it.skip('allow KB to add titles is selected true', () => {
+        expect(PackageEditPage.allowKbToAddTitlesRadio).to.be.true;
```
carolegodfrey: I think this is because in handleSelectionToggle we currently set packageAllowedToAddTitles=true, but the radio button state is using field name allowKbToAddTitles

Robdel12: This surfaced a bigger issue with my PR. I wanted to remove all references to setState but I don't think I'll be able to :/

src/components/package/edit-managed/managed-package-edit.js
```diff
@@ -54,13 +53,11 @@ class ManagedPackageEdit extends Component {
...
     if ((nextProps.initialValues.isSelected !== this.props.initialValues.isSelected) ||
-    (nextProps.initialValues.isVisible !== this.props.initialValues.isVisible) ||
-    (nextProps.initialValues.packageAllowedToAddTitles !== this.props.initialValues.packageAllowedToAddTitles)) {
+    (nextProps.initialValues.isVisible !== this.props.initialValues.isVisible)) {
       this.setState({
         ...this.state,
         packageSelected: nextProps.initialValues.isSelected,
-        packageVisible: nextProps.initialValues.isVisible,
-        packageAllowedToAddTitles: nextProps.initialValues.allowKbToAddTitles
+        packageVisible: nextProps.initialValues.isVisible
```
carolegodfrey: This looks good to me - so I assume once the other toggles are moved over to radios - we may be able to get rid of these other state settings as well?

src/components/package/edit-managed/managed-package-edit.js
```diff
@@ -83,9 +82,9 @@ class ManagedPackageEdit extends Component {
 
   handleSelectionToggle = (e) => {
     if (e.target.checked) {
+      this.props.change('allowKbToAddTitles', true);
```
sadatay: don't you also need to roll this back in `cancelSelectionToggle`?

src/routes/package-edit.js
```diff
@@ -103,7 +103,7 @@ class PackageEditRoute extends Component {
       }
 
       if ('allowKbToAddTitles' in values) {
-        model.allowKbToAddTitles = values.allowKbToAddTitles;
+        model.allowKbToAddTitles = values.allowKbToAddTitles === 'true'; // turn string into boolean
```
sadatay: does `!!` not work?

Robdel12: Nah, because JavaScript. The value coming from the radio is a string of either `"true"` or `"false"`. 
image

The best way to turn a string "boolean" to a real boolean is to triple equal compare it to string true haha.

src/components/package/edit-managed/managed-package-edit.js
Robdel12: Yess! I'm excited about that :D

src/components/package/edit-managed/managed-package-edit.js
Robdel12: 🤔 Hm, I don't think so. That is just to decide if you want to remove the package entirely. This modal: 
image

If you click yes it'll remove it and `allowKbToAddTitles` is irrelevant. If you click no then nothing changed :)

PR #412, opened 2018-05-29
[UIEH-373] Remove inline editing from package show 

SamKeathley:
[UIEH-373](https://issues.folio.org/browse/UIEH-373)

## Purpose
Continue efforts to remove inline editing from `package`.

## Approach
In addition to removing the inline editing, the PR deletes the `forms` directory and some tests that were specific to inline editing on the show page.

## Screenshots

NOTE: The toggle under `Automatically select new titles` has an animation problem that exists regardless of the changes in the PR.


SamKeathley:
It looks like my CircleCi failure is related to the threshold for the tests. The test files that I deleted were all testing for inline functionality. And the corresponding test cases for the edit view are already there with some minor changes being made. Does anyone have any suggestions as to what tests I need to add to bring the numbers back up?

carolegodfrey:
I think the problem is with the test files that are being deleted. Although inline editing is not happening on the pages, you still need to validate the logic is still correct and that elements are being display (read only) is as expected. For example the cases where hidden reason is being displayed

tests/custom-package-edit-visibility-test.js
```diff
@@ -51,24 +51,10 @@ describeApplication('CustomPackageEditVisibility', () => {
...
-        /*
-         * The expectations in the convergent `it` blocks
-         * get run once every 10ms.  We were seeing test flakiness
-         * when a toggle action dispatched and resolved before an
-         * expectation had the chance to run.  We sidestep this by
-         * temporarily increasing the mirage server's response time
-         * to 50ms.
-         * TODO: control timing directly with Mirage
-         */
-        this.server.timing = 50;
```
cherewaty: Best part of this PR: we don't have to jump through weird test timing hoops for the toggle switch animations anymore 😀

PR #414, opened 2018-05-30
[UIEH-386] Full Page Edit of Resource with Multiple Custom Coverage Dates Error

wwilsman:
## Purpose

[UIEH-386](https://issues.folio.org/browse/UIEH-386)

The missing `intl` from props was causing the validation to throw an error for coverage dates because `formatISODateWithoutTime` is dependent on `intl.formatDate()`.

## Approach

Provide `intl` as a prop in the route for editing resources

To cause a test failures, I only needed to add an additional coverage date to some existing tests. Providing the `intl` as described above caused them all to start passing again.

#### Open Questions
- Why is the validation run for multiple coverages but not single coverages?

## Screenshots

**Before**

before

**After**

after

### DIFF
--- a/src/routes/resource-edit.js
+++ b/src/routes/resource-edit.js
@@ -22,6 +22,7 @@ class ResourceEditRoute extends Component {
   };
 
   static contextTypes = {
+    intl: PropTypes.object.isRequired,
     router: PropTypes.shape({
       history: PropTypes.shape({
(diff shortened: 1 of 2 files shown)


sadatay:
oh my.  haven't looked at this form in a while.  fun stuff, nice catch 👍

PR #421, opened 2018-06-05
UIEH-351 Introduce translations file

cherewaty:
## Purpose
Resolves https://issues.folio.org/browse/UIEH-351

## Approach
Starts using `react-intl`'s `injectIntl()` higher-order component instead of getting `intl` from context. `injectIntl()` uses the old React context under the hood, but is intended to be forward-compatible when `react-intl` starts using the new context API. We should start phasing out uses of the old context API to be future-friendly (`react-router` being the biggest question mark).

### DIFF
--- a/src/components/search-paneset/search-paneset.js
+++ b/src/components/search-paneset/search-paneset.js
@@ -7,6 +7,8 @@ import {
   PaneMenu
 } from '@folio/stripes-components';
 import capitalize from 'lodash/capitalize';
+import { injectIntl, intlShape } from 'react-intl';
+
 import { qs } from '../utilities';
 import SearchPane from '../search-pane';
 import ResultsPane from '../results-pane';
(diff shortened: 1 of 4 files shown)


src/components/search-paneset/search-paneset.js
```diff
@@ -114,11 +116,7 @@ export default class SearchPaneset extends React.Component {
 
     let resultsPaneSub = 'Loading...';
     if (!isLoading) {
-      resultsPaneSub = `${intl.formatNumber(totalResults)} search result`;
-
-      if (totalResults > 1) {
-        resultsPaneSub += 's';
-      }
+      resultsPaneSub = intl.formatMessage({ id: 'ui-eholdings.resultCount' }, { count: totalResults });
```
wwilsman: Total nit, since this is > 80ch can we break this across multiple lines?

PR #424, opened 2018-06-07
UIEH-402 Complete Removal of Inline Editing for Title Package (Custom Coverage) 

carolegodfrey:
[UIEH-402 - Inline Editing - Title+Package Show (custom coverage dates)](https://issues.folio.org/browse/UIEH-402) 
[UIEH-409 - Custom Coverage Date Display Rules should be the same as Managed Coverage Date Display Rules](https://issues.folio.org/browse/UIEH-409) 
[UIEH-416 -Package and Resource Show Page : Update coverage labels](https://issues.folio.org/browse/UIEH-416) 

## Purpose

- This PR completes the removal of inline editing for title packages (aka resources) by removing the inline edit of custom coverage dates.  
- As part of the removal process, moved tests which are for inline edit to a full page equivalent test
- Updated custom coverage display logic to follow managed coverage rules - specifically displaying YYYY format for year only publication types.
- Update display on package and title package show pages to show label ‘Custom coverage dates’ instead of ‘Custom’
- Resolved 2 coverage date issues on full page edit
(1)When adding coverage dates from full page edit mode, observed saved dates were 1 day behind when displayed on show page
(2)validate function for coverage dates being outside package coverage dates was not working properly. No validation message was displayed in the form (only get error from back end)

## Approach

- Implemented read only logic for show page for custom coverage as follows:
(1) managed coverage dates (if available) always display
(2) custom coverage dates display (if available) and would only be available if isSelected = true
(3) if no managed or custom dates have been set - show message 'No coverage date has been set'

- Updated embargo logic for show page to be consistent with managed coverage as follows:
(1) managed embargo (if available) always display
(2) custom embargo displays (if available) and would only be available if isSelected = true
(3)if no managed or custom embargo(s) have been set - show message 'No embargo period has been set'

Added tz(‘UTC’) when saving and displaying date ranges from Datepicker
Updated some existing tests to display custom coverage in format of YYYY (due to publication type - Streaming Vidseo being a year only pub type) 

#### TODOS and Open Questions
- [ ] Still have skip on test for overlapping date range and entering multiple coverage dates for full page edit. Equivalent tests were skipped in inline edit test file. But have not got them to work
- [ ]  Concerns over duplication of test cases for both managed and custom resources (managed-resource-edit-custom-coverage-test.js and custom-resource-edit-custom-coverage-test.js). But this is the case for other fields as well.

## Screenshots
Show Page Display - Both Dates Present
screen shot 2018-06-07 at 1 57 53 pm

Show Page Display - No coverage dates set
screen shot 2018-06-07 at 2 07 32 pm

In Line Edit - Dates outside package date range
screen shot 2018-06-07 at 2 00 05 pm


cherewaty:
@carolegodfrey I think we may be able to simplify some of the date handling by leaning more on `react-intl` - I'll make a pass at that after you merge this.

PR #427, opened 2018-06-07
UIEH-212 Refactor root proxy to use new endpoints

elrickvm:
## Purpose
From the ui-eholdings application a user has the ability to pick from a list of available Root-Proxy and then set it. When a user reaches the settings page a list of available Root-Proxy is populated but when they have to send a request to update the Root-Proxy the request cannot contain the collection but rather a single root-proxy. This work here utilizes two new endpoints in mod-kb-ebsco to facilitate in the updating of what is deemed a singular resource in Root-Proxy. 

## Approach
 - Create two new models:
   - `Proxy-Types`: which is the collection of available of Root-Proxy
   - `Root-Proxy`: which is the Singular Root-Proxy and the Proxy that the user current has set or is trying to set from the UI.
 - Make use of the two new endpoints:
   - `eholdings/proxy-types`: will `GET` return a collection
   - `eholdings/root-proxy`: use to `GET` and `PUT` the user's root-proxy

### DIFF
--- a/mirage/config.js
+++ b/mirage/config.js
@@ -90,27 +90,38 @@ export default function configure() {
     return JSON.parse(request.requestBody);
   });
 
-  // Root Proxies
-  this.get('/root-proxies', {
+  // Current root proxy
+  this.get('/root-proxy', {
+    data:
+      {
+        id: 'eholdings/root-proxy',
+        type: 'rootProxies',
+        attributes: {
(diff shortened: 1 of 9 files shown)


elrickvm:
Thanks @sduvvuri-ebsco! You paved the way so it was smooth sailing!

PR #428, opened 2018-06-08
Use intl from react-intl instead of stripes context

cherewaty:
## Purpose
Now that we know it's on the fast train to deprecation, we want to avoid using the old React context API.

## Approach
Use `react-intl`'s `FormattedDate`, `FormattedNumber`, and `injectIntl` instead of `context.intl` or `stripes.context.intl`. When `react-intl` moves to the new context API under the hood, those abstractions should still work.

### DIFF
--- a/tests/custom-package-edit-test.js
+++ b/tests/custom-package-edit-test.js
@@ -93,7 +93,7 @@ describeApplication('CustomPackageEdit', () => {
         });
 
         it('displays new coverage dates', () => {
-          expect(PackageShowPage.customCoverage).to.equal('12/16/2018 - 12/18/2018');
+          expect(PackageShowPage.customCoverage).to.equal('12/16/2018 - 12/18/2018');
         });
 
         it('shows a success toast message', () => {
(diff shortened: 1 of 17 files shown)


tests/custom-package-edit-test.js
```diff
@@ -93,7 +93,7 @@ describeApplication('CustomPackageEdit', () => {
         });
 
         it('displays new coverage dates', () => {
-          expect(PackageShowPage.customCoverage).to.equal('12/16/2018 - 12/18/2018');
+          expect(PackageShowPage.customCoverage).to.equal('12/16/2018 - 12/18/2018');
```
cherewaty: I'm not exactly sure what happened with white space here, but copy/pasting got the tests to pass.

tests/package-custom-coverage-test.js
```diff
@@ -4,7 +4,7 @@ import { expect } from 'chai';
 import { describeApplication } from './helpers';
 import PackageShowPage from './pages/package-show';
 
-describeApplication('PackageCustomCoverage', () => {
+describeApplication.only('PackageCustomCoverage', () => {
```
wwilsman: ✂️

cherewaty: 🙈

PR #429, opened 2018-06-08
Update @bigtest/interactor and unskip tests

wwilsman:
## Purpose

With #424, came some tests that had to be skipped due to some strange behavior when trying to fill a second row of dates.

This was because, by default, nested interactors _are not_ scoped to the parent interactor when using the `new` keyword. Instead, we need to use the new `scoped` property helper from `@bigtest/interactor`.

## Approach

I added `@bigtest/react` to mount our test harness so it would be easier to debug our tests. With the old way, we had to use a `pause` because the `afterEach` was tearing down our application. This `pause` caused the runner to hang and it became cumbersome to have to completely stop and restart our tests after every use of `pause`.

The `setupAppForTesting` helper from `@bigtest/react` allows us to investigate and debug our application after a test _passes or fails_. Since we can now just use `only` to isolate a specific test, `pause` becomes obsolete and is only useful if you want to have to restart your tests every time.

---

For the original issue of the skipped tests, the following creates a nested interactor:

``` js
@interactor class SomeInteractor {
  foo = new FooInteractor('[date-test-foo]')
}
```

While the interactor is nested, _it is not scoped to `SomeInteractor`'s selector_. This was done purposely because everywhere we need a `Modal` interactor, the actual modal element _is not_ scoped within the parent element, but instead rendered elsewhere on the page. This allows us to have nested, unscoped interactors.

For nested _and scoped_ interactors, we need the `scoped` property helper which accepts an interactor class as the second argument.

To further complicate why the skipped tests were failing, the nested unscoped interactor was nested within another nested, but scoped, collection.

``` js
dateRangeRowList = collection('[data-test-eholdings-coverage-fields-date-range-row]', {
  beginDate: new Datepicker('[data-test-eholdings-coverage-fields-date-range-begin]'),
  endDate: new Datepicker('[data-test-eholdings-coverage-fields-date-range-end]'),
})
```

This meant that at any index of `dateRangeRowList`, the `beginDate` and `endDate` interactors reference _the exact same element_.

Using the `scoped` property helper keeps the elements scoped within their parent interactor element

``` js
dateRangeRowList = collection('[data-test-eholdings-coverage-fields-date-range-row]', {
  beginDate: scoped('[data-test-eholdings-coverage-fields-date-range-begin]', Datepicker),
  endDate: scoped('[data-test-eholdings-coverage-fields-date-range-end]', Datepicker),
})
```

## Learning
- https://github.com/bigtestjs/react
- https://bigtestjs.io/docs/interactor/#/scoped


tests/index.js
```diff
@@ -1,3 +1,5 @@
+import 'babel-polyfill';
```
wwilsman: This was necessary for the `async`/`await` syntax. We already have a indirect dependency on `babel-polyfill`, but I added it to the `package.json` anyway to be explicit.

PR #434, opened 2018-06-12
UIEH-352: Change package visibility toggle to radio inputs

keathleydavidj:
## Purpose
To be consistent with other Folio apps, we are changing the Visibility toggle to radio buttons.

This resolves [UIEH-352](https://issues.folio.org/browse/UIEH-352)

## Approach
There was some great work done for changing the visibility toggles in resources that helped guide my approach here for packages. 

## Screenshots
*_Before_*

*_After_*

### DIFF
--- a/src/components/package/show/package-show.js
+++ b/src/components/package/show/package-show.js
@@ -35,21 +35,24 @@ export default class PackageShow extends Component {
     queryParams: PropTypes.object
   };
 
+  static getDerivedStateFromProps(nextProps, prevState) {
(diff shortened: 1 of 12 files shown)


src/components/package/show/package-show.js
```diff
@@ -35,22 +35,25 @@ export default class PackageShow extends Component {
     queryParams: PropTypes.object
   };
 
+  static getDerivedStateFromProps(nextProps, prevState) {
```
Robdel12: very nice 😎

PR #435, opened 2018-06-13
[UIEH-388] Custom title adding identifiers and contributors

SamKeathley:
## Purpose
This change fixes the bug, outlined in [UIEH-388](https://issues.folio.org/browse/UIEH-388), that prevented the user from adding additional contributors and identifiers to a custom title. 

## Approach / Learning
Where I was hung up the most was in the tests. For the longest time I was unable to get the values of the _type_ of contributor and the _name_ of the contributor separately, as they appear to the user.
To do this, I ended up having to modify the components a touch to give the `type` and `name/id` separate data attributes as there wasn't a way to differentiate them in the test otherwise. 
Before, the test would return the text value of both the type and name combined for contributor, as the page object was looking for just any text in that row:
` expect(TitleShowPage.identifiersList(0).text).to.equal('ISSN (Online)1111'); `
 Now you can check for them separately.

### DIFF
--- a/tests/custom-title-edit-test.js
+++ b/tests/custom-title-edit-test.js
@@ -35,6 +35,13 @@ describeApplication('CustomTitleEdit', () => {
           type: 'author'
         }
       ],
+      identifiers: [
+        {
+          type: 'ISBN',
+          subtype: 'Online',
+          id: '11133394'
+        }
+      ],
       isTitleCustom: true
     });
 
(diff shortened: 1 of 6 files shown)


tests/custom-title-edit-test.js
```diff
@@ -91,6 +106,52 @@ describeApplication('CustomTitleEdit', () => {
...
+      describe('clicking save', () => {
+        beforeEach(() => {
+          return TitleEditPage.clickSave();
+        });
+
+        it('goes to the title show page', () => {
+          expect(TitleShowPage.$root).to.exist;
+        });
+
+        it('displays new author', () => {
```
carolegodfrey: Really, really minor comment on test - 'displays new editor'

PR #442, opened 2018-06-26
Upgrade stripes-components to 2.0.28

cherewaty:
Unblocks https://issues.folio.org/browse/UIEH-404 and https://issues.folio.org/browse/UIEH-406 by making changes from https://github.com/folio-org/stripes-components/pull/439 available in `ui-eholdings`.


sadatay:
😄 thanks jeffrey!

PR #443, opened 2018-06-26
UIEH-404 Refactor holding status away from ToggleSwitch

elrickvm:
## Purpose
UI-eHoldings users are able to add and remove `titles` also referred to as `resources` to their holdings. Here we are refactoring from using a toggle-switch to rather use a button to add and then removing is done from the action menu in the dropdown. This change and UX is more in line with what the rest FOLIO applications are doing so we conforming to ensure users have the same experience across the the various FOLIO applications. 

## Approach
 - Replace the ToggleSwitch component with Button component
 - Add the actions to the drop-down within the Resource Page header

### DIFF
--- a/src/components/resource/show.js
+++ b/src/components/resource/show.js
@@ -117,14 +117,14 @@ export default class ResourceShow extends Component {
 
     if (resourceSelected === true) {
       actionMenuItems.push({
-        'label': 'Remove title from Holdings',
+        'label': 'Remove title from holdings',
         'state': { eholdings: true },
         'onClick': this.handleHoldingStatus,
         'data-test-eholdings-remove-resource-from-holdings': true
       });
     } else if (resourceSelected === false) {
       actionMenuItems.push({
-        'label': 'Add to Holdings',
+        'label': 'Add to holdings',
         'state': { eholdings: true },
         'onClick': this.handleHoldingStatus,
         'data-test-eholdings-add-resource-to-holdings': true
--- a/src/components/resource/show.js
+++ b/src/components/resource/show.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import {
   Button,
   IconButton,
+  Icon,
   KeyValue,
   PaneMenu
 } from '@folio/stripes-components';
@@ -12,7 +13,6 @@ import DetailsView from '../details-view';
 import Link from '../link';
 import IdentifiersList from '../identifiers-list';
 import ContributorsList from '..//contributors-list';
-import ToggleSwitch from '../toggle-switch';
 import CoverageDateList from '../coverage-date-list';
 import { isBookPublicationType, isValidCoverageList, processErrors } from '../utilities';
 import Modal from '../modal';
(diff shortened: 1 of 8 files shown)


tests/pages/resource-show.js
```diff
@@ -18,6 +19,16 @@ import Toast from './toast';
   hasDeselectFinalTitleWarning = isPresent('[data-test-eholdings-deselect-final-title-warning]');
 }
 
+@interactor class ResourceShowDropDown {
+  clickDropDownButton = clickable('button');
+  isExpanded = attribute('button', 'aria-expanded');
```
Robdel12: 💯 😍

src/components/resource/show.js
```diff
@@ -116,6 +115,22 @@ export default class ResourceShow extends Component {
...
+    if (resourceSelected === true) {
+      actionMenuItems.push({
+        'label': 'Remove title from Holdings',
+        'state': { eholdings: true },
+        'onClick': this.handleHoldingStatus,
+        'data-test-eholdings-remove-resource-from-holdings': true
+      });
+    } else if (resourceSelected === false) {
+      actionMenuItems.push({
+        'label': 'Add to Holdings',
```
cherewaty: Switch to sentence case "Add to holdings"

src/components/resource/show.js
```diff
@@ -116,6 +115,22 @@ export default class ResourceShow extends Component {
       });
     }
 
+    if (resourceSelected === true) {
+      actionMenuItems.push({
+        'label': 'Remove title from Holdings',
```
cherewaty: Switch to sentence case: "Remove title from holdings"

tests/resource-selection-test.js
```diff
@@ -40,86 +40,124 @@ describeApplication('ResourceSelection', () => {
...
-            title: 'There was an error'
-          }]
-        }, 500);
-
+        /*
+         We need to slow the timing in order for the animation to have to time
+         to run. Reason being in test the Mirage timing is set to 0 so
+         requests resolve immediately.
+         * TODO: control timing directly with Mirage
+        */
```
cherewaty: Looking forward to figuring out a reliable way to do this 😎

src/components/resource/show.js
```diff
@@ -250,14 +265,23 @@ export default class ResourceShow extends Component {
...
-                    id="resource-show-toggle-switch"
-                  />
+                  { ((!resourceSelected && !isSelectInFlight) || (!this.props.model.isSelected && isSelectInFlight)) && (
+                    <Button
+                      buttonStyle="primary"
+                      onClick={this.handleHoldingStatus}
+                      disabled={model.destroy.isPending || isSelectInFlight}
+                      data-test-eholdings-resource-add-to-holdings-button
+                    >
+                      Add to holdings
```
cherewaty: Does this work nicely or is it too much loading status happening? 
`{isSelectInFlight ? 'Adding' : 'Add to holdings'}` 
instead of just
`Add to holdings`

elrickvm: I don't think it will add nor take away anything. We have the loading ellipsis and the disabling of the button happening. If the button remained blue then i would say yeah the `adding` text would help. But since is turns gray that is why i am on the fence here.

cherewaty: Cool, I'll defer to your instinct on that one.

PR #449, opened 2018-07-03
Upgrade eslint-config-stripes to 2.0.2

cherewaty:
- Cleans up some component declaration ordering.
- Disables the warnings for the remaining lifecycle hook deprecations.
- Adds `.eslintignore` to improve linting performance.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+/artifacts
+/dist
+/node_modules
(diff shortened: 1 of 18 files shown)


cherewaty:
The ordering stuff was a rule change that dropped in after your work on the lifecycle hooks - your approach was correct, the rules just got way more opinionated about the organization.

PR #450, opened 2018-07-05
Add data layer documentation

wwilsman:
## Purpose

The data layer in eHoldings is pretty unique. We should have some documentation about it

## Approach

There was some old documentation, but it was written before some refactoring happened and was also more about why we moved away from reducers and actions for each individual type of resource.

I gave a walkthrough of the data layer not too long ago, so I took my notes from that and added some paragraphs & examples. I didn't spend too much time on it, so it's not nearly as details as it could be, and just focuses on how to use the data layer and where to debug it.

### TODOs and Open Questions

- Not sure if I should elaborate somewhere I didn't. Hoping some reviews will suss that out.
- Also wasn't sure of where docs should live, so I just created a `docs` directory. Let me know if anybody thinks it should go somewhere else.

### DIFF
--- a/docs/data-layer.md
+++ b/docs/data-layer.md
@@ -2,7 +2,7 @@
 
 ## Redux
 
-In eHoldings, data is connected on `routes` only. This is done via
+In eHoldings, only `routes` are connected to data. This is done via
 `react-redux`'s `connect` decorator.
 
 ``` javascript
@@ -15,15 +15,15 @@ connect(({ eholdings: { data } }) => ({
 });
 ```
 
-When data is needed, you must first request it with one of the actions
+When in need of data, you must first request it with one of the actions
 defined in `mapDispatchToProps`. The data received in
-`mapStateToProps` will then update with a pending request, and update
+`mapStateToProps` will then update with a pending request, then update
 again once the request resolves or rejects.
 
 ## Models
 
-In eHoldings, you must define a model to request and store data in the
-redux store.
+You must first define a model to request and store data in the redux
+store.
 
 ``` javascript
 // src/redux/provider.js
@@ -48,8 +48,9 @@ export default model({
 
 ### Requesting & Resolving Data
 
-Using this model and the eHoldings resolver within `connect` allows us
-to request and receive data from the backend Okapi server.
+After defining a model, we can use the eHoldings `resolver` and
+react-redux's `connect` to allow us to make requests for and receive
+data from our backend API.
 
 ``` javascript
 // src/routes/provider-show.js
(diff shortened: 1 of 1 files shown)


sadatay:
>Also wasn't sure of where docs should live, so I just created a docs directory. Let me know if anybody thinks it should go somewhere else.

+1 for `docs`, it seems relatively standard across OSS that i've come across and its also in parity with mod-kb

docs/data-layer.md
```diff
@@ -0,0 +1,178 @@
+# Data Layer
+
+## Redux
+
+In eHoldings, data is connected on `routes` only. This is done via
```
sadatay: it'd be `data are` as `data` is woefully plural.  also you might wanna chuck in a link to the redux `connect` docs

docs/data-layer.md
```diff
@@ -0,0 +1,178 @@
...
+  // data needed by the component is accessed here
+}), {
+  // ...mapDispatchToProps
+  // actions used to send requests for data are defined here
+});
+```
+
+When data is needed, you must first request it with one of the actions
+defined in `mapDispatchToProps`. The data received in
+`mapStateToProps` will then update with a pending request, and update
```
sadatay: `will then update with a pending request, and update` minor nit but if you split this up it would be `The data received... will then update` and `The data recieved... and update again` so you should bridge these by subbing `and` for `then` or doing something weird like `will update with a pending request and, upon resolution of said request, will update once more`.

docs/data-layer.md
```diff
@@ -0,0 +1,178 @@
...
+again once the request resolves or rejects.
+
+## Models
+
+In eHoldings, you must define a model to request and store data in the
```
sadatay: `In eHoldings` is echoed and also probably provides context the reader already has given where they're reading this

docs/data-layer.md
```diff
@@ -0,0 +1,178 @@
...
+  // relationships are defined using `hasMany(type)` or `belongsTo(type)`
+  // the type defaults to the relationship key
+  packages = hasMany();
+}
+
+export default model({
+  type: 'providers',           // used as the key when storing data
+  path: '/eholdings/providers' // used in the URL during requests
+})(Provider);
+```
```
sadatay: :+1: these comments are super clear and illustrative

docs/data-layer.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Requesting & Resolving Data
+
+Using this model and the eHoldings resolver within `connect` allows us
```
sadatay: `eHoldings resolver` has not yet been defined.  and i dunno if i'd mention `okapi` specifically here since it's not _really_ a server, or at least not primarily.  the request goes through okapi but the data is coming from the backend module (mod-kb-ebsco, mod-go-kb, mod-et-cetera)

docs/data-layer.md
```diff
@@ -0,0 +1,178 @@
...
+
+When dealing with a collection of models, such as with `hasMany`, the
+returned collection object lazily looks up other pages from the same
+query in the data store. This allows you to map over all records that
+have been returned by a query regardless of page, such as when
+rendering the search results lists.
+
+For search results, the resolver must know the exact params you passed
+to the search action in order to properly find the records from the
+request. In the case of provider packages, the search params are
```
sadatay: is `provider package` synonymous with `managed package`?  don't think i've heard this term used much

wwilsman: `provider packages` as in packages belonging to a provider

wwilsman: Maybe I should change this to `In the case of a provider's packages`

sadatay: i think it becomes pretty clear further down but it couldn't hurt 🤷‍♀️

docs/data-layer.md
wwilsman: I tried rewording this to be an introduction to the resolver instead of pretending we already know what it is.

docs/data-layer.md
```diff
@@ -0,0 +1,178 @@
...
+    // will request packages at `/eholdings/provider/:id/packages?:params`
+    getPackages: (id, params) => Provider.queryRelated(id, 'packages', params)
+  }
+)(ProviderShowRoute);
+```
+
+The model returned by the resolver has request objects at
+`model.request`, `model.update`, and `model.destroy` which provide
+information about the different types of requests, such as
+`isPending`, `isResolved`, `isRejected`, `errors`, etc.
```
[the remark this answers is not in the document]
wwilsman: Took your advice with the show by adding some code and comments. It's not exactly example code, but I hope it better illustrates models and requests.

docs/data-layer.md
```diff
@@ -0,0 +1,205 @@
...
+to redux runs through request reducers to generate the request,
+followed by an observable epic to actually make the request, then that
+epic dispatches a resolve or reject action which runs back through the
+reducers to mark the request's status and add our data to the store.
+
+1. Request action is dispatched (`find`, `query`, etc.)
+2. Redux reducers create request object for the type of request and resource type
+3. The observable epic reacts to the action and sends the real request
+4. The observable epic dispatches `resolve` or `reject` in response to the request
+5. Redux reducers update the request object and store records from any resolved data
```
sadatay: this 👍

PR #455, opened 2018-07-09
Use webpack 4

cherewaty:
## Purpose
Start using webpack 4: https://github.com/folio-org/stripes-core/pull/347

## Approach
- We needed to update the version `ui-eholdings` requests of `webpack` too.
- `favicons-webpack-plugin` didn't like `ico` as the input - I swapped it out for an `svg` that scales very nicely.


ttamj:
You may also need the latest CLI.  I forgot to merge it earlier today.  https://github.com/folio-org/stripes-cli/pull/88

cherewaty:
Latest Stripes CLI added too.

PR #458, opened 2018-07-10
Add document titles

cherewaty:
## Purpose
Document titles are part of the essential UX for any browser-based app.

## Approach
Use `TitleManager` introduced in https://github.com/folio-org/stripes-core/pull/324

I don't love reaching into the `stripes-core` directory structure, but `stripes-core` doesn't have an `index.js` that exports anything.

### DIFF
--- a/src/routes/package-create.js
+++ b/src/routes/package-create.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import moment from 'moment';
+import TitleManager from '@folio/stripes-core/src/components/TitleManager';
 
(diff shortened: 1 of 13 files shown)


cowboyd:
With a routing structure like ui-eholdings, this _screams_ HoC to me.

For static titles:

```js
@title("New custom package")
class PackageCreateRoute extends Component {
   //...
}
```

For dynamic titles:
```js
@title(model => `Edit ${model.name}`)
class PackageEditRoute extends Component {
  //...
}
```

Robdel12:
@cowboyd that's what I was thinking. Or the pattern react-navigation takes. You specify a static object or function on your component called `navigationOptions` which contains the title and can be configured per component.

PR #461, opened 2018-07-12
UIEH-448: Add * for certain required fields in custom packages and custom titles

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-448, we are supposed to indicate that certain fields are required while creating or editing custom packages and custom titles.

## Approach
- We tried using the *required* prop in Stripes Components TextField and Select but they did not work per our expectations. We figured that setting those props had nothing to do with the needed asterisk. Moreover, if we set those props, we figured that the browser validation was superseding our redux-form validations.
- We modified our approach to just add an asterisk next to the label referring to what ui-inventory module does.
- I was able to verify the change locally for custom packages and new custom title creation but not edit page of custom title since there seems to be a different error in loading custom titles at the time of this PR.

#### TODOS and Open Questions
[ ] - I see that *Name* label is available in translations/ui-eholdings/en.json. Do we hav to use that in place of these labels to support i8ln? Even so, I would think we need to create two separate entries for names because these asterisks only need to appear on certain labels and not all *Name* labels. I am not sure if there is a separate story for that.

## Learning
- Usage of react dev tools for debugging
- A bit of insight into ui-inventory

## Screenshots
adding_asterisk

### DIFF
--- a/src/components/package/_fields/name/package-name-field.js
+++ b/src/components/package/_fields/name/package-name-field.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import { Field } from 'redux-form';
+import { FormattedMessage } from 'react-intl';
 
 import { TextField } from '@folio/stripes-components';
 import styles from './package-name-field.css';
@@ -15,7 +16,7 @@ export default class PackageNameField extends Component {
           name="name"
           type="text"
           component={TextField}
-          label="Name"
+          label={<FormattedMessage id="ui-eholdings.package.name.isRequired" />}
         />
       </div>
     );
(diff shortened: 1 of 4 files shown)


elrickvm:
That is good question in regards to the i8ln translation file and adding an entry for the required fields. I am not sure if you can concat the string in the jsx like `label="Name" + "*"`. If not then adding a new entry to `en.json` like `package.name.isRequired: Name *` is a valid option.

sduvvuri-ebsco:
Thanks @elrickvm for your feedback, verified that jsx does not allow string concatenation in labels. So added few more entries into en.json within the scope of this ticket and updated code using that. Please review at your convenience for internationalization. Thanks!

elrickvm:
screen shot 2018-07-12 at 12 31 06 pm

I wonder if we can use the current name but do something like this….and pass `isRequired` as a value….then in the `en.json` do the logic in there in the same string similar to was is done with resultCount?

PR #463, opened 2018-07-16
UIEH-407: Change holding status toggle into a button on package-edit

keathleydavidj:
## Purpose
The "Select / Deselect" toggle switch is totally out of place now on the edit page, so this PR is to bring the edit page back in line with the pattern already on the show page. This replaces the toggle with an "add" button when deselected and a set of actions in the pane header.

#### TODOS
- [x] manged package edit page
- [x] custom package edit page

## Screenshots
*Before*

*After*


cherewaty:
Same comments as https://github.com/folio-org/ui-eholdings/pull/465#issuecomment-406003594: maybe we should have a loading state in the Remove modal?

keathleydavidj:
@elrickvm Yeah, was having some real problems with copying over your implementation for keeping the modal open. Think I have it working now.

keathleydavidj:
Soooo, what happened to CircleCI?

elrickvm:
I think the subtitles for your animated gifs are mismatched 'before' should be 'after'

PR #464, opened 2018-07-17
UIEH-129: External Link Component in Resource Show

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-129, we are supposed to display an external link icon next to managed/custom url in resource show page. Implement that functionality.

## Approach
- Create a functional component named `ExternalLink`
- Use Stripes-components Icon component within the anchor link
- Add associated unit tests

#### TODOS and Open Questions
- [x] Icon size "small" supports size 16x16 where as our font size is 14x14. Per @cherewaty  suggestion, reached out to @filipjakobsen and @rasmuswoelk to find out if that is intentional or something that we can change to 14x14. Waiting on @rasmuswoelk response. 
- [x] Based on the above response, decide whether we need external-link.css for styling the icon with padding etc. because currently, since the size is bigger, it is not aesthetically pleasing.
- [x] Added a unit test to check if the icon is present on page, thought of adding one more to ensure that once its clicked, it opens link in a new tab but looks that the existing test for that is being skipped. Should we add any more tests?
- [ ] Ticket also mentions about contributing this component back to stripes, do we need to make any changes to be able to do that?

## Screenshots
updated_css

### DIFF
--- a/src/components/external-link/external-link.css
+++ b/src/components/external-link/external-link.css
@@ -1,0 +1,3 @@
+.external-link-icon {
+  vertical-align: middle;
+}
--- a/src/components/external-link/external-link.js
+++ b/src/components/external-link/external-link.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Icon } from '@folio/stripes-components';
+import classNames from 'classnames/bind';
+import styles from './external-link.css';
+
+classNames.bind(styles);
+
+export default function ExternalLink(props) {
+  return (
+    <a href={props.href} target={props.target} rel={props.rel}>{props.href}{' '}
+      <span data-test-eholdings-resource-show-external-link-icon className={styles['external-link-icon']}>
+        <Icon icon="external-link" size="small" color="currentColor" />
+      </span>
+    </a>
+  );
+}
+
+ExternalLink.propTypes = {
+  href: PropTypes.string,
+  target: PropTypes.string,
+  rel: PropTypes.string
+};
--- a/src/components/external-link/index.js
+++ b/src/components/external-link/index.js
@@ -1,0 +1,1 @@
+export { default } from './external-link';
--- a/src/components/resource/show.js
+++ b/src/components/resource/show.js
@@ -13,12 +13,14 @@ import {
 
 import DetailsView from '../details-view';
 import Link from '../link';
+import ExternalLink from '../external-link/external-link';
 import IdentifiersList from '../identifiers-list';
 import ContributorsList from '..//contributors-list';
 import CoverageDateList from '../coverage-date-list';
 import { isBookPublicationType, isValidCoverageList, processErrors } from '../utilities';
 import DetailsViewSection from '../details-view-section';
 import Toaster from '../toaster';
+
 
 export default class ResourceShow extends Component {
   static propTypes = {
@@ -253,7 +255,11 @@ export default class ResourceShow extends Component {
                 {model.url && (
                   <KeyValue label={`${model.title.isTitleCustom ? 'Custom' : 'Managed'} URL`}>
                     <div data-test-eholdings-resource-show-url>
-                      <a href={model.url} target="_blank" rel="noopener noreferrer">{model.url}</a>
+                      <ExternalLink
+                        href={model.url}
+                        target="_blank"
+                        rel="noopener noreferrer"
+                      />
                     </div>
                   </KeyValue>
                 )}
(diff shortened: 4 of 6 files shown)


sduvvuri-ebsco:
@elrickvm : Coordinating with Rasmus regarding the icon size. Icon should always be present when there is an external link, no cases when it should not be present. As I mentioned in PR, test that clicking on the link opens in a new tab is being skipped for some reason, so I guess this test is going to be similar and probably, we want it skipped. So, did not put that together.

elrickvm:
@sduvvuri-ebsco 
- Are you planning on holding off the merge until Rasmus releases a new icon size?
- Ah, I just looked at the tests case for the opening in a new tab that test was a tricky and i think it was failing because there was no clean way to test for opening for a new tab using window if i remember correctly.

sduvvuri-ebsco:
Rasmus approved our changes to Icon component small size and it is now 14x14. Observed how it looks with 14x14 and it still looks misaligned with text and bigger. So, added css that works. Updated .gif in PR.

src/components/external-link/external-link.css
```diff
@@ -0,0 +1,6 @@
+[class^="icon"] {
+  width: 12px;
+  height: 12px;
+  position: relative;
+  top: 1.5px;
+}
```
cherewaty: This CSS affects all elements in the app that have a class starting with the string `icon` (which is definitely not what we want, that gets a little funky). Try rebasing your branch again and re-`yarn`ing - it looks like the changes you made to `stripes-components` might not have made it here. This CSS file shouldn't be needed.

screen shot 2018-07-19 at 7 40 31 pm

sduvvuri-ebsco:
@cherewaty: Thanks for your feedback, its my bad that I used css that affected the entire app icons. I somehow thought it would be picked up only from within the specific .js file. I fixed the css to have a class and reference that within my .js file but I still feel its required. See the difference without and with css: 
Without: 
14x14
screen shot 2018-07-20 at 9 53 20 am
With css:
screen shot 2018-07-20 at 9 54 16 am
12x12

I feel that the icon with css has better alignment with text. Please let me know what you think?

src/components/external-link/external-link.css
```diff
@@ -0,0 +1,4 @@
+.external-link-icon {
+  position: relative;
+  top: 1.5px;
+}
```
cherewaty: You're right, we still have a need for a little CSS to make it look right.

Try this instead:
```
.external-link-icon {
  vertical-align: middle;
}
```

sduvvuri-ebsco:
@cherewaty : Tried the css tweak you suggested and it looks good. 
screen shot 2018-07-20 at 10 42 14 am
Updated code but I agree - probably we can wait for the other component to get merged or get this out and revisit when that gets merged. I am ok with either.

rasmuswoelk:
Hi @sduvvuri-ebsco 

I'm sure there's something I'm missing here but I'm just curious to hear the reasoning behind. We can't just use a regular link tag?

I think the component filename should be renamed to ExternalLink to align with what's written in https://github.com/folio-org/stripes-core/blob/master/doc/component-example.md ("As is conventional, the filename is the same as that of the component.")

As @cherewaty mentions I think https://github.com/folio-org/stripes-components/pull/507 will be useful for the purpose of displaying the icon inline with the label. It takes RTL in to account etc.

rasmuswoelk:
The "color"-prop of the Icon has been phased out.

sduvvuri-ebsco:
Hi @rasmuswoelk : We were using a regular link tag until the requirement changed to have an icon next to it. Instead of using an image tag for the icon, we are making use of the Icon component and at the time, we were not aware of the handy component you are developing. So, we thought that if we could develop an external link component, we could contribute it back to the community after we ensure it works as expected in ui-eholdings. 
Regarding the naming convention, all of the filenames in UIEH are following the same as we have in this PR. We shall have more discussion within the team regarding https://github.com/folio-org/stripes-core/blob/master/doc/component-example.md
I certainly agree that folio-org/stripes-components#507 is much more handy and we will pick it up when it is merged to stripes-components.
I am not sure I understand what you mean by `The "color"-prop of the Icon has been phased out.`. I see it as one of the props in Icon component. Thanks!

rasmuswoelk:
@sduvvuri-ebsco 

Alright 👍 

Regarding the color-prop: My mistake - it's getting phased out but it's still an option for now. I think for now it has been removed from the readme but in the future, it will be removed entirely.

You can use the global class for the SVG in the Icon: "stripes__icon" to style the color of the icon using :global(. stripes__icon)

rasmuswoelk:
And for the filename convention: I actually thought this PR was in stripes-components which were why I referenced the guidelines for stripes-components. Sorry.

I think it's time to go to weekend ;)

sduvvuri-ebsco:
Thanks for the heads up regarding the color prop being removed @rasmuswoelk. Have a great weekend :)

PR #465, opened 2018-07-17
UIEH-405 Refactor resource-edit page holding status

elrickvm:
## Purpose
eHoldings users are able to manage their holdings by adding or removing a resource from their holdings. Here we are also providing them that ability when in the 'edit' mode of a resource aka package-title. eHoldings has the concept of a managed-resource and a custom-resource and even though the behavior slightly differs between the two types the base is the same -- a user managing the status of their holdings.

Prior this area used a toggle switch but we are refactoring towards a explicit button and action
from the dropdown to conform with the UX from other FOLIO applications.

## Approach
To ensure that the eHoldings UX is inline with other FOLIO applications:
 - Add a button which allows a user to add a resource to their holdings when the resource is unselected.
 - Move the removal of a resource action into the dropdown menu.
## Jira
https://issues.folio.org/browse/UIEH-405

## Screenshots
*Custom Resource:*

*Managed Resource:*

### DIFF
--- a/tests/resource-edit-deselection-test.js
+++ b/tests/resource-edit-deselection-test.js
@@ -17,7 +17,7 @@ describeApplication('ResourceEditDeselection', () => {
     });
 
     title = this.server.create('title', {
-      publicationType: 'Streaming Video'
+      publicationType: 'Streaming Video',
     });
 
     providerPackage = this.server.create('package', {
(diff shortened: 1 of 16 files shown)


cherewaty:
When removing from holdings: should the modal stay up until the server responds?

Something like:
- Click "Yes, remove"
- "Yes, remove" becomes "Removing" and disabled
- When async finished, modal disappears, redirect happens, and toast pops up

tests/resource-edit-deselection-test.js
```diff
@@ -36,69 +36,67 @@ describeApplication('ResourceEditDeselection', () => {
...
+         * later in tests.
+         */
+        let resolveRequest;
+
+        beforeEach(function () {
+          this.server.put('/resources/:id', () => {
+            return new Promise((resolve) => {
+              resolveRequest = resolve;
+            });
+          });
```
cherewaty: 🔥

PR #466, opened 2018-07-17
UIEH-432 Replace fork of stripes-components modal

cherewaty:
## Purpose
Resolves https://issues.folio.org/browse/UIEH-432

`ui-eholdings` will now exclusively use the `stripes-components` `<Modal>`, not a fork.

## Approach
- Uses the new `<ModalFooter>` component to get nice layout for the buttons.
- The `<ModalFooter>` API doesn't feel real great - we might want to work on its composability at some later point.
- Fixed some issues where `<FormatMessage>` was being passed in to props that expected a string. Those use `intl.formatMessage()` now.
- Most of these modals fit the use case of a `<ConfirmationModal>` (https://github.com/folio-org/stripes-components/pull/471) - we might want to use that instead when that PR's merged.

#### TODO
- [x] Merge https://github.com/folio-org/stripes-components/pull/496. It's blocking this PR.

## Screenshots
localhost_3000_eholdings_packages_1 iphone 5_se


src/components/navigation-modal/navigation-modal.js
```diff
@@ -112,24 +111,18 @@ export default class NavigationModal extends Component {
...
-              <Button
-                fullWidth
-                onClick={this.dismiss}
-                buttonStyle="primary"
-                data-test-navigation-modal-dismiss
-              >
-                {dismissLabel}
-              </Button>
-            </div>
+            <ModalFooter
```
Robdel12: I agree, this probably should be a render prop instead :)

PR #467, opened 2018-07-18
UIEH-478: Localize providers show page

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-478, we are supposed to support localization in providers show page.

## Approach
- Navigate through code to find strings
- Add them to en.json, rename common ones to packages and providers like action menu items which will be shared
- Ensure that all tests pass

## Screenshots
provider_show_localize

### DIFF
--- a/src/components/provider-show/provider-show.js
+++ b/src/components/provider-show/provider-show.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { KeyValue } from '@folio/stripes-components';
-import { FormattedNumber } from 'react-intl';
+import { FormattedNumber, FormattedMessage } from 'react-intl';
 
 import { processErrors } from '../utilities';
 import DetailsView from '../details-view';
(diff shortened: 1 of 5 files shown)


src/components/provider-show/provider-show.js
```diff
@@ -44,14 +44,14 @@ export default function ProviderShow({
         paneTitle={model.name}
         actionMenuItems={actionMenuItems}
         bodyContent={(
-          <DetailsViewSection label="Provider information">
-            <KeyValue label="Packages selected">
+          <DetailsViewSection label={<FormattedMessage id="ui-eholdings.provider.providerInformation" />}>
```
cherewaty: This might throw an error into the console, but https://github.com/folio-org/ui-eholdings/pull/469 should fix it.

sduvvuri-ebsco: Thanks @cherewaty for making the change in #469

PR #470, opened 2018-07-19
STCOR-208 use new context API

rjruizes:
Accompanies this refactor https://github.com/folio-org/stripes-core/pull/358, removing legacy context.

PR test may fail since stripes-core PR hasn't been merged yet. If we need to support a temporary migration scenario, where stripes-core is not updated (using legacy context) and eHoldings is updated (using new context), let me know. Otherwise it's faster to just remove the legacy code from both repos at the same time.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -110,14 +110,14 @@
     history "^4.7.2"
 
 "@folio/eslint-config-stripes@^2.0.2":
-  version "2.0.200030"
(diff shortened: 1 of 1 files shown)


cherewaty:
Thanks! Looks fine. `ui-eholdings` has a lock file, so you can merge the `stripes-core` change first, then we can upgrade the version of `stripes-core` requested by `ui-eholdings` to get the tests passing.

PR #473, opened 2018-07-19
Use ReduxForm format() on Datepickers

cherewaty:
## Purpose
Changes in https://github.com/folio-org/stripes-components/pull/487 needed these updates to `Datepicker` usage.

## Approach
The `format()` prop on Redux Form `<Field>` takes care of the date transformation we need to do from `YYYY-MM-DD` to whatever the user's locale is.

Seems to work with the current iteration of the `Datepicker` as well, so might as well get it in.

## Learning
https://redux-form.com/7.4.2/docs/api/field.md/#-code-format-value-name-gt-formattedvalue-code-optional-

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "webpack": "^4.0.0"
   },
   "dependencies": {
-    "@folio/stripes-components": "3.0.1000441",
+    "@folio/stripes-components": "3.0.1000442",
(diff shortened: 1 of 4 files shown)


cherewaty:
Added commit that pulls in the latest `stripes-components` too.

PR #477, opened 2018-07-20
[UIEH-317] Search Within Modal : Apply Filters Updates

SamKeathley:
## Purpose
The purpose of this change is to allow users to search within a list of packages by filter only (without depending on a search string), and to persist within a single search without being interrupted by the modal closing before you're done. 

Resolves [UIEH-317](https://issues.folio.org/browse/UIEH-317)

## Approach
While implementing these changes @cherewaty and I made some design choices that fulfill the desired functionality from the ticket, but also adds concision to the modal that we both find easier to understand for the user. Instead of adding an `apply` link in the header, which would be hidden when no filter selections changed, we've left the `search` button as the only button that submits changes. We moved the `search` button into the `modalFooter` and have that entire footer hidden when no changes in the modal have been made at all. 

The justification for this slight design change is that it is more responsive on smaller screens by using the `stripes-components` modal and button design, and it cuts down on the possible confusion of looking for an `apply` link in the header or trying to click a greyed-out `search` button.

Additionally, you can use the escape key to cancel a search.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,8 +60,7 @@
     "impagination": "^1.0.0-alpha.3",
     "inflected": "^2.0.3",
     "react-intl": "^2.4.0",
-    "redux-actions": "^2.2.1",
-    "shallowequal": "^1.1.0"
+    "redux-actions": "^2.2.1"
   },
   "peerDependencies": {
     "@folio/stripes-core": "^2.10.0",
--- a/package.json
+++ b/package.json
@@ -60,7 +60,8 @@
     "impagination": "^1.0.0-alpha.3",
     "inflected": "^2.0.3",
     "react-intl": "^2.4.0",
-    "redux-actions": "^2.2.1"
+    "redux-actions": "^2.2.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -60,7 +60,8 @@
     "impagination": "^1.0.0-alpha.3",
     "inflected": "^2.0.3",
     "react-intl": "^2.4.0",
-    "redux-actions": "^2.2.1"
+    "redux-actions": "^2.2.1",
+    "shallowequal": "^1.1.0"
```
taras: We can remove this dependency because it's not being used anymore.

cherewaty:
Approved, but we'll need to figure out what to do about those unrelated `PackageSelection` tests before merging.

carolegodfrey:
This looks great -- but I did notice one thing when pulling down this branch and testing. I enter a search term and search from the modal, I then go back to the modal and clear the search term (x button in text box) and set some filters and again search. The old search term persists.

cherewaty:
:shipit:

PR #478, opened 2018-07-23
UIEH-477: Add null checks to avoid trying to focus search results when there are none

keathleydavidj:
## Purpose
There is currently a bug wherein selecting a search filter without having a query string in the "Search" field will cause the UI to blow up, showing an uncaught error to the user.

## Approach
It turns out that there is one place where we should be doing a null check to ensure we're not accessing properties on an object that is not there. In addition I found a spot where a boolean property was being coerced into a string somehow.

## Todos
- [x] I have an open question about the intended behavior in this particular case here:  https://issues.folio.org/browse/UIEH-477?focusedCommentId=33268&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-33268

### DIFF
--- a/src/components/search-paneset/search-paneset.js
+++ b/src/components/search-paneset/search-paneset.js
@@ -82,7 +82,7 @@ export default class SearchPaneset extends React.Component { // eslint-disable-l
     let isSameSearchType = prevProps.resultsType === this.props.resultsType;
 
     // focus the pane title when a new search happens within the same search type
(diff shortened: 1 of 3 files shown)


keathleydavidj:
Khalilah did answer the question I had about behavior. It should not try to filter anything until a search term is entered by the user.

keathleydavidj:
Good point, I'll work on that now.

PR #483, opened 2018-07-24
UIEH-476 - Package Edit Page - Partially Selected Package Handling

carolegodfrey:
[UIEH-476 - Holding status: package edit page: Partially Selected Package Handling](https://issues.folio.org/browse/UIEH-476)

resolves #484 

## Purpose

The Managed Package Edit Page currently displays a holding status area with 2 states:
- Selected
- Not Selected (with an `Add to holdings` button)

The menu area also shows `Remove from holdings` or `Add to holdings` depending upon state.

This PR enhances the Managed Package Edit Page to support partially selected packages. (ie a package that does not have all of its titles selected). Specifically, it allows the user to completely select a partially selected package.  It adds a 3rd state:

- (# selected titles)  of (# total titles) titles selected (with an `Add all to holdings` button)

Added menu item `Add all to holdings`

Additional Items Addressed:
- Corrected text which appears on package show page menu for custom package. Specifically display `Delete package` instead of `Remove from holdings`
- Added use of selection status on custom package edit (to be consistent). For custom packages, state will simply show Selected. Since once a custom package is de-selected it is deleted.
- Moved packageSelectionPending state to redux model - merged in work from https://github.com/folio-org/ui-eholdings/pull/484 to remove state maintenance out of component to simplify.

Note -- test refactoring work was done in separate PR(s) https://github.com/folio-org/ui-eholdings/pull/488, https://github.com/folio-org/ui-eholdings/pull/490, https://github.com/folio-org/ui-eholdings/pull/494. There are still some issues -- so these tests are skipped temporariliy.
~~Additionally, we need to Refactor Package Selection tests. We currently see intermittent test failures when testing 'selecting a package title to add to holdings'.  The problem occurs when checking a temporary isSelecting state (due to timing)~~

## Approach 

Update Managed Package Edit Page as follows:
- Add a Status indicator which includes number of selected titles/total titles and a Button 'Add all to holdings' when a package is partially selected.
- Updated Package Edit Page(Pane Header Dropdown) to include 'Add all to holdings' depending on state

Plan to understand the work which was done for `UIEH-442 - Holding status: package show page: Partially Selected Package Handling`  as a model for functionality that is required. Pull request is https://github.com/folio-org/ui-eholdings/pull/475. This work implemented partially selected handling for the `package show` page. This PR adds the same functionality for the `package edit` page

Additional Goals:
- Move actions to route (as much as possible)
- Strive to move ManagedPackageEdit towards a functional component (like SelectionStatus)
- Move complex state(such as isPending) into redux/package.js model object (see isPartiallySelected as example)

#### TODOS and Open Questions
- [x] Per UIEH-406 Custom Package Show/Edit,  'Delete Package' should appear as text in header dropdown. Currently shows 'Remove from holdings' on the show page. Placeholder to fix in either this PR or as a follow on

### DIFF
--- a/src/components/package/edit-custom/custom-package-edit.js
+++ b/src/components/package/edit-custom/custom-package-edit.js
@@ -22,6 +22,7 @@ import DetailsViewSection from '../../details-view-section';
 import NavigationModal from '../../navigation-modal';
 import Toaster from '../../toaster';
 import PaneHeaderButton from '../../pane-header-button';
+import SelectionStatus from '../selection-status';
 import styles from './custom-package-edit.css';
 
 class CustomPackageEdit extends Component {
@@ -32,7 +33,8 @@ class CustomPackageEdit extends Component {
     model: PropTypes.object.isRequired,
     onSubmit: PropTypes.func.isRequired,
     pristine: PropTypes.bool,
-    intl: intlShape.isRequired
+    intl: intlShape.isRequired,
+    addPackageToHoldings: PropTypes.func.isRequired
   };
 
   static contextTypes = {
@@ -220,17 +222,10 @@ class CustomPackageEdit extends Component {
                 <DetailsViewSection
                   label={intl.formatMessage({ id: 'ui-eholdings.label.holdingStatus' })}
                 >
-                  <label
-                    data-test-eholdings-package-details-selected
-                    htmlFor="custom-package-details-toggle-switch"
-                  >
-                    <h4>
-                      {packageSelected ?
-                        (<FormattedMessage id="ui-eholdings.selected" />) :
-                        (<FormattedMessage id="ui-eholdings.notSelected" />)
-                      }
-                    </h4>
-                  </label>
+                  <SelectionStatus
+                    model={model}
+                    onAddToHoldings={this.props.addPackageToHoldings}
+                  />
                 </DetailsViewSection>
                 <DetailsViewSection
                   label={intl.formatMessage({ id: 'ui-eholdings.label.packageInformation' })}
--- a/src/components/package/edit-managed/managed-package-edit.js
+++ b/src/components/package/edit-managed/managed-package-edit.js
@@ -5,7 +5,6 @@ import isEqual from 'lodash/isEqual';
 import { intlShape, injectIntl, FormattedMessage } from 'react-intl';
 
 import {
-  Button,
   Icon,
   Modal,
   ModalFooter,
@@ -20,6 +19,7 @@ import DetailsViewSection from '../../details-view-section';
 import NavigationModal from '../../navigation-modal';
 import Toaster from '../../toaster';
 import PaneHeaderButton from '../../pane-header-button';
+import SelectionStatus from '../selection-status';
 import styles from './managed-package-edit.css';
 
 class ManagedPackageEdit extends Component {
@@ -30,7 +30,8 @@ class ManagedPackageEdit extends Component {
     handleSubmit: PropTypes.func,
     onSubmit: PropTypes.func.isRequired,
     pristine: PropTypes.bool,
-    intl: intlShape.isRequired
+    intl: intlShape.isRequired,
+    addPackageToHoldings: PropTypes.func.isRequired
   };
 
   static contextTypes = {
@@ -166,7 +167,6 @@ class ManagedPackageEdit extends Component {
     } = this.context;
 
     let visibilityMessage = model.visibilityData.reason && `(${model.visibilityData.reason})`;
-    let packageSelectionPending = model.update.isPending && 'isSelected' in model.update.changedAttributes;
 
     let actionMenuItems = [
       {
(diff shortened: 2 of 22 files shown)


src/routes/package-edit.js
```diff
@@ -121,6 +121,13 @@ class PackageEditRoute extends Component {
     }
   };
 
+  addPackageToHoldings = () => {
```
cowboyd: It would be awesome if we didn't have to duplicate this. How would we share the model between the package show and edit routes? /cc @taras @cherewaty 

Seems like we should at least include a note for future folks looking to make architectural improvements.

tests/pages/package-edit.js
```diff
@@ -14,6 +15,12 @@ import { hasClassBeginningWith } from './helpers';
 import Toast from './toast';
 import Datepicker from './datepicker';
 
+@interactor class PackageSelectionStatus {
```
cowboyd: We should definitely use the same interactor as for the show test since it is the same component.

tests/pages/package-edit.js
```diff
@@ -32,8 +39,12 @@ import Datepicker from './datepicker';
 }
 
 @interactor class PackageEditDropDownMenu {
+  addToHoldings = new Interactor('.tether-element [data-test-eholdings-package-add-to-holdings-action]');
+  removeFromHoldings = new Interactor('.tether-element [data-test-eholdings-package-remove-from-holdings-action]');
   clickRemoveFromHoldings = clickable('.tether-element [data-test-eholdings-package-remove-from-holdings-action]');
   clickAddToHoldings = clickable('.tether-element [data-test-eholdings-package-add-to-holdings-action]');
+  isAddToHoldingsPresent = isPresent('.tether-element [data-test-eholdings-package-add-to-holdings-action]');
```
cowboyd: I believe we can just use `expect(interactor.addToHoldings.isVisible).to.equal(true)`

tests/pages/package-edit.js
```diff
@@ -32,8 +39,12 @@ import Datepicker from './datepicker';
 }
 
 @interactor class PackageEditDropDownMenu {
+  addToHoldings = new Interactor('.tether-element [data-test-eholdings-package-add-to-holdings-action]');
+  removeFromHoldings = new Interactor('.tether-element [data-test-eholdings-package-remove-from-holdings-action]');
   clickRemoveFromHoldings = clickable('.tether-element [data-test-eholdings-package-remove-from-holdings-action]');
```
cowboyd: We should consider removing these, maybe making a note that they're deprecated? because instead of

```js
interactor.clickRemoveFromHoldings();
```

we can say
```js
interactor.removeFromHoldings.click()
```

tests/pages/package-edit.js
carolegodfrey: I did try that --  it is cleaner and it works for the true case. But If I have the case to check that the menu item is not present `expect(PackageEditPage.dropDownMenu.addToHoldings.isVisible).to.equal(false);`
I get 
` Error: unable to find ".tether-element [data-test-eholdings-package-add-to-holdings-action]"`
So not sure how to handle that within an interactor like that

src/routes/package-edit.js
carolegodfrey: I've added a comment for now. Would appreciate any recommendations to avoid duplication @taras or @cherewaty. Don't believe I can move to a common utilities method since model and update are needed by the method

tests/pages/package-edit.js
carolegodfrey: Moved to separate file and adjusted unit tests

cowboyd:
We should probably get this merged first and rebase onto it https://github.com/folio-org/ui-eholdings/pull/484

src/components/package/edit-custom/custom-package-edit.js
```diff
@@ -220,17 +222,11 @@ class CustomPackageEdit extends Component {
...
-                    <h4>
-                      {packageSelected ?
-                        (<FormattedMessage id="ui-eholdings.selected" />) :
-                        (<FormattedMessage id="ui-eholdings.notSelected" />)
-                      }
-                    </h4>
-                  </label>
+                  <SelectionStatus
+                    model={model}
+                    isPending={false}
```
cowboyd: We shouldn't need this property anymore because it's on `model` right?

src/components/package/edit-managed/managed-package-edit.js
```diff
@@ -241,35 +245,11 @@ class ManagedPackageEdit extends Component {
...
-                        onClick={this.handleSelectionAction}
-                        data-test-eholdings-package-add-to-holdings-button
-                      >
-                        <FormattedMessage id="ui-eholdings.package.addToHoldings" />
-                      </Button>
-                    }
-                  </label>
+                  <SelectionStatus
+                    model={model}
+                    isPending={packageSelectionPending}
```
cowboyd: Same here, we should be able to get rid of this, correct since it will be ignored by the `SelectionStatus` component.

src/components/package/edit-custom/custom-package-edit.js
carolegodfrey: Thanks for catching @cowboyd -- just removed!

src/components/package/edit-managed/managed-package-edit.js
carolegodfrey: All set!

PR #485, opened 2018-07-24
UIEH-473: Localize strings in title - create, edit, show

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-473, we are supposed to localize strings in title - create, edit and show pages.

## Approach
- Move strings into ui-eholdings/translations/ui-eholdings/en.json
- Reference them using react-intl `FormattedMessage` component
- Manually test and verify on the above mentioned pages that they look ok locally
- Run acceptance tests and double check that nothing is broken

## Screenshots
title_localize


src/components/title/_fields/publication-type/publication-type.js
```diff
@@ -13,7 +14,7 @@ export default function PublicationTypeField() {
       <Field
         name="publicationType"
         component={Select}
-        label="Publication type"
+        label={intl.formatMessage({ id: 'ui-eholdings.title.publicationType' })}
         dataOptions={[
           { value: 'Audio Book', label: 'Audio Book' },
```
carolegodfrey: On Publication Type data -- depending on spike outcome for localization of filters -- we may want to translate these as well

PR #486, opened 2018-07-24
UIEH-320 - Provider Detail: List of Packages: Search Within Modal : Unable to Reset Search

SamKeathley:
## Purpose
Per [UIEH-320](https://issues.folio.org/browse/UIEH-320) we currently do not have a way to reset all filters back to default within the search modal. This change adds that function to the modal.

## Approach
With this change I am removing the feature that hides the modal footer and buttons that I put in with [UIEH-371](https://issues.folio.org/browse/UIEH-317) in favor of always displaying the buttons, but just enabling/disabling them when applicable. The `Reset all` button also uses the `stripes-components` modal button styles for consistency.

### DIFF
--- a/src/components/details-view/details-view.js
+++ b/src/components/details-view/details-view.js
@@ -60,7 +60,7 @@ export default class DetailsView extends Component {
   state = {
     isSticky: false,
     showSearchModal: false,
-    showSearchModalSearchButton: false,
+    showSearchModalSearchButton: true,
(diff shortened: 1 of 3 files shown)


SamKeathley:
Thanks, guys! I'll wait on @Robdel12 to take a look at this one before merging to make sure it doesn't conflict with his React 16.4 work.

PR #490, opened 2018-07-25
Use Mirage Response to block endpoint with response code

cherewaty:
## Purpose
The `indicates it is working to get to desired state` assertion was still being flaky after https://github.com/folio-org/ui-eholdings/pull/488.

## Approach
My best guess is that the shorthand for the mirage endpoint was going ahead and resolving its error code (500), and not being blocked.

To get around that, I used mirage's `Response` to manually build the full response, with error code and no headers.

## Next Steps
There are few more tests relying on Promises that, although they're thankfully not flaky, would benefit from `server.block()` and `server.unblock()`.


cowboyd:
This looks great, and frankly, I like the `response` format better anyhow.

PR #495, opened 2018-07-26
UIEH-474 Localize resource show and edit

elrickvm:
## Purpose
Localization allows for any text that is presented to a eHoldings user to be translated for the particular language and region the user is using. Translations are done my an outside vendor.

## Approach
 - Replace all raw strings  and import the translations using a component or function from React-Intl package that pulls them from a translations file.


elrickvm:
Its all good. Yeah i am trying to repair those merge conflicts now. Its a dozy. Git is all tripped up. So I have to resolve them manually and edit the diffs directly.

PR #496, opened 2018-07-26
[UIEH-499] Upgrade to React 16.4.x

Robdel12:
## Purpose

The React team shipped a breaking change in the `16.4.x`  release (see: https://github.com/facebook/react/issues/12898). They changed the way `getDerivedStateFromProps` works. In `16.3.x` it wasn't called when `setState` was invoked. But in the new version `getDerivedStateFromProps` gets called when props change, `setState` is called, and on `forceUpdate()`. 

This is the old React lifecycle:
image

[(source)](https://twitter.com/dan_abramov/status/981712092611989509)

This is the new React lifecycle: 
image

[(source)](https://twitter.com/dan_abramov/status/1019527421303951360)

## Approach

With the change to `getDerivedStateFromProps` described above, I went through and removed `state` from the `search-form` component so it's only using props now. This makes it much easier to manage what is going on with the `search-form` component.

Previously there were props that would come in and the state would basically extend the props. When we moved to `getDerivedStateFromProps` in `16.3.x` it was easy to just update the state when props changed because `getDerivedStateFromProps` was only called for prop changes. 

With the new update `getDerivedStateFromProps` is called on `setState` so there's no way to know if you should be using the props or the newly `setState` version. You also have no way of knowing if it's a state change or prop change without storing props separately on the state. 

## Learning

You probably don't need derived state was a helpful blog post: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html


src/components/search-form/search-form.js
```diff
@@ -32,104 +31,44 @@ export default class SearchForm extends Component {
     }),
     onSearch: PropTypes.func.isRequired,
     searchString: PropTypes.string,
-    filter: PropTypes.object,
-    searchfield: PropTypes.string,
+    searchField: PropTypes.string,
+    searchFilter: PropTypes.object, // TODO: shape it
```
keathleydavidj: When does this get shaped?

Robdel12: HA, great catch on this. I'll push an update doing that.

PR #499, opened 2018-07-27
UIEH-512: Expand/collapse all accordions

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-512, we want to implement expand all/collapse all accordions in provider/package/title detail record show pages. 

## Approach
- Implement expand all/collapse all following docs here: https://github.com/folio-org/stripes-components/tree/master/lib/Accordion
- Manually test to ensure that they are working as expected on the package-show, title-show and provider-show pages.
- Convert ProviderShow from a functional component to a class component to be able to apply expand/collapse all accordions.

#### TODOS and Open Questions
- [x] Port this code over for provider-show
- [ ] Currently, does not deal with accordions in lists on each of these pages. It is WIP for providers currently.
- [x] Write unit tests.

## Screenshots
expandcollapse


src/components/package/show/package-show.js
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-
+import _ from 'lodash';
```
taras: We should be more specific here so we don't import the entire `lodash` library if we can avoid it. If you need deepCopy, then you should be able to do `import deepCopy from 'lodash/deepCopy'`. If this doesn't work, let me know.

src/components/package/show/package-show.js
```diff
@@ -81,14 +87,31 @@ class PackageShow extends Component {
     this.setState({ isCoverageEditable });
   };
 
+  handleSectionToggle = ({ id }) => {
+    this.setState((curState) => {
```
taras: Instead of cloning the object with deep copy, we could use `lodash/fp/update`. `fp` stands for functional programming. It's unfortunately not well documented :( but it allows to do a set as an immutable operation. 

I made an example of you of how it can be used https://runkit.com/taras/5b606de43fb58c001196f7a3

You should be able to replace this code with,

```js
let next = update(`sections.${id}`, value => !value, this.state);

this.setState(next);
```

Here is the documentation for [update function](https://gist.github.com/jfmengels/6b973b69c491375117dc#_updatepath-updater-object).

src/components/package/show/package-show.js
```diff
@@ -81,14 +87,31 @@ class PackageShow extends Component {
...
+      const newState = _.cloneDeep(curState);
+      newState.sections[id] = !newState.sections[id];
+      return newState;
+    });
+  }
+
+  handleExpandAll = (sections) => {
+    this.setState((curState) => {
+      const newState = _.cloneDeep(curState);
```
taras: Here you can use `lodash/fp/set`.

```js
import set from 'lodash/fp/set';

let next = set('sections', sections, this.state);
this.setState(next);
```

You could also just use spread operator here,

```js
this.setState({
  ...this.state,
  sections
});
```

sduvvuri-ebsco: @taras: is there an advantage of doing one way versus the other? ```let next = set('sections', sections, this.state);
this.setState(next);``` vs.
```this.setState({
  ...this.state,
  sections
});```?

taras: I would say that you probably want to use the simplest way possible. In this case, the spread operator relies on something that's built into JavaScript so it's the simplest way.

sduvvuri-ebsco: @taras: also, just to ensure my understanding between the old way and using the fp way: Using the new way is advantageous because we are saving on memory by not creating an entirely new object of state and replacing in existing state with the new value using fp..is that right?

taras: In React apps, it's generally good idea to practice doing things immutably. In this case, the implementation is correct but style can be better. The patterns around working with immutable data add up once you start recognizing them, so it's a good habit to start writing immutable code rather than using immutable patterns.

sduvvuri-ebsco:
I feel that this pattern is more elegant than the plain JS way..I have updated the code to use this pattern instead of using the spread operator. Hope that should be ok..

PR #504, opened 2018-08-01
Lay out detail record data with KeyValueColumns

cherewaty:
## Purpose
To visually resemble other FOLIO modules, lay out the "____ information" section of detail records in columns.

Resolves https://issues.folio.org/browse/UIEH-206 and https://issues.folio.org/browse/UIEH-507.

## Approach
- Created `<KeyValueColumns>` that can wrap the collections of `<KeyValue>` pairs. To create a column, use a `<div>`.
- The most difficult decision on each was where to place the break:
  - Provider: none, so that "_____ selected" is always above "Total ______" like a fraction
  - Package: before "Titles selected" and "Total titles" counts
  - Title: just tried to pick a halfway point; after identifiers, before subjects
  - Resource: break is between fields inherited from the title and fields inherited from the package
- When the detail record container is less than 630px wide, there's only a single column. That breakpoint was chosen just based on looking at what felt comfortable with some existing records.
- Gutter between columns set at 15px using the `stripes-components` `gutter` variable.

## Why not `react-flexbox-grid`?
`stripes-components` exports `react-flexbox-grid` as a recommendation for FOLIO modules to use.

### It relies on the window's context instead of its container.
`react-flexbox-grid` uses media queries, so it can only be responsive to the width of the entire window. This layout is being used both on its own and inside a pane.

In this example, the title pane is only 374px wide, at which we'd want just a single column to maintain legibility. `react-flexbox-grid` can't do anything with the pane width information.
#### With `react-flexbox-grid`
43521050-825adf14-955a-11e8-8a55-78a546d45c29

#### With `KeyValueColumns`
43521052-85589684-955a-11e8-9692-eb01c9c95847

### It wraps a limited, dated API around a much more powerful abstraction.
`react-flexbox-grid` uses the idea of a flexible-width 12-column layout, which far predates having the modern flexbox and CSS grid APIs available in browsers.

`react-flexbox-grid` uses flexbox CSS under the hood, but removes much of its power (similar to the problem with `stripes-components` `<Pane>`s). The API for `react-flexbox-grid` receives a number 1-12 to indicate a width as a percentage of 12. That means `react-flexbox-grid` really only provides a way to set `flex-basis` to a percentage.

But flexbox can do so much more! Specifically, `<KeyValueColumns>` uses:
- `flex-flow: row wrap`: tells the container to wrap the columns to be a single column if they don't fit horizontally
- `flex: 1 0 20rem`: sets `flex-grow` to 1, meaning the child can grow as wide as it needs to fill; sets `flex-shrink` to 0, meaning "don't shrink below the `flex-basis`"; sets `flex-basis` (the default size) to `20rem`

## Learning
[A Complete Guide to Flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)
[Don't Use My Grid System (or any others) by Miriam Suzanne](https://youtu.be/mDRfFEcj3-Q)

### Package
package

### Title
title
another title

### Resource
resource

### DIFF
--- a/src/components/key-value-columns/index.js
+++ b/src/components/key-value-columns/index.js
@@ -1,0 +1,1 @@
+export { default } from './key-value-columns';
--- a/src/components/key-value-columns/key-value-columns.css
+++ b/src/components/key-value-columns/key-value-columns.css
@@ -1,0 +1,15 @@
+@import '@folio/stripes-components/lib/variables';
+
+.key-value-columns {
+  display: flex;
+  flex-flow: row wrap;
+
+  & > div {
+    flex: 1 0 20rem;
+    margin-right: var(--gutter);
+
+    [dir="rtl"] {
+      margin-left: var(--gutter);
+    }
+  }
+}
--- a/src/components/key-value-columns/key-value-columns.js
+++ b/src/components/key-value-columns/key-value-columns.js
@@ -1,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './key-value-columns.css';
+
+export default function KeyValueColumns({ children }) {
+  return (
+    <div
+      className={styles['key-value-columns']}
+    >
+      {children}
+    </div>
+  );
+}
+
+KeyValueColumns.propTypes = {
+  children: PropTypes.node
+};
--- a/src/components/package/show/package-show.js
+++ b/src/components/package/show/package-show.js
@@ -21,6 +21,7 @@ import NavigationModal from '../../navigation-modal';
 import Toaster from '../../toaster';
 
 import SelectionStatus from '../selection-status';
+import KeyValueColumns from '../../key-value-columns';
 import styles from './package-show.css';
 
 class PackageShow extends Component {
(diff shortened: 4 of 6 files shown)


JohnC-80:
I like this - and I am immediately curious if there are any other 'grid' components out there that do expose more of the power of flexbox. Dated as the columns are, it's a familiar convention that does lend more control to the width based on media queries - of course, the weakness of that is highlighted here. @filipjakobsen @rasmuswoelk I'd really like to expand this as necessary to become a 'better' grid - it's quite simple in its current implementation... I'm immediately curious of any gotchas that might exist here if we do adopt something similar.

cherewaty:
Merging; @JohnC-80 I'll leave more comments and follow up with y'all when I'm back from vacation on Tuesday.

cherewaty:
Thinking on this some more:
- I think `<KeyValueColumns>` could very easily be moved to `stripes-components` so other modules have that available as an alternative to `react-flexbox-grid`.
- Forms using `react-flexbox-grid` probably shouldn't - vertically stacked forms (one form element per row) perform much better in usability studies than ones that attempt to lay out inputs in columns.

PR #512, opened 2018-08-07
UIEH-479: Localize Catch All - Details View and Accordion List

SamKeathley:
## Purpose
Continue work on [UIEH-479](https://issues.folio.org/browse/UIEH-479)

## Approach
This is a small one because I am pushing a potential fix to the `Search \u0026 Filters` bug to hopefully render an ampersand after building.

#### Screenshot of bug
screen shot 2018-08-07 at 11 59 49 am

### DIFF
--- a/translations/ui-eholdings/en.json
+++ b/translations/ui-eholdings/en.json
@@ -133,6 +133,9 @@
     "label.sortOptions": "Sort options",
     "label.selectionStatus": "Selection status",
     "label.search": "Search",
+    "label.accordionList": "Records found",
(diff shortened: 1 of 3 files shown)


translations/ui-eholdings/en.json
```diff
@@ -133,6 +133,9 @@
     "label.sortOptions": "Sort options",
     "label.selectionStatus": "Selection status",
     "label.search": "Search",
+    "label.accordionList": "Records Found",
```
cherewaty: Should probably make this sentence case: `Records found`

PR #516, opened 2018-08-09
UIEH-165 Provider Detail: Allow user to choose proxy

carolegodfrey:
## Purpose

Per https://issues.folio.org/browse/UIEH-165, users should have the ability to view and set proxy information at the Provider Level. If proxy is set at Provider level - then all packages/titles tied to provider will inherit this setting.

## Approach

Confirm Proxy information can be retrieved and set through mod-kb-ebsco 

**Providers endpoint**

mod-kb-ebsco currently supports a proxy object as part of attributes returned from providers GET request
`{"data":{"id":"38","type":"providers","attributes":{"name":"Taylor and Francis","packagesTotal":710,"packagesSelected":6,"providerToken":null,"supportsCustomPackages":false,"proxy":{"id":"TestingFolio","inherited":true}}...}`

Providers PUT request supports setting proxy id as part of the request body
`{
  "data": {
    "type": "providers",
    "attributes": {
      "providerToken": {
        "value": null
      },
      "proxy": {
        "id": "<n>"
      }
    }
  }
}`

GET eholdings/root-proxy
`{
    "data": {
        "id": "root-proxy",
        "type": "rootProxies",
        "attributes": {
            "id": "root-proxy",
            "proxyTypeId": "Proxy-ID-123"
        }
    },
    "jsonapi": {
        "version": "1.0"
    }
}`

GET eholdings/proxy-types
`{
    "data": [
        {
            "id": "<n>",
            "type": "proxyTypes",
            "attributes": {
                "id": "<n>",
                "name": "None",
                "urlMask": ""
            }
        },
        {
            "id": "Proxy-ID-123",
            "type": "proxyTypes",
            "attributes": {
                "id": "Proxy-ID-123",
                "name": "mOHAN-Proxy-name-123",
                "urlMask": "http://ezproxy.myinstitute.edu/login?url=www.ebsco.com"
            }
        },
        {
            "id": "Proxy-ID-234",
            "type": "proxyTypes",
            "attributes": {
                "id": "Proxy-ID-234",
                "name": "mOHAN-Proxy-name-234",
                "urlMask": "http://ezproxy.myinstitute.edu/login?url=www.ebsco.com"
            }
        }
    ],
    "jsonapi": {
        "version": "1.0"
    }
}`

Update Provider Show Page as follows:
- Add a `Provider settings` section
- Apply an accordion to the Provider settings section
- Add support for edit (via pane header dropdown or pencil icon)
- Display proxy -- prePend proxy name with `Inherited` if proxy has been set a higher level (Root Proxy Setting)
- inherited indication is returned as part of GET Providers response
- needed to GET Proxy-Types on show page to properly look up name from ID

Add Provider Edit Page as follows:
- Added provider-edit route
- Create provider-edit component with a`Provider settings` section
- show dropdown with available selections or None and disabled (if there are no selections available)
- Provider Edit Route needs to GetProvider, GetProxyTypes (to populate select dropdown) and GetRootProxy (to know where to place inherited in select dropdown)

Reorganized folder structure for Provider component to include both edit and show components

Unit tests

#### TODOS and Open Questions
- While testing, discovered that a Provider cannot be updated if there are no selected packages for  the provider. Attempts to update will result in error message - No Packages Selected. This should be handled more gracefully by disabling edit or proxy drop down


bigtest/interactors/provider-edit.js
```diff
@@ -0,0 +1,53 @@
...
+  name = text('[data-test-eholdings-details-view-name="provider"]');
+  nameHasFocus = is('[data-test-eholdings-details-view-name="provider"]', ':focus');
+  clickCancel= action(function () {
+    return this
+      .dropDown.clickDropDownButton()
+      .dropDownMenu.cancel.click();
+  });
+  clickSave = clickable('[data-test-eholdings-provider-save-button]');
+  isSaveDisabled = property('[data-test-eholdings-provider-save-button]', 'disabled');
+  hasErrors = isPresent('[data-test-eholdings-details-view-error="package"]');
```
sduvvuri-ebsco: should this be `="provider"` instead of package?

carolegodfrey: Yes -- thanks good catch

bigtest/tests/provider-edit-test.js
```diff
@@ -0,0 +1,99 @@
...
+import { describeApplication } from '../helpers/describe-application';
+import ProviderShowPage from '../interactors/provider-show';
+import ProviderEditPage from '../interactors/provider-edit';
+
+describeApplication('ProviderEdit', () => {
+  let provider,
+    packages;
+
+  beforeEach(function () {
+    provider = this.server.create('provider', 'withPackagesAndTitles', 'withProxy', {
```
sduvvuri-ebsco: Not sure if this param being passed is serving any specific purpose - but should `withPackagesAndTitles` be renamed to `withPackages` or is it legit to say `withPackagesAndTitles`?

src/components/provider/_fields/proxy-select/proxy-select-field.js
```diff
@@ -0,0 +1,52 @@
...
+      data-test-eholdings-provider-proxy-select-field
+      className={styles['proxy-select-field']}
+    >
+      <Field
+        id="eholdings-providers-proxy-id"
+        name="proxyId"
+        component={Select}
+        dataOptions={options}
+        label={intl.formatMessage({ id: 'ui-eholdings.provider.proxy' })}
+        disabled={options.length < 2}
```
sduvvuri-ebsco: If we disable the dropdown, would the user be necessarily aware as to why it is disabled? Not sure if this is an obvious user experience? We should get feedback from others regarding this.

bigtest/tests/provider-edit-test.js
```diff
@@ -0,0 +1,99 @@
...
+          expect(ProviderShowPage.$root).to.exist;
+        });
+
+        it('shows newly saved proxy', () => {
+          expect(ProviderShowPage.proxy).to.equal('bigTestJS');
+        });
+
+
+        it('shows a success toast message', () => {
+          expect(ProviderShowPage.toast.successText).to.equal('Provider saved.');
```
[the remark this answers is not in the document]
carolegodfrey: I followed the same pattern as is done elsewhere (resources, packages) -- the success toast for  all of these are done on the show page after the page transitions from edit to show.

bigtest/tests/provider-edit-test.js
carolegodfrey: It is legit to say  `withPackagesAndTitles`  in fact there is not a trait on Providers for just `withPackages`

bigtest/interactors/provider-edit.js
carolegodfrey: Updated and also added unit tests for Provider PUT and GET server errors

bigtest/tests/provider-edit-test.js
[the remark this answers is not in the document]
sduvvuri-ebsco: sounds good 👍

bigtest/tests/provider-edit-test.js
sduvvuri-ebsco: ok 👍

src/components/provider/_fields/proxy-select/proxy-select-field.js
carolegodfrey: I'll follow up with Khalilah -- we did discuss this and the case where all packages for the provider are not selected. In both of these cases the user should not be able to change the proxy selection

elrickvm: Maybe there could be some messaging around the disabled drop-down. Saying something like "No proxies available"

carolegodfrey:
Thanks for review @elrickvm -- I did confirm this dropdown disable behavior with Khalilah for this PR  - noting comments on this PR -- so for now I will keep as is --

PR #517, opened 2018-08-09
UIEH-509 Create routing architeture documentation for ui-eholdings

elrickvm:
## Purpose
The URL is a powerful and vital part of Web applications. When a user navigates or pastes a URL into the browser that URL is used by the Application to construct the page or location where the user is. This documentation is set to serve as an explanation and illustration of how this construction of a page happens and how it all stems from the URL.

## Approach
 - The overall document will give the reader:
    -  a high-level view of how data flows between the Route Component and the Views components.
    - what role the URL plays in Routing, Routes, a View Components.
    - what roles the Routes and Views play in regards to data
    - a general idea of how the all these Components are used together

## TODO
 - Possibly create a visual diagram that illustrates how a request/response cycle flows through the application and what Components and parts of the internal application take part.


elrickvm:
@sduvvuri-ebsco diagram finished in time for PR and added. :-)

sduvvuri-ebsco:
Perfect...looking pretty good..thanks @elrickvm 👍

PR #520, opened 2018-08-10
[UIEH-214] Implement filtering for package resources

Robdel12:
## Purpose

Allow the ability for you to filter packages nested resources (which are titles). 

## Approach

I started with implementing the search modal inside of the `package-show` detail view. A lot of the work was getting the mirage config setup to where we can filter & search nested package resources. Once that work was done I was able to whip up some tests to make sure filtering titles worked properly. 

One thing I noticed while working through this is we could leverage the composability of interactors. We still have a lot of _page objects_. It would be nice to start to refactor our interactors to map to their component. This would allow us to compose interactors the same way we can compose the components. 

For example, the details view component could have it's own interactor. Then we could refactor our `show` pages to consume the `details-view` interactor. In this PR I had to copy over some interactors that were defined on a specific `show` page but they all are rendered by the `details-view` component.

### DIFF
--- a/src/components/search-modal/search-modal.js
+++ b/src/components/search-modal/search-modal.js
@@ -175,7 +175,7 @@ class SearchModal extends React.PureComponent {
               searchFilter={query.filter}
               searchField={query.searchfield}
               sort={query.sort}
-              onSearch={this.handleListSearch}
+              onSearch={this.updateSearch}
               displaySearchTypeSwitcher={false}
               displaySearchButton={false}
               onFilterChange={this.handleFilterChange}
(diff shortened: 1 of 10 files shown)


src/components/search-modal/search-modal.js
```diff
@@ -85,6 +85,15 @@ class SearchModal extends React.PureComponent {
     });
   }
 
+  handleSerchFieldChange = (searchfield) => {
```
cherewaty: `handleSerchFieldChange` -> `handleSearchFieldChange`

cherewaty:
When I open up the search modal, type a string in the query field, then hit Enter, the modal just closes and clears my search. I would expect the search modal form to submit the text I typed instead. Maybe "Reset all" is being used as the form's submit action? Unsure.

Robdel12:
@cherewaty Eesh, yea, that exists for all search modals. I'll see if it's a quick fix.

PR #523, opened 2018-08-15
Minor search form modal fix

rasmuswoelk:
This is a very small fix to make sure the newly added Modal transition can run correctly. 

Because of the condition wrapping around the `<Modal>` it would only transition when entering. Now it works both when the modal enters and leaves the DOM.

Also fixed a tiny linting error.


cherewaty:
Rebased to give Jenkins a kick in the pants.

PR #525, opened 2018-08-15
UIEH-217 - allow to close search pane and show filter count

taras:
## Purpose

This PR makes the search pane collapsible by pressing on the search button. The search button also show a count of active filters when the search pane is closed.

Story [UIEH-217](https://issues.folio.org/secure/RapidBoard.jspa?rapidView=40&projectKey=UIEH&view=detail&selectedIssue=UIEH-217)

## Approach

- [x] Show search button all the time
- [x] Search button open and closes the pane
- [x] Show the badge with number of filters
- [x] Only show badge when search filters pane is closed
- [x] Apply button only shows in overlay
- [x] At smaller only show overlay when details are closed

*Note*: I removed the mechanism that automatically closed the search pane when search query changed. It didn't make sense anymore in the context of having a button that toggles the search pane.

#### TODOS

- [x] Add a test that verifies that the search button if visible at all times
- [x] Add a test that clicking on the search button, opens and closes the search pane
- [x] Add a test that shows that filter count badge only shows when search pane is closed and there are filters
- [ ] ~Apply button only shows when there is an overlay~


bigtest/interactors/search-badge.js
```diff
@@ -0,0 +1,12 @@
...
+  clickable,
+  isPresent,
+  text
+} from '@bigtest/interactor';
+
+export default @interactor class SearchBadge {
+  clickIcon = clickable('[data-test-eholdings-search-filters="icon"]')
+  filterIsPresent = isPresent('[data-test-eholdings-search-filters="badge"]')
+  filterText = text('[data-test-eholdings-search-filters="badge"]')
+}
```
Robdel12: 😍 composable interactors

src/components/search-modal/search-badge.js
```diff
@@ -0,0 +1,31 @@
...
+      {filterCount > 0 && (
+        <span data-test-eholdings-search-filters="badge">
+          <Badge className={styles['filter-count']}>{filterCount}</Badge>
+        </span>
+      )}
+      <IconButton
+        icon="search"
+        onClick={onClick}
+        data-test-eholdings-search-filters="icon"
+      />
```
cherewaty: In this situation...
- Apply a filter
- Click the search icon to collapse the leftmost pane
- There's now a number of filters count next to the icon
.. .the search icon has a CSS problem where the inner part of the header overlaps it, making very little of the icon clickable (discoverable by hovering over the icon).

Looks like we're supposed to use the `badgeCount` prop on `<IconButton>` instead: https://github.com/folio-org/stripes-smart-components/blob/0565ade001de925a0df27a0be31d1a71b3070451/lib/SearchAndSort/SearchAndSort.js#L512-L517

carolegodfrey:
Looks really nice @taras -- just have a couple of comments/questions:
http://localhost:3000/eholdings?searchType=providers&q=taylor&sort=name - if I search for a provider and then apply a sort -- I see a filter count =2. I don't believe the sort is supposed to contribute to the filter count (at least that's my understanding)
On the unit tests, was wondering if there should be similar tests added for package and title (as was added for provider)? possibly redundant but filters are different for each of the search types

taras:
> if I search for a provider and then apply a sort -- I see a filter count =2. I don't believe the sort is supposed to contribute to the filter count (at least that's my understanding)

This is how the search modal badge works. It actually works by making any query value non default. When this happens, it counts the filter. I think it makes sense because it shows how many adjustable values lead to the current results.

> On the unit tests, was wondering if there should be similar tests added for package and title (as was added for provider)? possibly redundant but filters are different for each of the search types

I didn't implement this because it's just a lot of duplication. I confirmed with @cherewaty that this is ok.

PR #530, opened 2018-08-23
Fix search paneset vignette behavior

cherewaty:
## Purpose
Broken in https://github.com/folio-org/ui-eholdings/pull/525

## Approach
The layout needs two different `<SearchPaneVignette>`s: one behind the search filters pane and one behind the detail record preview pane. At some sizes they can simultaneously appear.

#### Open Questions
The `hideFilters` refactor lifted the state up for that a level... but there's one use case it didn't account for:
- When I'm on a mobile-sized screen and I click "Search", I want to be taken to the results. BUT on a larger screen, that exact same interaction shouldn't hide the filters pane. It seems like we need one more boolean in the state to deal with that.

### DIFF
--- a/src/components/preview-pane/preview-pane.css
+++ b/src/components/preview-pane/preview-pane.css
@@ -10,13 +10,19 @@
   position: absolute;
   right: 0;
   top: 0;
-  z-index: 11;
+  z-index: 12;
   background-color: white;
 }
 
 @media (--mediumUp) {
   .preview-pane {
     width: 600px;
+
+    /* preview pane is open, but not alone */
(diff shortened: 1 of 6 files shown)


cherewaty:
@taras too easy of a mistake to make, since we don't have good test coverage around these patterns that rely on slightly different stacking/behavior based on screen size. I've done testing with Selenium based on screen size, but if you have some better ideas on how to test that, they're very welcome 😁

taras:
Is this something that we could address if we run tests in an iframe? Then we can specify parameters to the iframe that we can configure with an interactor for the test harness?

PR #531, opened 2018-08-23
UIEH-526 Create coverage display section on resource forms

cherewaty:
## Purpose
Part of https://issues.folio.org/browse/UIEH-526

## Approach
Bulks up the `CoverageStatementFields` component with radio buttons to indicate whether the end patron will see dates or a custom coverage statement.

#### TODOS and Open Questions
Resources belonging to a custom title don't need to show the managed coverage dates option - that should be the final PR needed to wrap up UIEH-526.

## Learning
Gotcha I ran into: when dealing with radio buttons and Redux Form `Field`s, `type="radio"` on the `<Field>` manages the input's `checked` for you.

### DIFF
--- a/src/components/resource/_fields/coverage-statement/coverage-statement-fields.js
+++ b/src/components/resource/_fields/coverage-statement/coverage-statement-fields.js
@@ -64,7 +64,9 @@ export function validate(values, { intl }) {
   }
 
   if (values.hasCoverageStatement === 'yes' && values.coverageStatement.length === 0) {
-    errors.coverageStatement = 'If selected, coverage statement cannot be blank.';
+    errors.coverageStatement = intl.formatMessage({
+      id: 'validate.errors.coverageStatement.blank'
+    });
   }
 
   return errors;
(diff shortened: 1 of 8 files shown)


src/components/resource/_fields/coverage-statement/coverage-statement-fields.js
```diff
@@ -34,5 +63,9 @@ export function validate(values, { intl }) {
     });
   }
 
+  if (values.hasCoverageStatement === 'yes' && values.coverageStatement.length === 0) {
+    errors.coverageStatement = 'If selected, coverage statement cannot be blank.';
```
SamKeathley: This wording feels a little weird to me. We use `selected` in different ways and, if this gets read through a screenreader, that might be confusing. I don't have many suggestions on how to reword this, but maybe changing to, 
"Coverage statement field cannot be left blank if selected." might be clearer?

cherewaty: Changed to "Statement field cannot be blank if selected."

PR #532, opened 2018-08-27
UIEH-166: Support user selection of proxy on editing a package

keathleydavidj:
## Purpose
To resolve: [UIEH-166](https://issues.folio.org/browse/UIEH-166)
A user should be able to set their proxy at the package level so that all titles in that package inherit the same proxy. We've decided to support this functionality by introducing a simple `<Select />` element labelled `Proxy` to the package edit view.

## Approach
- [x] Fetch relevant data in the route (Proxy types, Provider proxy)
- [x] Display current value of a packages' proxy on the show view
- [x] Add the select box for assigning proxy value to the edit view
- [x] Support inheritance of proxies from provider level and prepend `inherited` appropriately in both package-show and package-edit
- [x] Associated tests

## Screenshots
package_proxy


src/components/package/edit-managed/managed-package-edit.js
```diff
@@ -311,6 +314,10 @@ class ManagedPackageEdit extends Component {
                           </label>
                         )}
                       </div>
+                      <ProxySelectField
+                        proxy={model.proxy}
+                        proxyTypes={proxyTypes}
+                      />
```
sduvvuri-ebsco: Around this, you probably need to add a conditional like this
```{(!proxyTypes.request.isResolved) ? (
                      ) : (
                          <ProxySelectField proxy={model.proxy} proxyTypes={proxyTypes} />
                      )}

Looks like there is delay in the model resolving when your dropdown is showing empty list

File: src/components/provider/_fields/proxy-select/proxy-select-field.js [hunk 2]
```diff
@@ -36,7 +36,7 @@ function ProxySelectField({ proxyTypes, rootProxy, intl }) {
         name="proxyId"
         component={Select}
         dataOptions={options}
-        label={intl.formatMessage({ id: 'ui-eholdings.provider.proxy' })}
+        label={intl.formatMessage({ id: 'ui-eholdings.proxy' })}
```
SamKeathley: Thanks for fixing this for me!

src/routes/package-edit.js
```diff
@@ -117,6 +128,11 @@ class PackageEditRoute extends Component {
         model.contentType = values.contentType;
       }
 
+      if ('proxyId' in values) {
+        model.proxy.id = values.proxyId;
+        model.proxy.inherited = false;
+      }
+
```
carolegodfrey: Really minor thing in packageEditSubmitted function -- may want to reset `model.proxy = {}` when package is being de-selected

sduvvuri-ebsco: Addressed 👍

PR #546, opened 2018-09-05
Move BigTest directory

cherewaty:
Seems unnecessary in isolation, but sets up nicely for having multiple kinds of test suites (like, for example, Pact). Matches the UI app blueprint provided by Stripes CLI.

Related to https://github.com/folio-org/ui-checkin/pull/61 and https://github.com/folio-org/ui-checkin/pull/62


cherewaty:
That rebase turned out to not be too bad 😂

sduvvuri-ebsco:
awesome 👍

PR #551, opened 2018-09-06
UIEH-167 - Allow User to Choose Proxy at Resource Level

carolegodfrey:
## Purpose
Per https://issues.folio.org/browse/UIEH-167 - users should be able to select a proxy at the Resource level

## Approach
- Update Resource Edit route to get `ProxyTypes` to populate the list of proxies on edit page Select.
- Pass `proxyId` in initialValues to resource edit pages to select the current proxy value associated with the resource (model.proxy.id)
- Move `proxy-select-field` (to be used by Provider, Package and Resource) up from Provider folder and adjust so that it can be used in multiple places
- Update managed and custom resource edit pages to include Proxy Select element
- Use resource's package proxy id to determine which select element should be tagged 
as `Inherited` This is `model.package.proxy.id` and is available in model since the resource is retrieved in edit route including both associated package and title Resource.find(id, { include: ['package', 'title'] })
- added associated unit tests and update to mirage configurations
- added in package level proxy at redux level since using to check for inherited
- updated ProxyDisplay commonly used function to accept inheritedProxyId as parameter. We are looking up proxy id of parent since inherited attribute is not always reliable (specifically observed at package level)
- updated interactors to use `selectable` when choosing proxies -- tests reflect displayed text as opposed to value.

#### TODOS and Open Questions
- [ x] Updates for Custom Resource Edit and associated tests
- [ x] Confirm translations are all correctly handled
- [ x ] Coordinate on work done in https://github.com/folio-org/ui-eholdings/pull/532 -- Note - updated common proxy-select-field.js to use similar logic as introduced in this PR -- only check Inherited value from parent proxy since believe it is redundant to check both parent and proxy inherited setting at current level. Proxy inherited setting at package level is not reliable -- for consistency we use parent proxy to determine inherited


bigtest/network/factories/package.js
```diff
@@ -74,6 +74,27 @@ export default Factory.extend({
...
+
+  withInheritedProxy: trait({
+    afterCreate(resource, server) {
+      let proxy = server.create('proxy', {
+        inherited: true,
+        id: 'bigTestJS'
+      });
+      resource.update('proxy', proxy.toJSON());
+      resource.save();
+    }
```
sduvvuri-ebsco: Do we really need both of these traits, specially since they are the same or can we get rid of one?

carolegodfrey: Adjusting the withProxy to reflect inherited = false and id = microstates so they are distinct

sduvvuri-ebsco: is there a specific reason why you chose to use the work `resource` carole? The reason I am asking is that in David's PR, he is using pkg - since its for package - if we change this, we could reduce number of conflicts - please let me know what you think!

bigtest/network/factories/package.js
```diff
@@ -86,5 +107,13 @@ export default Factory.extend({
...
       packageObj.update('visibilityData', visibilityData.toJSON());
     }
+    if (!packageObj.proxy) {
+      let proxy = server.create('proxy', {
+        inherited: false,
+        id: 'bigTestJS'
+      });
+      packageObj.update('proxy', proxy.toJSON());
+      packageObj.save();
+    }
```
sduvvuri-ebsco: There is a naming conflict here too, but your is better. So, we'll leave it this way and I'll adjust mine after you merge!

bigtest/network/factories/package.js
carolegodfrey: Updated to use `packageObj` so it is consistent with others -- good catch

src/components/resource/edit-custom-title/resource-edit-custom-title.js
```diff
@@ -244,6 +251,15 @@ class ResourceEditCustomTitle extends Component {
                   {resourceSelected ? (
                     <Fragment>
                       <VisibilityField disabled={visibilityMessage} />
+                      <div>
+                        {(!hasInheritedProxy || !proxyTypes.request.isResolved) ? (
```
sduvvuri-ebsco: here, if a resource does not have inherited proxy, then we probably do not want to keep spinning. I am not sure ..is it possible that a resource can have proxy but the package containing it might not have one? If its possible, then we have to modify this check, otherwise, I think we can leave it as is.

src/components/resource/edit-managed-title/resource-edit-managed-title.js
```diff
@@ -268,6 +275,15 @@ class ResourceEditManagedTitle extends Component {
                 {managedResourceSelected && (
                   <DetailsViewSection label={<FormattedMessage id="ui-eholdings.resource.resourceSettings" />}>
                     <VisibilityField disabled={visibilityMessage} />
+                    <div>
+                      {(!hasInheritedProxy || !proxyTypes.request.isResolved) ? (
```
sduvvuri-ebsco: same as above here

src/components/resource/show.js
```diff
@@ -328,12 +331,13 @@ class ResourceShow extends Component {
                   </div>
                 </KeyValue>
 
-                {(proxyTypes.isLoading || model.isLoading) ? (
+                {(!hasInheritedProxy || !proxyTypes.request.isResolved || model.isLoading) ? (
```
sduvvuri-ebsco: same as above here. probably, there is no situation where a resource can have proxy but not a package containing it but we need to double check on that assumption

PR #557, opened 2018-09-11
UIEH-556 Remove duplicate search call to server

cherewaty:
## Purpose
Resolves https://issues.folio.org/browse/UIEH-556

## Approach
The search experience only needed to lean on the URL changing and didn't explicitly have to call `this.search()` to fire off the server event.

### DIFF
--- a/src/routes/search.js
+++ b/src/routes/search.js
@@ -228,7 +228,6 @@ class SearchRoute extends Component {
     };
 
     this.updateURLParams(params);
-    this.search(params);
   };
 
   /**


sduvvuri-ebsco:
@cherewaty: I looked at the code and did the same thing you did to not have 2 calls to server but the part I am confused is -> I do not understand how `this.updateURLParams(params);` is conducting search 🤔 ...I do not see anything in that function conducting a search..Could you please help me understand?

sduvvuri-ebsco:
Is `history.replace` or `history.push` is triggering another search?

cherewaty:
The flow is this:
- `this.updateURLParams(params)` updates the URL
- because the URL changed, the `SearchRoute`'s `location` prop changed
- the `<ProviderSearchList>` realizes it needs to re-render, because the `location` has changed
- the `<ProviderSearchList>` calls its `fetch()` prop, which is the `SearchRoute`'s `fetchPage()`
- `fetchPage()` calls `this.search()`

You're right that `history` events will trigger searches. It's the change to the `SearchRoute`'s `location` that starts to communicate that change down to the `impagination`-ed `<ProviderSearchList>`

sduvvuri-ebsco:
makes sense, thanks @cherewaty !

PR #567, opened 2018-09-24
UIEH-546: Update number records found when results have been limited to first 10K

carolegodfrey:
## Purpose

Per https://issues.folio.org/browse/UIEH-546, there is currently a discrepancy in the 
total number of titles displayed in package detail information and number of records found in the title result list accordion. This specifically happens for a package when the number of records (ie titles) found is greater than 10,000. This is currently a limitation in RM API as only the first 10,000 title results are returned. In these cases, UI should display `Over 10,000` as the number of records found to indicate this fact.

## Approach
Compare the number of titles (as returned from package details request meta data ) to the number of results (as returned from request to get title list for a package). If package meta data is greater than number of results returned from list, add `Over` to the accordion header display.

#### TODOS and Open Questions
- [ ] Create unit tests -- how to simulate large result list
- [ ] AccordionListHeader is used in multiple places (as part of DetailsView). Understand impact of adding a new parameter to determine of Over gets displayed.
- [ ] Should over apply to only titles list and if the count is 10,000?

## Learning

Sample package details request to mod-kb-ebsco 

`GET packages/19-5207`

Note - response includes titleCount  attribute:

"titleCount":1552561

Sample request to GET package associated resources

`GET packages/19-5207/resources?page=1`

Note - response includes totalResults  meta data:

"totalResults":10000

## Screenshots
Before:
screen shot 2018-09-23 at 8 59 56 pm
After:
screen shot 2018-09-23 at 9 02 59 pm

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,6 +25,7 @@ module.exports = {
     "no-console": "warn",
     "prefer-const": "off",
     "react/forbid-prop-types": "off",
-    "react/destructuring-assignment": "off"
+    "react/destructuring-assignment": "off",
+    "react/sort-prop-types": ["error"]
   }
 };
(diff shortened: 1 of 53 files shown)


cherewaty:
Oops, I was too hasty in my review. I'll look again when you've got tests.

carolegodfrey:
Creating a new branch instead

PR #569, opened 2018-09-24
Lift routing actions into routes

cherewaty:
## Purpose
Only do routing mutations in route components.

## Approach
Whenever a component tried to do `history.push()`, `history.goBack()`, etc, instead fire an action up to the parent route. The parent route will handle the `history` event. Data down, actions up.

### Benefits
- We no directly longer use any old React context API (`react-intl` and `react-router` still do under the hood).
- Got rid of the final deprecated `componentWillReceiveProps`.
- Only routes need to know about `history`, `location`, and `match`. Since they're direct children of the router, they don't even have to receive those through context - they're props on routes. I used `react-router-prop-types` for those three props.
- Routing changes are more isolated, for future router upgrades or replacements.

There are two exceptions: `<DetailsView>` and `<NavigationModal>`. Both of those are very much router-aware components. I used `react-router`'s `withRouter()` higher-order component with those instead of `context.router`.

## Learning
Builds on the data flow standards communicated in https://github.com/folio-org/ui-eholdings/blob/master/docs/routing-architecture/routing-architecture.md

## Next Steps
The toaster notification pattern still doesn't feel quite right. That could use another iteration.

### DIFF
--- a/src/components/details-view/details-view.js
+++ b/src/components/details-view/details-view.js
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
-import { intlShape, injectIntl } from 'react-intl';
 import PropTypes from 'prop-types';
+import ReactRouterPropTypes from 'react-router-prop-types';
+import { withRouter } from 'react-router';
+import { intlShape, injectIntl } from 'react-intl';
 import classNames from 'classnames/bind';
 import capitalize from 'lodash/capitalize';
 import { ExpandAllButton } from '@folio/stripes-components/lib/Accordion';
@@ -35,10 +37,12 @@ class DetailsView extends Component {
     actionMenuItems: PropTypes.array,
     bodyContent: PropTypes.node.isRequired,
     handleExpandAll: PropTypes.func,
+    history: ReactRouterPropTypes.history.isRequired,
     intl: intlShape.isRequired,
     lastMenu: PropTypes.node,
     listSectionId: PropTypes.string,
     listType: PropTypes.node,
+    location: ReactRouterPropTypes.location.isRequired,
     model: PropTypes.shape({
       isLoaded: PropTypes.bool.isRequired,
       isLoading: PropTypes.bool.isRequired,
@@ -61,11 +65,6 @@ class DetailsView extends Component {
     searchModal: PropTypes.node,
     sections: PropTypes.object,
     type: PropTypes.string.isRequired
-  };
-
-  static contextTypes = {
-    router: PropTypes.object,
-    queryParams: PropTypes.object
   };
 
   static defaultProps = {
@@ -202,10 +201,10 @@ class DetailsView extends Component {
       handleExpandAll,
       listSectionId,
       onListToggle,
-      intl
+      intl,
+      history,
+      location
     } = this.props;
-
-    let { router, queryParams } = this.context;
 
     let { isSticky } = this.state;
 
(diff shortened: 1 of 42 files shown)


src/components/details-view/details-view.js
```diff
@@ -202,36 +201,36 @@ class DetailsView extends Component {
...
-    let historyState = router.history.location.state;
+    let historyState = history.location.state;
 
     let isListAccordionOpen = sections && sections[listSectionId];
 
     return (
       <div data-test-eholdings-details-view={type}>
         <PaneHeader
-          firstMenu={queryParams.searchType ? (
+          firstMenu={location.search ? (
```
wwilsman: This is true for any `?` in the URL while the previous check was only true when you were on a search route

src/components/resource/edit-managed-title/resource-edit-managed-title.js
```diff
@@ -33,60 +32,36 @@ class ResourceEditManagedTitle extends Component {
...
-        managedResourceSelected: nextProps.initialValues.isSelected,
-      });
-    }
-
+  static getDerivedStateFromProps(nextProps, prevState) {
     if (nextProps.model.update.errors.length) {
-      this.setState({
-        showSelectionModal: false
-      });
+      return { showSelectionModal: false };
```
wwilsman: I'm concerned that there might be an issue with the `initialValue` logic below never being encountered in this case. Although I suppose if there are errors the `initialValue` stuff doesn't matter?

src/components/resource/edit-managed-title/resource-edit-managed-title.js
```diff
@@ -33,60 +32,36 @@ class ResourceEditManagedTitle extends Component {
...
-    if (wasUnSelected || isCurrentlySelected) {
-      if (wasPending && needsUpdate) {
-        this.context.router.history.push(
-          `/eholdings/resources/${this.props.model.id}`,
-          { eholdings: true, isFreshlySaved: true }
-        );
-      }
+    if (nextProps.initialValues.isSelected !== prevState.initialValues.isSelected) {
+      return {
+        ...prevState,
```
wwilsman: The return from this method _updates_ state, so we don't need to include (or return below) the previous state.

src/components/package/create/package-create.js
```diff
@@ -80,11 +64,11 @@ class PackageCreate extends Component {
           <PaneHeader
             paneTitle={<FormattedMessage id="ui-eholdings.package.create.custom" />}
             actionMenuItems={actionMenuItems}
-            firstMenu={historyState && historyState.eholdings && (
+            firstMenu={onCancel && (
```
wwilsman: Looks like `onCancel` is always provided to these `*-create` views from the create routes. So we either don't need this `onCancel &&` or the logic in those routes is meant to sometimes return a falsey value (it uses the ternary operator, so it always returns either `history.goBack` or `history.push`).

This applies to this `package-create` route and also the `title-create` route.

cherewaty: Changed to sometimes return a falsey value. We don't want there to be a back button if there's nowhere to go back to.

src/components/resource/edit-managed-title/resource-edit-managed-title.js
cherewaty: Removed all instances of `getDerivedStateFromProps` returning `prevState`.

src/components/resource/edit-managed-title/resource-edit-managed-title.js
cherewaty: Fixed just in case.

wwilsman:
> Skip tests that now would need an extra visit()

Could you elaborate on this slightly? I see they're all related to the `onCancel` with the creation routes, and with the adjustments there we don't render it when there is no history. Is this why they need to be skipped? Wasn't that the behavior before the changes (and why did they pass if it was)?

I'm hesitant to approve skipping tests unless there is good reason.

As a fix, we could either nest them one `describe` deeper with another `visit`, or we can move them to another suite (same file or otherwise) that would visit some other route before visiting the proper route.

cherewaty:
Commit added that un-skips the `onCancel` assertions (and fixes a regression!).

PR #570, opened 2018-09-24
UIEH-546: Update number of records found when results limited to first 10K

carolegodfrey:
## Purpose

Per https://issues.folio.org/browse/UIEH-546, there is currently a discrepancy in the 
total number of titles displayed in package detail information and number of records found in the title result list accordion. This specifically happens for a package when the number of related records (ie titles) found is greater than 10,000. This is currently a limitation in RM API as it does not return the exact number of titles and returns 10,000 instead. In these cases, UI should display `Over 10,000` as the number of records found to indicate this fact.

## Approach

- Check `totalResults` which is returned from a package's related resources request
For example - `eholdings/packages/19-5207/resources?page=1`
- Total results is available as `model.resources.length` in PackageShow Page. Package associated resources are obtained from a request to `getPackageTitles: (id, params) => Package.queryRelated(id, 'resources', params)` in PackageShow route.
- If totalResults is 10,000 for titles or 10,001 for providers or packages, pre-pend `Over` to count.
- Update to apply `Over` is made in accordion-list-header component
- This approach has limitation of potentially displaying Over 10,000 titles if a title list has exactly 10,000 titles.
- Added checks for providers and packages list (total results = 10,001) although these scenarios are unlikely due to number of providers and packages that are available
- Added tests for both Package and Provider show pages.

Original approach was to compare the number of titles (as returned from package details request meta data ) to the number of results (as returned from request to get related titles for a package). If package meta data is greater than number of results returned from list, add `Over` to the accordion header display. This approach unfortunately does not work as we only get package details and associated count for the entire package. If the package resources list is filtered  this case will not be handled 

#### TODOS and Open Questions
- [x] Create unit tests -- how to simulate large result list?

Need to adjust `json.meta.totalResults` for related resources - currently it gets set based on number of records in resources array (we need to be 10K)
https://github.com/folio-org/ui-eholdings/blob/71c8c748fbf6b28401f0098bb0a2d3e4fc874d4f/test/bigtest/network/helpers.js#L154

Tried overriding the response for retrieving a package's related resources in unit test. but not successful 
```
  this.server.get(`packages/${largeProviderPackage.id}/resources?page=1`,
        { 'data':[],
          'meta':{ 'totalResults': 10000 },
          'jsonapi':{ 'version':'1.0' } }, 200);
```
Was able to successfully override request (needed to exclude `?page=1`)

```
this.server.get(`packages/${largeProviderPackage.id}/resources`,
        { 'data':[],
          'meta':{ 'totalResults': 10000 },
          'jsonapi':{ 'version':'1.0' } }, 200);
```

- [x] AccordionListHeader is used in multiple places (as part of DetailsView). Understand impact of adding a new parameter to determine if Over gets displayed.
Added handling for titles and other cases 
- [x] Should over apply to only titles list and if the count is 10,000?
TotalResuts of of 10,000 applies to titles, 10,001 applies to packages and providers. 

## Learning

Sample package details request to mod-kb-ebsco 
`GET packages/19-5207`

Note - response includes Package details with titleCount  attribute: "titleCount":1552561 and indicates the total number of titles in a package

```{"data":{"id":"19-5207","type":"packages","attributes":{"contentType":"E-Book","customCoverage":{"beginCoverage":"","endCoverage":""},"isCustom":false,"isSelected":true,"name":"EBSCO eBooks","packageId":5207,"packageType":"Selectable","providerId":19,"providerName":"EBSCO","selectedCount":1,"titleCount":1554086,"vendorId":19,"vendorName":"EBSCO","visibilityData":{"isHidden":false,"reason":""},"allowKbToAddTitles":false,"packageToken":null,"proxy":{"id":"EZProxy","inherited":true}},"relationships":{"resources":{"meta":{"included":false}},"vendor":{"meta":{"included":false}},"provider":{"meta":{"included":false}}}},"jsonapi":{"version":"1.0"}}```

Sample request to GET package associated resources
`GET packages/19-5207/resources?page=1`

Note - response includes an array of package's related resources and totalResults  meta data:
"totalResults":10000 - indicates total number of titles for the search  and is capped at 10,000

```{"data":[{"id":"19-5207-4187001","type":"resources","attributes":{"description":null,"edition":null,"isPeerReviewed":null,"isTitleCustom":false,"publisherName":"Unspecified","titleId":4187001,"contributors":[],"identifiers":[{"id":"978-1-4619-6039-3","type":"ISBN","subtype":"Online"}],"meta":{"totalResults":10000},"jsonapi":{"version":"1.0"}}```

## Screenshots
Before:
screen shot 2018-09-23 at 8 59 56 pm
After:
screen shot 2018-09-23 at 9 02 59 pm

### DIFF
--- a/src/components/details-view/accordion-list-header.js
+++ b/src/components/details-view/accordion-list-header.js
@@ -1,18 +1,25 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { FormattedNumber, injectIntl, intlShape } from 'react-intl';
+import { FormattedMessage, FormattedNumber, injectIntl, intlShape } from 'react-intl';
 import { KeyValue } from '@folio/stripes-components';
 import AccordionHeader from '@folio/stripes-components/lib/Accordion/headers/DefaultAccordionHeader';
 import styles from './accordion-list-header.css';
 
 function AccordionListHeader(props) {
   let { intl } = props;
+  let showOver = props.metaLength > props.resultsLength;
   return (
     <div className={styles['accordion-list-header']}>
       <AccordionHeader {...props} />
       {props.open && (
         <KeyValue label={intl.formatMessage({ id: 'ui-eholdings.label.accordionList' })}>
           <div data-test-eholdings-details-view-results-count>
+            {showOver && (
+            <span>
+              <FormattedMessage id="ui-eholdings.over" />
+              &nbsp;
+            </span>
+            )}
             <FormattedNumber value={props.resultsLength} />
           </div>
         </KeyValue>
@@ -23,6 +30,7 @@ function AccordionListHeader(props) {
 
 AccordionListHeader.propTypes = {
   intl: intlShape.isRequired,
+  metaLength: PropTypes.number,
   open: PropTypes.bool,
   resultsLength: PropTypes.number
 };
--- a/src/components/details-view/accordion-list-header.js
+++ b/src/components/details-view/accordion-list-header.js
@@ -9,7 +9,7 @@ function AccordionListHeader(props) {
   let { intl } = props;
   // RM API does not return exact number of results when count is over 10K
   // For title lists, resultsLength of 10000 indicates this.
-  // For other lists (package and vendor) resultsLength of 10001 indicates this.
+  // For other lists (package and provider) resultsLength of 10001 indicates this.
   let overCount = props.listType === 'titles' ? 10000 : 10001;
   let showOver = props.resultsLength === overCount;
   let displayOverCount = 10000;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/components/details-view/accordion-list-header.js [hunk 1]
```diff
@@ -1,19 +1,34 @@
...
+import { FormattedMessage, FormattedNumber, injectIntl, intlShape } from 'react-intl';
 import { KeyValue } from '@folio/stripes-components';
 import AccordionHeader from '@folio/stripes-components/lib/Accordion/headers/DefaultAccordionHeader';
 import styles from './accordion-list-header.css';
 
 function AccordionListHeader(props) {
   let { intl } = props;
+  // RM API does not return exact number of results when count is over 10K
+  // For title lists, resultsLength of 10000 indicates this.
+  // For other lists (package and vendor) resultsLength of 10001 indicates this.
```
sduvvuri-ebsco: a very minor nitpick - could we please update the comment to say "provider" instead of "vendor"?

carolegodfrey: Update for correct terminology -- thanks

PR #581, opened 2018-10-04
Set onCancel to empty function instead of null

cherewaty:
Gets rid of
```
ERROR: 'Warning: Failed prop type: The prop `onCancel` is marked as required in `PackageCreate`, but its value is `null`.
```

### DIFF
--- a/src/components/package/create/package-create.js
+++ b/src/components/package/create/package-create.js
@@ -22,7 +22,7 @@ class PackageCreate extends Component {
   static propTypes = {
     handleSubmit: PropTypes.func.isRequired,
     intl: intlShape.isRequired,
-    onCancel: PropTypes.func.isRequired,
+    onCancel: PropTypes.func,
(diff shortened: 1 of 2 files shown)


src/routes/package-create.js
```diff
@@ -52,7 +52,7 @@ class PackageCreateRoute extends Component {
 
   render() {
     const { history, location } = this.props;
-    let onCancel = null;
+    let onCancel = () => {};
```
wwilsman: If `onCancel` is not required, having this be `undefined` should work. `let onCancel;`

cherewaty: hmmmmm still getting the prop type warning. I might just switch `onCancel` to not be required.

wwilsman: Oh yea. If it's required it should be a no-op. I think un-requiring is better than a no-op though

PR #590, opened 2018-10-09
Run tests on Jenkins and scan with SonarQube

cherewaty:


sduvvuri-ebsco:
Doesn't this mean that tests and lint will be run both by Circle and Jenkins? Are we going to disable Circle?

cherewaty:
The next step after this is to disable Circle, yes.

sduvvuri-ebsco:
Will Jenkins support running tests on both Chrome and Firefox like we do today in Circle?

funkymalc:
@sduvvuri-ebsco - I have a stub for FF but it isn't enabled.  I'll work on getting that enabled.

funkymalc:
@cherewaty - disable sonarqube in this PR.  I'll add it directly to master after you merge since the scan needs to run on master once before other branches can be scanned.

cherewaty:
SonarQube line removed.

cherewaty:
@funkymalc could you hop into Circle (you should be able to log in with your GitHub credentials) and turn the builds there off? It should be fairly easy to find, but let me know if you can't find the switch.

cherewaty:
https://circleci.com/gh/folio-org/ui-eholdings

PR #597, opened 2018-10-19
UIEH-455: Apply accordions to edit custom/managed resource detail record

anastasiaHahanova:
## Purpose
According to [UIEH-455](https://issues.folio.org/browse/UIEH-455) librarians should be able to expand and collapse any section displayed on the Edit Custom/Managed Resource Detail Record so that they can easily view the section that they most want to review.

## Approach
- Add the Accordion component from Stripes-components
- Wrap sections of the selected custom/managed resource in edit mode. The names of the sections are "Holding Status", "Resource settings" and"Coverage settings"
- Create a property containing the status of the sections(whether a section is expanded) in a local state.
- Pass appropriate props to Accordion components.
- Add associated tests.


sduvvuri-ebsco:
@anastasiaHahanova : code changes look good but before you commit to master, please check if you have to fix the lint issues that are reported above by Jenkins.

PR #598, opened 2018-10-19
UIEH-121: add title search results sorting

MaksymDryha:
## Purpose
Users should have the ability to choose in which manner to sort title search results (by relevance or alphabetically)

## Approach
* Adjust TitleSearchFilters component so that sorting filter appears among other filters 
* Write related tests

### Before:
image

### DIFF
--- a/src/components/title-search-filters.js
+++ b/src/components/title-search-filters.js
@@ -17,6 +17,14 @@ function TitleSearchFilters(props) {
     <SearchFilters
       searchType="titles"
       availableFilters={[{
+        name: 'sort',
+        label: intl.formatMessage({ id: 'ui-eholdings.label.sortOptions' }),
+        defaultValue: 'relevance',
+        options: [
+          { label: intl.formatMessage({ id: 'ui-eholdings.filter.sortOptions.relevance' }), value: 'relevance' },
+          { label: intl.formatMessage({ id: 'ui-eholdings.label.title' }), value: 'name' }
+        ]
+      }, {
         name: 'selected',
         label: intl.formatMessage({ id: 'ui-eholdings.label.selectionStatus' }),
         defaultValue: 'all',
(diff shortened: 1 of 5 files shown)


MaksymDryha:
@sduvvuri-ebsco, I decided not to use those methods, because it wouldn't do a lot of profit. Nothing changes on the frontend side, depending on how returned items are sorted.

MaksymDryha:
@sduvvuri-ebsco, I decided not to use those methods, because it wouldn't do a lot of profit. Nothing changes on the frontend side, depending on how returned items are sorted.

MaksymDryha:
@sduvvuri-ebsco, I decided not to use those methods, because it wouldn't do a lot of profit. Nothing changes on the frontend side, depending on how returned items are sorted.

MaksymDryha:
@sduvvuri-ebsco, I decided not to use those methods, because it wouldn't do a lot of profit. Nothing changes on the frontend side, depending on how returned items are sorted.

MaksymDryha:
@sduvvuri-ebsco, I decided not to use those methods, because it wouldn't do a lot of profit. Nothing changes on the frontend side, depending on how returned items are sorted.

MaksymDryha:
@sduvvuri-ebsco, I decided not to use those methods, because it wouldn't do a lot of profit. Nothing changes on the frontend side, depending on how returned items are sorted.

MaksymDryha:
@sduvvuri-ebsco, I decided not to use those methods, because it wouldn't do a lot of profit. Nothing changes on the frontend side, depending on how returned items are sorted.

carolegodfrey:
Just an added note on the sort helper method for the tests -- they get applied under the hood. You should not need to explicitly call them. In this case the sort results will only be different if there is a multiple term search performed. For example searching for 'heart attack' will return results containing both heart and attack as more relevant.

PR #604, opened 2018-10-26
Fix moment usage to UTC

wwilsman:
## Purpose

This is related to [STCOR-265](https://issues.folio.org/browse/STCOR-265) to get tests to run in consistent timezones.

While investigating, I found that the app _is_ running in the UTC timezone during testing. Poking around in eholdings revealed that it was actually `moment` which was causing the dates to be incorrect.

## Approach

In other places in eholdings, we are using `.tz('UTC')` to set the timezone to be consistent for manipulation or comparison. We were **not** doing this for the date validation functions.

Moment has a `.utc()` method to set the moment object to UTC mode for every manipulation. All instances of `moment()` were replaced with `moment.utc()` and the existing places that used `.tz()` were also updated to `.utc()`.

It should be safe to do this for manipulation and comparison of dates within the code and just let `intl` handle adjusting the timezone for display. However correct me if I'm wrong, and we can try to come up with some other solution to fix this issue.

### Update

There were some failing tests related to `Datepicker` (see below). This is because the datepicker component uses `moment(value, dateFormat)` which assumes local time. But we are using UTC time because that's how the server stores those dates. There is no way to tell the datepicker component which timezone or format we are providing to it, so the work-around is instead of providing the date string to `Datepicker`, use a `moment.utc` object which will not be converted to local time by `moment()`.

## Learning

- [Moment docs on UTC mode](http://momentjs.com/docs/#/parsing/utc/)

### DIFF
--- a/src/components/package/_fields/custom-coverage/package-coverage-fields.js
+++ b/src/components/package/_fields/custom-coverage/package-coverage-fields.js
@@ -85,7 +85,7 @@ export function validate(values, props) {
   values.customCoverages.forEach((dateRange, index) => {
     let dateRangeErrors = {};
 
-    if (dateRange.beginCoverage && !moment(dateRange.beginCoverage).isValid()) {
+    if (dateRange.beginCoverage && !moment.utc(dateRange.beginCoverage).isValid()) {
       dateRangeErrors.beginCoverage = intl.formatMessage({ id: 'ui-eholdings.validate.errors.dateRange.format' }, { dateFormat });
     }
 
@@ -93,7 +93,7 @@ export function validate(values, props) {
       dateRangeErrors.beginCoverage = intl.formatMessage({ id: 'ui-eholdings.validate.errors.dateRange.format' }, { dateFormat });
     }
 
-    if (dateRange.endCoverage && moment(dateRange.beginCoverage).isAfter(moment(dateRange.endCoverage))) {
+    if (dateRange.endCoverage && moment.utc(dateRange.beginCoverage).isAfter(moment.utc(dateRange.endCoverage))) {
       dateRangeErrors.beginCoverage = intl.formatMessage({ id: 'ui-eholdings.validate.errors.dateRange.startDateBeforeEndDate' });
     }
 
--- a/src/components/package/_fields/custom-coverage/package-coverage-fields.js
+++ b/src/components/package/_fields/custom-coverage/package-coverage-fields.js
@@ -35,7 +35,7 @@ class PackageCoverageFields extends Component {
             type="text"
             component={Datepicker}
             label={intl.formatMessage({ id: 'ui-eholdings.date.startDate' })}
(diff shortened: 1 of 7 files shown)


carolegodfrey:
Thanks for making the moment adjustments @wwilsman - those updates make sense for making comparisons of dates.
However, I still see issues when running unit tests -- per below image.
I tried to recreate the issue (as experienced by the Spitfire team when running unit tests) - 
Specifically,  I set my systems local time zone to "Eastern European Standard Time" which I believe is 2 hours ahead of UTC and then ran the unit tests.Per below output, some of the tests are failing with the display of dates ( actual date is 1 day earlier)
screen shot 2018-10-28 at 9 30 17 pm

Yurii-Danylenko:
> Thanks for making the moment adjustments @wwilsman - those updates make sense for making comparisons of dates.
> However, I still see issues when running unit tests -- per below image.
> I tried to recreate the issue (as experienced by the Spitfire team when running unit tests) -
> Specifically, I set my systems local time zone to "Eastern European Standard Time" which I believe is 2 hours ahead of UTC and then ran the unit tests.Per below output, some of the tests are failing with the display of dates ( actual date is 1 day earlier)
> screen shot 2018-10-28 at 9 30 17 pm

Same for me. I assume that an issue is inside a mirage server logic.

wwilsman:
Interesting... I also tested in the EEST but my tests passed! In fact, I put an `.only` on those specific resource edit coverage tests that you posted a screenshot of. 🤔 

I'll look into it more today and see if maybe I missed something, or if my search-and-replace affected something It shouldn't.

carolegodfrey:
Was wondering if it also has to do with the time of the day that the test are run?

wwilsman:
### Update: 

_The change to `moment.utc` did indeed fix the issue in eholdings._ 

Notice that the remaining failing tests are related only to displaying the existing date values within the date inputs? Digging around shows that the `format` property of the date `Field` **is correct**, but for some reason, the `DatePicker` component still renders the incorrect date.

Still investigating further to see exactly _what_ in the `DatePicker` is causing this and if we can work around it for now

wwilsman:
@carolegodfrey could you try again? I actually didn't catch this failure before because I was using `only` on managed dates not custom dates. This workaround fixes the datepicker parsing issue for me locally, but I'd like to confirm with you.

carolegodfrey:
@wwilsman -- I re-ran the tests agains and no longer see the date display issue with your latest fix -- thank you!!

wwilsman:
Yay! 🎉 

Is there anything I can do to rerun Jenkins without force pushing? I'm experiencing some of those convergence timeouts you warned me about. First the `pr-merge` failed, then the `branch` failed. I'd like to just rerun `branch` to avoid `pr-merge` failing again

carolegodfrey:
I just tried re-running the branch -- hope it builds ok

wwilsman:
Thanks @carolegodfrey !!

PR #605, opened 2018-10-27
Accommodate lack of publisher name in title results

cherewaty:
## Purpose
When searching for a title, sometimes a result doesn't include a publisher name. That would result in blank lines.

## Approach
- Combine the publication type and publisher name into a single line.
- Show publication type first, since anecdotally the publication type seems to always be defined.
- If both the publication type and the publisher name are present, separate them with a bullet.
- Long publisher names get cut off with an ellipsis.
- Now that all title search results only have two lines, decrease the result height from 80px to 60px.
- Skeleton loading state for title search results updated to only include two lines.

### Before
localhost_3000_eholdings_searchtype titles q health iphone 5_se 1

### After
localhost_3000_eholdings_searchtype titles q health iphone 5_se

### DIFF
--- a/src/components/title-list-item/title-list-item.css
+++ b/src/components/title-list-item/title-list-item.css
@@ -54,19 +54,16 @@
   }
 
   &.is-publisher-and-type-visible {
-    /* 1. name  2. publisher  3. type */
+    /* 1. name  2. type and publisher */
     background-image:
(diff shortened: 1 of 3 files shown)


cherewaty:
Got the thumbs up to proceed with this.

PR #610, opened 2018-11-02
Use children-based <Select> for easier i18n

cherewaty:
Prefers...
```
<Select>
  <option value="">Label</option>
  ...
</Select>
```
... over...
```
<Select dataOptions={options}>
```

so we can `<FormattedMessage>` with `<Select>`. The fewer instances of `injectIntl()` the better.


src/components/title/_fields/contributor/contributor-field.js
```diff
@@ -34,12 +34,17 @@ class ContributorField extends Component {
             autoFocus={Object.keys(fields.get(index)).length === 0}
             label={intl.formatMessage({ id: 'ui-eholdings.type' })}
```
wwilsman: I think we can clean up these labels as well

cherewaty: Happening in a separate PR 😁

PR #611, opened 2018-11-02
Use FormattedMessage for aria-label i18n

cherewaty:
Uses `<FormattedMessage>` instead of `intl.formatMessage().`

### DIFF
--- a/src/components/title/create/title-create.js
+++ b/src/components/title/create/title-create.js
@@ -8,7 +8,7 @@ import {
   PaneHeader
 } from '@folio/stripes/components';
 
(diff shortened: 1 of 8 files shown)


src/components/title/create/title-create.js
```diff
@@ -78,12 +78,16 @@ class TitleCreate extends Component {
             paneTitle={intl.formatMessage({ id: 'ui-eholdings.title.create.paneTitle' })}
```
wwilsman: We can clean up these pane titles too, but probably another PR

PR #612, opened 2018-11-03
Remove all intl.formatMessage()

cherewaty:
Removes all instances of `intl.formatMessage()` in favor of the preferred `<FormattedMessage>`.

`react-intl` provides the two different ways of getting localized strings, but `<FormattedMessage>` provides better React ergonomics, and means we don't have to `injectIntl()` all over the place.


src/components/package/create/package-create.js
```diff
@@ -123,7 +121,7 @@ class PackageCreate extends Component {
 }
 
 const validate = (values, props) => {
-  return Object.assign({}, validatePackageName(values, props), validateCoverageDates(values, props));
+  return Object.assign({}, validatePackageName(values), validateCoverageDates(values, props));
 };
 
 export default injectIntl(reduxForm({
```
Godlevskyi: @cherewaty, I wanted to suggest for such compositions to use compose functions for example from redux.
I belive it would be much more readable, like:
compose(
injectIntl,
reduxForm({
...
})(PackageCreate)

what do you think?

docs/routing-architecture/routing-architecture.md
```diff
@@ -214,7 +214,7 @@ class PackageShow extends Component {
 
   render() {
     // get the model off props that was passed into the View
-    let { model, fetchPackageTitles, intl } = this.props;
+    let { model, fetchPackageTitles } = this.props;
```
Godlevskyi: Jeffrey, I would suggest to write all the destructured props on the new line. I understand it is arguable
The big advantede of it is in tracking of fuether changes e.g. if changed one prop than git will track only that prop line change.
I belive it will decrease the amout of merge conflicts and makes I belive more readable code.

src/components/resource/_fields/custom-coverage/custom-coverage-fields.js
```diff
@@ -134,29 +131,26 @@ const validateWithinPackageRange = (dateRange, packageCoverage, intl) => {
...
-        timeZone: 'UTC',
-        year: 'numeric',
-        month: 'numeric',
-        day: 'numeric'
-      }
-    );
+    let startDate =
+      <FormattedDate
+        value={packageBeginCoverageDate}
+        timeZone="UTC"
```
Godlevskyi: @cherewaty, Shouldn't we adjust the date/time to the local timeZone?

cherewaty: I believe in this case, no, since the dates we're dealing with are "fixed" like a user's birthday would be, and do not change based on time zone.

PR #613, opened 2018-11-05
Use field-level validation on package form

cherewaty:
## Purpose
Most of the validations we're doing in forms are solely on the field level, but we're registering them all at the global level of each `redux-form`. That leads to some clunkier-than-necessary exports of `validate()` functions and `injectIntl()` happening one level higher than it needs to.

## Approach
Use field-level validations where possible.

I ran into one exception: validating that a coverage end date is after its begin date. Redux Form provides a `<Fields>` component that could handle that in an elegant way, but the `stripes-components` `<Datepicker>` is still too hard-wired to being the child of a `<Field>` to be able to pull that off. When https://issues.folio.org/browse/STCOM-325 is done, we can make that better.

## Next steps
Apply this same pattern to title and resource forms.

### DIFF
--- a/src/components/package/_fields/name/package-name-field.js
+++ b/src/components/package/_fields/name/package-name-field.js
@@ -4,7 +4,21 @@ import { FormattedMessage } from 'react-intl';
 
 import { TextField } from '@folio/stripes/components';
 
-function PackageNameField() {
+const validate = (value) => {
+  let errors;
+
+  if (value === '') {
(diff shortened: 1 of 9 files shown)


src/components/package/_fields/name/package-name-field.js
```diff
@@ -4,6 +4,20 @@ import { FormattedMessage } from 'react-intl';
 
 import { TextField } from '@folio/stripes/components';
 
+export function validate(value) {
```
wwilsman: No need to export anymore, right?

PR #614, opened 2018-11-05
Refactor title forms validation

cherewaty:
Continues field-level validation patterns established in https://github.com/folio-org/ui-eholdings/pull/613 for title forms.

The `initialValue` pattern with contributors and identifiers has a funky smell. React Final Form Array has `meta.initial` to make that better, but Redux Form's `FieldArray` is lacking that property.


src/components/title/_fields/custom-url/custom-url-fields.js
```diff
@@ -0,0 +1,35 @@
...
+      errors = <FormattedMessage id="ui-eholdings.validate.errors.customUrl.include" />;
+    }
+
+    return errors;
+  }
+
+  render() {
+    return (
+      <div>
+        <div data-test-eholdings-custom-url-textfield>
```
wwilsman: Is there a structural reason for these nested `div`s?

cherewaty: There isn't. Removed. I should do another pass on `<Field>`s to see if we can put `data-test-*` attributes on those to render on their children `stripes-components`. We have a lot of `<div>`s whose sole purpose is to apply a `data-test-*`.

PR #622, opened 2018-11-20
UIEH-573 Replace actionMenuItems with actionMenu

cherewaty:
## Purpose
The composable `actionMenu` prop is replacing the config-object-style `actionMenuItems`: https://github.com/folio-org/stripes-components/pull/705

Resolves https://issues.folio.org/browse/UIEH-573

## Approach
- I introduced a `getActionMenu()` method on most components using this new pattern.
- In `onClick()` events within the pane header dropdowns, I had to explicitly call `onToggle()` to close the dropdown.
- Introduced new `FullViewLink` component to DRY up CSS.
- I had to add `fullWidth` to the `buttonStyle` to prevent CSS sibling selectors from adding margin to any button after the first one. I'll follow up with a `stripes-components` PR to prevent having to do that in the future.

### Before
screen shot 2018-11-20 at 3 26 28 pm

### After
screen shot 2018-11-20 at 3 22 29 pm


rasmuswoelk:
Regarding the necessesity of the `fullWidth` buttonStyle:

I have updated the dropdownItem style in this PR:
https://github.com/folio-org/stripes-components/pull/723

Here I have forced 100% width and removed margin for this specific button style. Once this is merged in you don't need to add the fullWidth style.

PR #623, opened 2018-11-21
Apply react-final-form to title forms

cherewaty:
## Purpose
`react-final-form` fits our use case better than `redux-form`, and is better maintained.

Continues the work from https://github.com/folio-org/ui-eholdings/pull/608.

## Approach
- I made some interactor tweaks to better take advantage of `collection` for identifiers and contributors.
- Getting this working with `<RepeatableField>` took a lot of iterations, but I got there.
- `initialValues` isn't explicitly required when setting up a `Form`, so I got to remove some code there.


src/components/title/_fields/name/title-name-field.js
```diff
@@ -1,17 +1,17 @@
...
+import { Field } from 'react-final-form';
 import { FormattedMessage } from 'react-intl';
 
 import { TextField } from '@folio/stripes/components';
 
 function validate(value) {
   let errors;
 
-  if (value === '') {
+  if (value === undefined) {
```
wwilsman: Should we check for a falsey value here to cover `null`, `undefined`, & `''`?

src/components/title/_forms/add-title-to-package-form.js
```diff
@@ -0,0 +1,21 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+
+import PackageSelectField from '../_fields/package-select';
+import CustomURLField from '../_fields/custom-url';
+
+function AddTitleToPackageForm({ packageOptions }) {
```
wwilsman: Is this still accurate as a `*Form` with these changes? If this becomes a common pattern, would something like `*Fieldset` better describe these field groups that don't actually belong to forms yet?

cherewaty: The `react-final-form` docs call this concept a "field group", so I changed to that nomenclature.

src/components/title/_fields/name/title-name-field.js
cherewaty: Changed.

PR #628, opened 2018-12-04
UIEH-582 Don’t show edit link for managed titles

cherewaty:
## Purpose
Managed title pages shouldn't link to an edit page, since there's nothing editable.

Regression introduced in https://github.com/folio-org/ui-eholdings/pull/569.

## Approach
Only show the edit link when the title is custom.

Tests added to verify fixed regression.

### DIFF
--- a/test/bigtest/tests/title-show-test.js
+++ b/test/bigtest/tests/title-show-test.js
@@ -56,7 +56,7 @@ describe('TitleShow', () => {
       expect(TitleShowPage.nameHasFocus).to.be.true;
     });
 
-    it('does not display the edit button for a managed title', () => {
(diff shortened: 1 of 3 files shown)


test/bigtest/tests/title-show-test.js
```diff
@@ -56,6 +56,10 @@ describe('TitleShow', () => {
       expect(TitleShowPage.nameHasFocus).to.be.true;
     });
 
+    it('does not display the edit button for a managed title', () => {
```
wwilsman: We probably want `it.always` here unless the button was previously visible

cherewaty: Added.

PR #634, opened 2018-12-11
STCOM-274 Pane Animations

wwilsman:
## Purpose

This begins work on [STCOM-274](https://issues.folio.org/browse/STCOM-274). While the panes in eholdings were already responsive, they were decoupled from the rest of stripes and a good place to start iterating on the next version of panes and panesets while also migrating eholdings to use stripes panes at the same time.

## Approach

At first I took the existing pane and paneset components from `stripes-components` and dropped them into eholdings to start refactoring. It turned out that the majority of pane and paneset code was concerning animations which were not yet implemented. After taking out the non-implemented animations, the two components were completely gutted, and this turned into a total rewrite of panes instead of a refactor.

Instead of trying to manage mounting and unmounting the panes and calculating width to do all of the animation work ourselves, I leveraged `CSSTransition` from `react-transition-group`. The paneset, without its old context, is now almost entirely empty with only a prop-types check to ensure children are `Pane`s. This may not be necessary if we're planning on allowing non-panes or other components that render panes.

`CSSTransition` will apply an `enter` class followed by an `enter-active` class one tick later to allow applying styles at the beginning of an animation (`exit` classes for the end of the animation). Using the provided callbacks, we can also set the custom `flexGrow` style when we animate in and remove it when we start animating out. This allows us to forgo manually calculating the widths of all panes within a paneset.

On larger sizes, the panes animate their width on enter / exit. On smaller sizes they slide in from the side and vignettes fade in. This corresponds with the [FOLIO UX Guidelines here](https://ux.folio.org/universal-navigation/uxi-3/).

**Nested routes can have issues with animations.** When there is a route within the pane, and the pane animates due to a route change, the route will update before the animation begins. This can result in routes that change, or become unmounted before the animation. To work around this, the `<Switch>` component can accept a `location` prop which will override the `location` context from the router. With this knowledge, we can defer changing the `location` of nested routes within panes until after the animation is done with `onExited`. In eholdings, the `detailsView` prop in the search-paneset is actually a `<Switch>` component from the eholdings root. In the search-paneset, we clone the `detailsView` with the location prop, and wait until after animating to remove it.

#### Pane props

- `aside` is used for panes that animate from the opposite direction and are only tangentially related to the main content (such as navigation in settings, or filters in search).
- `static` is used for panes which will always be visible. This is usually just a single pane such as the results pane in search.
- `visible` toggles the panes visibility. When mounting for the first time, the pane will mount in its desired state, only animating when this prop is toggled.
- `onDismiss` is called when the vignette of a non-static pane is clicked. Nothing is done to actually dismiss the pane, this callback can be used for toggling the `visible` prop.
- `onEnter`, `onEntering`, `onEntered`, `onExit`, `onExiting`, `onExited` are all callback props accepted by `CSSTransition` group. Useful for performing actions before or after animation states.
- `flexGrow` replaces `defaultWidth`. It is used in the `style` of the pane after animating in and as such does exactly what the `flexGrow` [css property](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow) does.

#### Future work
- Use these same components in the settings view. That is what the original ticket refers to.
- Make adjustments for use in settings. That ticket is more about adding buttons similar to what the details button in eholdings does, which is essentially a "go back" button.
- This wasn't tested with more than one right-aligned pane. I think settings may be the place for this as the main navigation in settings would not be an aside, but the `static` pane with two right-aligned panes.
- The `max-width` of non-static panes at smaller sizes might not work well in settings. If that's the case, it should be moved to `search-paneset` where we can apply a custom class to the details pane.

## Learning
- https://reactcommunity.org/react-transition-group/
- https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow

## Screenshots

**Large Screens**

**Medium Screens**

**Small Screens**


test/bigtest/tests/provider-search-test.js
```diff
@@ -299,7 +299,7 @@ describe('ProviderSearch', () => {
           expect(this.location.search).to.include('sort=name');
         });
 
-        it('shows search filters on smaller screen sizes (due to filter change only)', () => {
+        it.skip('shows search filters on smaller screen sizes (due to filter change only)', () => {
```
wwilsman: I noticed this test was skipped already in `package-search-test.js`. I skipped it in all of these tests too because it's actually quite confusing. It fails, but the test title `it shows search filters` contradicts the assertion within `expect(isSearchVignetteHidden).to.equal(false)`. When the vignette is hidden, that means the search filters are also hidden, but the test title suggests we want them to be visible?

rasmuswoelk:
The transitions look good! I'll be looking at the implementation later but I would just post this test site:

http://folio-styling.testing.samhaeng.com

I uploaded your changes here for people to test it out. Note: It uses the folio-testing API.

rasmuswoelk:
Hi @wwilsman 

Firstly, these additions are awesome! I'm really looking forward to getting this implemented in the whole FOLIO app.

The implementation looks good to me but there's probably some people who has a better understanding of the various use-cases and the requirements for this functionality.

Here's a few things that I noticed:

1. @cherewaty changed the transition duration to 200ms in the CSS but it remains at 300ms in the CSSTransition timeout in Pane.js:110. I assume that these values should be identical?

2. The transition timing function is set to "ease-out" (fast/abrupt at the beginning, slow at the end) in the prototype. Can we try setting the same timing function in this code?

3. The content wraps when entering and exiting which makes the layout change slightly while transitioning. We want to avoid this some way. Maybe by adding an inner wrapper element or similar. (It's wrapping in the prototype but in the real implementation it shouldn't)

dec-17-2018 14-48-35

cherewaty:
Commit added that adopts some recommendations from https://material.io/design/motion/speed.html:
- Stretched the entering animation back out to 300ms; exiting is 250ms, to match the "large" size recommendation: _Animated elements that traverse a large portion of the screen have the longest durations._
- Went with `ease-out` for entering and `ease-in` for exiting: _Elements exiting a screen use acceleration easing, where they start at rest and end at peak velocity._

cherewaty:
I'll have to think some more about the wrapping problem 🤔

rasmuswoelk:
@cherewaty I think this is a very good start. And yes, we can always iterate later. Great work!!

PR #636, opened 2018-12-12
UIEH-581: Fix displaying of the comma separator for custom date ranges

Yurii-Danylenko:
https://issues.folio.org/browse/UIEH-581

## Purpose

  There is an issue related to comma separator for custom coverage date ranges on custom and managed title edit pages. The Comma is displaying when there are empty date ranges. The reason is empty objects which are passing to the CoverageDateList component. So the component renders them as empty strings with a comma separator.

## Approach

 - Filter empty date ranges and pass to the CoverageDateList component only filled date ranges.
 - Add tests
 - Add property type check to CoverageDateList

## Screenshots
managed title 1

managed title 2

managed title 3

custom title 1

custom title 2

custom title 3

### DIFF
--- a/src/components/coverage-date-list/coverage-date-list.js
+++ b/src/components/coverage-date-list/coverage-date-list.js
@@ -3,9 +3,28 @@ import PropTypes from 'prop-types';
 import moment from 'moment';
 import { FormattedDate, FormattedMessage } from 'react-intl';
 
+const containsNonEmptyObjectsWithStringValues = (propValue, key, componentName, location, propFullName) => {
+  const BEGIN_COVERAGE = 'beginCoverage';
+  const END_COVERAGE = 'endCoverage';
+  const error = new Error(`Invalid prop \`${propFullName}\` supplied to \`${componentName}\`. Validation failed.`);
+
+  const curObject = propValue[key];
+  const curObjectKeys = Object.keys(curObject);
+  const containsBeginCoverageProp = curObjectKeys.includes(BEGIN_COVERAGE);
+  const containsEndCoverageProp = curObjectKeys.includes(END_COVERAGE);
+  const doesNotContainBeginAndEndCoverage = !containsBeginCoverageProp && !containsEndCoverageProp;
+  const beginCoverageIsNotOfStringType = containsBeginCoverageProp && typeof curObject.beginCoverage !== 'string';
+  const endCoverageIsNotOfStringType = containsEndCoverageProp && typeof curObject.endCoverage !== 'string';
+  const propTypeIsWrong = doesNotContainBeginAndEndCoverage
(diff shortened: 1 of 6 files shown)


src/components/coverage-date-list/coverage-date-list.js
```diff
@@ -3,9 +3,38 @@ import PropTypes from 'prop-types';
 import moment from 'moment';
 import { FormattedDate, FormattedMessage } from 'react-intl';
 
+/* eslint-disable consistent-return */
+const containsNonEmptyObjectsWithStringValues = (propValue, key, componentName, location, propFullName) => {
+  const error = new Error(
+    'Invalid prop `' + propFullName + '` supplied to' +
+    ' `' + componentName + '`. Validation failed.'
+  );
```
anastasiaHahanova: isn't it better to use template strings here?

src/components/coverage-date-list/coverage-date-list.js
```diff
@@ -3,9 +3,38 @@ import PropTypes from 'prop-types';
...
+    return error;
+  }
+
+  if (containsBeginCoverageProp && typeof curObject.beginCoverage !== 'string') {
+    return error;
+  }
+
+  if (containsEndCoverageProp && typeof curObject.endCoverage !== 'string') {
+    return error;
+  }
```
anastasiaHahanova: possibly it will be better to refactor this to (u can use other variable name):
```suggestion
const isError =  (!containsBeginCoverageProp && !containsEndCoverageProp)
        || (containsBeginCoverageProp && typeof curObject.beginCoverage !== 'string') 
        || (containsEndCoverageProp && typeof curObject.endCoverage !== 'string') ;
        
  if (isError) {
     return error;
   }
```

src/components/coverage-date-list/coverage-date-list.js
```diff
@@ -3,9 +3,35 @@ import PropTypes from 'prop-types';
 import moment from 'moment';
 import { FormattedDate, FormattedMessage } from 'react-intl';
 
+/* eslint-disable consistent-return */
```
cherewaty: Does placing `return {}` at the end of `containsNonEmptyObjectsWithStringValues()` work? That might be preferable to the `eslint-disable` declaration.

Yurii-Danylenko: Have deleted `eslint-disable` and added `return null` at the end.

Yurii-Danylenko: Jeffrey, are you OK with this?

cherewaty: 👍 yup, that's better.

PR #658, opened 2019-01-03
UIEH-605 Rename initialValue to initial

cherewaty:
## Purpose
Part of https://issues.folio.org/browse/UIEH-605

## Approach
Instead of passing in an `initialValue` to pre-packaged field components, we'll be able to pick it off `meta.initial` with React Final Form.
### Before this PR
```
renderRepeatableField = ({ fields, name }) => {
    const { initialValue } = this.props;

    const hasAddButton = fields.length === 0 || (fields.length === 1 && !initialValue[0]);
    const hasEmptyMessage = initialValue.length > 0 && initialValue[0].beginCoverage;
...
```

### After this PR
```
renderRepeatableField = ({ fields, name }) => {
    const { initial } = this.props;

    const hasAddButton = fields.length === 0 || (fields.length === 1 && !initial[0]);
    const hasEmptyMessage = initial.length > 0 && initial[0].beginCoverage;
...
```

### Later, with React Final Form
```
renderRepeatableField = ({ fields, name, meta: { initial } }) => {
    const hasAddButton = fields.length === 0 || (fields.length === 1 && !initial[0]);
    const hasEmptyMessage = initial.length > 0 && initial[0].beginCoverage;
...
```

Making this change now will reduce the changes needed with the full migration to React Final Form.

### DIFF
--- a/src/components/resource/_fields/custom-embargo/custom-embargo-fields.js
+++ b/src/components/resource/_fields/custom-embargo/custom-embargo-fields.js
@@ -15,7 +15,7 @@ import styles from './custom-embargo-fields.css';
 class CustomEmbargoFields extends Component {
   static propTypes = {
     change: PropTypes.func.isRequired,
-    initialValue: PropTypes.object,
+    initial: PropTypes.object,
     showInputs: PropTypes.bool
   };
 
@@ -36,7 +36,7 @@ class CustomEmbargoFields extends Component {
   }
 
   render() {
-    let { initialValue } = this.props;
+    let { initial } = this.props;
     let { showInputs } = this.state;
 
     return (showInputs) ? (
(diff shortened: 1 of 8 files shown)


src/components/resource/_fields/custom-embargo/custom-embargo-fields.js
```diff
@@ -15,7 +15,7 @@ import styles from './custom-embargo-fields.css';
 class CustomEmbargoFields extends Component {
   static propTypes = {
     change: PropTypes.func.isRequired,
-    initialValue: PropTypes.object,
+    initial: PropTypes.object,
     showInputs: PropTypes.bool
   };
```
Godlevskyi: could you add the non-requred prop to the defaultProps, please?

cherewaty: 👍 Good catch. Added.

PR #661, opened 2019-01-08
UIEH-590: Bug: Unsaved changes modal not firing on return to title and package creation forms

Godlevskyi:
## Purpose

 This PR resolves the https://issues.folio.org/browse/UIEH-590 Bug: Unsaved changes modal not firing on return to title and package creation form.

## Approach
Inspecting the props of the `<TitleCreate>` component, when returning to the creation form request.isResolved is already true.
 
In order to solve it the new action for clearing the create requests was added.

So, the approach was to clear the create requests when we unmount the components.


src/routes/package-create.js
```diff
@@ -78,6 +84,7 @@ export default connect(
   ({ eholdings: { data } }) => ({
     createRequest: createResolver(data).getRequest('create', { type: 'packages' })
   }), {
-    createPackage: attrs => Package.create(attrs)
+    createPackage: attrs => Package.create(attrs),
+    removeCreateRequests: () => Package.removeCreateRequests(),
```
Yurii-Danylenko: Please, delete function wrapper

src/routes/title-create.js
```diff
@@ -74,10 +75,11 @@ class TitleCreateRoute extends Component {
     return (
       <TitleManager record="New custom title">
         <View
-          request={createRequest}
+          request={this.props.createRequest}
```
Yurii-Danylenko: Please move to destructurization

src/routes/title-create.js
```diff
@@ -97,6 +99,7 @@ export default connect(
     };
   }, {
     createTitle: attrs => Title.create(attrs),
+    removeCreateRequests: () => Title.removeCreateRequests(),
```
Yurii-Danylenko: Please, delete function wrapper

src/components/package/create/package-create.js
```diff
@@ -25,9 +25,14 @@ class PackageCreate extends Component {
     onCancel: PropTypes.func,
     onSubmit: PropTypes.func.isRequired,
     pristine: PropTypes.bool.isRequired,
+    removeCreateRequests: PropTypes.func.isRequired,
```
Yurii-Danylenko: Do we have a timestamp in the request here?

PR #671, opened 2019-01-12
UIEH-605 Use React Final Form with packages

cherewaty:
## Purpose
Switches package edit and creation forms to use React Final Form instead of Redux Form. Part of https://issues.folio.org/browse/UIEH-605.

## Approach
- I was able to completely move `validate()` for package coverage dates to be encapsulated in its accompanying component, so it no longer has to be exported and consumed by parent forms. That's a huge architectural win I'm very excited about with `react-final-form` - `<FieldArray>` has a working `validate` prop!
- I had problems using `<RadioButtonGroup>` from `stripes-components` with React Final Form's `<Field>`; instead I just converted the the relevant fields to sets of individual `<Field>`s that have `component` prop `RadioButton`.
- Just like https://github.com/folio-org/ui-eholdings/pull/669, I duplicated the proxy and token fields. I'll DRY that up with the sequel to this PR, moving resource editing to React Final Form.

There are several more simple architectural enhancements that could be made here, but this gets the package forms successfully migrated to React Final Form.

### DIFF
--- a/src/components/package/_fields/custom-coverage/package-coverage-fields.js
+++ b/src/components/package/_fields/custom-coverage/package-coverage-fields.js
@@ -1,6 +1,6 @@
 import React, { Component, Fragment } from 'react';
-import { Field, FieldArray } from 'redux-form';
-import PropTypes from 'prop-types';
+import { Field } from 'react-final-form';
+import { FieldArray } from 'react-final-form-arrays';
 import moment from 'moment';
 import {
   FormattedMessage,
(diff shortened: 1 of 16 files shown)


src/components/package/_fields/custom-coverage/package-coverage-fields.js
```diff
@@ -18,13 +18,28 @@ import styles from './package-coverage-fields.css';
...
+      if (endCoverage && !moment.utc(endCoverage).isAfter(moment.utc(beginCoverage))) {
+        errors.beginCoverage = <FormattedMessage id="ui-eholdings.validate.errors.dateRange.startDateBeforeEndDate" />;
+      }
+
+      errorArray.push(errors);
+    });
+
+    if (errorArray.some(object => Object.keys(object).length)) {
+      return errorArray;
+    } else {
```
Godlevskyi: no need for the else clause.

cherewaty: Commit added simplifying this function.

Godlevskyi: Thanks!

PR #679, opened 2019-01-20
STCOM-387 Eject from Settings smart component

cherewaty:
## Purpose
https://issues.folio.org/browse/STCOM-387

The `<Settings>` smart component contains routing glue, which is at odds with the routing-based architecture strategy outlined in https://issues.folio.org/browse/STRIPES-589.

## Approach
We have all the UI we need from `stripes-components` to be able to construct this functionality with the `<Settings>` smart component, in a way that `ui-eholdings` is in complete control of its settings routing.

### Breaking change
The `<Settings>` smart component attempts to programmatically alphabetize the items in each `<NavListSection>` - this PR does not maintain that functionality.

### DIFF
--- a/src/components/settings.js
+++ b/src/components/settings.js
@@ -1,0 +1,64 @@
+import React, { Component, Fragment } from 'react';
+import PropTypes from 'prop-types';
+import ReactRouterPropTypes from 'react-router-prop-types';
+import { FormattedMessage } from 'react-intl';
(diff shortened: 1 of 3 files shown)


cherewaty:
Related to https://github.com/folio-org/ui-users/pull/674 and https://github.com/folio-org/ui-circulation/pull/184

PR #678, opened 2019-01-18
 Use new ModalFooter syntax

cherewaty:
Adopts patterns introduced in https://github.com/folio-org/stripes-components/pull/779.

Removes deprecation warnings re: using the `primaryButton` and `secondaryButton` props.

PR #680, opened 2019-01-21
UIEH-605 Finish replacing Redux Form with React Final Form

cherewaty:
## Purpose
Finishes https://issues.folio.org/browse/UIEH-605
Sequel to https://github.com/folio-org/ui-eholdings/pull/671

**`ui-eholdings` now has zero dependency on `redux-form`.**

## Approach
- Just like in https://github.com/folio-org/ui-eholdings/pull/671, I was able to completely move `validate()` for resource coverage dates to be encapsulated in its accompanying component, so it no longer has to be exported and consumed by parent forms. 
- I de-duplicated the proxy and token fields I had temporarily forked.
- I found a lot more reliable `<NavigationModal>` behavior with the `when` logic `!pristine && !request.isPending && !request.isResolved`. I applied to that all forms, not just resources.

## Future work
- The package and resource forms both have some lingering code smells from when adding them to holdings was a more complicated UI. `allowFormToSubmit` in particular stands out as something that should just be rolled into the form state instead of the component state.
- I adjusted some embargo editing tests to reflect some logical differences in validation between `redux-form` and `react-final-form`. All the same use cases are covered; the text returned just changed for some cases. The validation there could probably be revisited and simplified.
- The embargo field component could also use a brush-up re: its `change` and `initial` props.
- The `coverageStatementDecorator` pattern is pretty cool, but I'd rather place it within the `<CoverageStatementFields>` component. Unsure if that's possible.
- The package radio button groups (visibility, allow KB to automatically add titles) could be turned into reusable components in `src/components/package/_fields/`.


cherewaty:
Commit added to use `final-form-focus` on all forms.

cherewaty:
Rebased and force-pushed to resolve conflicts with https://github.com/folio-org/ui-eholdings/pull/678.

src/routes/resource-edit.js
```diff
@@ -105,7 +105,7 @@ class ResourceEditRoute extends Component {
       });
       model.isSelected = values.isSelected;
       model.url = customUrl;
-      model.visibilityData.isHidden = isVisible === 'false';
+      model.visibilityData.isHidden = !isVisible;
```
Yurii-Danylenko: In this file, there is a  mutation of the `modal` prop. It's not a React way of changing data. I know, Jeffrey, it's not your changes, but maybe it's a point to be discussed and reworked.

cherewaty: Commit added that improves that. ✨

src/routes/resource-edit.js
```diff
@@ -77,24 +77,24 @@ class ResourceEditRoute extends Component {
...
-      model.identifiersList = [];
-      model.identifiers = [];
-      model.customStatement = '';
-      model.customEmbargoPeriod = {};
-      model.contributors = [];
-      model.coverageStatement = '';
-      model.proxy = {};
-
-      updateResource(model);
+      updateResource(Object.assign(model, {
```
Godlevskyi: I believe it is the same as it was before as we will mutate the model.
You can pass the first argument (target) of object.assign as an empty object and then it will not be mutation

PR #685, opened 2019-01-23
Expand use of new <Pane>

cherewaty:
Expands use of new `<Pane>` introduced in https://github.com/folio-org/ui-eholdings/pull/634:
- added to knowledge base and root proxy forms
- added to eHoldings settings navigation
- added to custom package and custom title creation forms (those previously didn't have a fixed-position header)

### DIFF
--- a/src/components/paneset/Pane.js
+++ b/src/components/paneset/Pane.js
@@ -14,11 +14,6 @@ export default class Pane extends Component {
     // eslint-disable-next-line react/forbid-foreign-prop-types
     ...PaneHeader.propTypes,
 
-    // used for panes that animate from the opposite direction and are
-    // only tangentially related to the main content (such as
-    // navigation in settings, or filters in search)
-    aside: PropTypes.bool,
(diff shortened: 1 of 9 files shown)


src/components/paneset/Pane.js
```diff
@@ -93,19 +100,21 @@ export default class Pane extends Component {
...
+      tagName,
       visible,
       ...rest
     } = this.props;
     const {
       entered
     } = this.state;
 
-    let Element = aside ? 'aside' : 'section';
+    let Element = aside ? 'aside' : tagName;
```
Yurii-Danylenko: I think, `aside` prop is like a fifth wheel :) It's possible to assign `aside` tag through tagName, too. Is it possible to get rid of `aside` prop?

cherewaty: Commit added that cleans that up a little. I dropped the `aside` prop and just use `tagName="aside"` instead.

PR #686, opened 2019-01-23
Adjust search results heights for smaller base font size

cherewaty:
### Before
localhost_3000_eholdings_searchtype packages q t iphone 5_se 1

### After
localhost_3000_eholdings_searchtype packages q t iphone 5_se

### DIFF
--- a/src/components/package-search-list.js
+++ b/src/components/package-search-list.js
@@ -5,6 +5,8 @@ import { FormattedMessage } from 'react-intl';
 import QueryList from './query-list';
 import PackageListItem from './package-list-item';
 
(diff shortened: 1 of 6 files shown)


src/components/package-search-list.js
```diff
@@ -21,7 +21,7 @@ function PackageSearchList({
       fetch={fetch}
       collection={collection}
       onUpdateOffset={onUpdateOffset}
-      itemHeight={80}
+      itemHeight={70}
```
Godlevskyi: Can we move the hardcoded heights to constants?

cherewaty: Sure, commit added.

PR #737, opened 2019-02-15
Passing <AppIcon> to 'appIcon'-prop on <Pane>

rasmuswoelk:
The implementation of AppIcon's in Pane/PaneHeader's has changed in stripes-components.

In this PR I changed the search results component to reflect that change.


MaksymDryha:
Done in #820

PR #740, opened 2019-02-20
UIEH-627: fix form fields reverting to previous values on form submit

MaksymDryha:
## Purpose
Currently, when we edit a title or a package and hit save button, the form fields briefly revert back to their previous values which are no longer relevant

## Approach
Previously, the forms' obtained their initial values from props and rendered it. They always rendered what props provided them with. The problem is that props don't usually contain relevant data.
Now, the form values, coming with props, are first saved to the component's local state and the state is used as the source of data to render. This way can decide if we want to update the state with the props' data on the basis of the props' data relevance 

[Jira ticket](https://issues.folio.org/browse/UIEH-627)

### DIFF
--- a/src/components/package/edit-custom/custom-package-edit.js
+++ b/src/components/package/edit-custom/custom-package-edit.js
@@ -37,7 +37,6 @@ const focusOnErrors = createFocusDecorator();
 export default class CustomPackageEdit extends Component {
   static propTypes = {
     addPackageToHoldings: PropTypes.func.isRequired,
-    initialValues: PropTypes.object.isRequired,
     model: PropTypes.object.isRequired,
     onCancel: PropTypes.func.isRequired,
     onFullView: PropTypes.func,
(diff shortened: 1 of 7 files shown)


src/components/package/edit-managed/managed-package-edit.js
```diff
@@ -44,36 +43,73 @@ export default class ManagedPackageEdit extends Component {
...
+      provider,
+    } = nextProps;
+
+    const providerTokenWasLoaded = !initialValues.providerTokenValue && provider.providerToken.value;
+    const selectionStatusChanged = model.isSelected !== initialValues.isSelected;
+
+    if (selectionStatusChanged || providerTokenWasLoaded) {
+      stateUpdates = {
+        initialValues: {
+          isSelected: model.isSelected,
```
Yurii-Danylenko: It will be a little bit cleaner to extract first level props from model

src/components/package/edit-managed/managed-package-edit.js
```diff
@@ -44,36 +43,73 @@ export default class ManagedPackageEdit extends Component {
...
-    }
+  getInitialValuesFromModel() {
+    const {
+      model,
+      provider,
+    } = this.props;
 
-    return stateUpdates;
+    return {
+      isSelected: model.isSelected,
```
Yurii-Danylenko: It will be a little bit cleaner to extract first level props from model

src/components/package/edit-custom/custom-package-edit.js
```diff
@@ -46,12 +45,51 @@ export default class CustomPackageEdit extends Component {
...
+      proxy,
+      visibilityData,
+      destroy,
+    } = nextProps.model;
+
+    const selectionStatusChanged = isSelected !== initialValues.isSelected;
+
+    if (selectionStatusChanged) {
+      stateUpdates = {
+        initialValues: {
```
Godlevskyi: you can use your  getInitialValuesFromModel() 

Just add the model as an argument

src/components/package/edit-managed/managed-package-edit.js
```diff
@@ -44,36 +43,88 @@ export default class ManagedPackageEdit extends Component {
...
+      },
+      provider: { providerToken },
+    } = nextProps;
+
+    const providerTokenWasLoaded = !initialValues.providerTokenValue && providerToken.value;
+    const selectionStatusChanged = isSelected !== initialValues.isSelected;
+
+    if (selectionStatusChanged || providerTokenWasLoaded) {
+      stateUpdates = {
+        initialValues: {
```
Godlevskyi: same here

PR #743, opened 2019-02-26
UIEH-637: fix bug on closing unsaved form

MaksymDryha:
## Purpose
Currently, we have a bug on closing unsaved form, which is described below:
1. Search for a package
2. Edit it and save (if you clicked close button after changing some form fields, a modal notifying about unsaved changes would appear)
3. Start editing this package again, change some fields of the form, but this time, instead of saving the changes, click close button
4. See that no confirmation modal appeared, even though we have unsaved changes

Same issue can be reproduced for titles and providers. 
The problem was caused by the update request object, which contained not relevant information

## Approach

1. Create `REMOVE_REQUEST` action which takes the type of request and then is used to remove requests of that type 
2. Use this action in title/package/provider edit components to delete the update requests after the title/package/provider is updated

### DIFF
--- a/src/redux/model.js
+++ b/src/redux/model.js
@@ -8,7 +8,7 @@ import {
   unload,
   destroy,
   create,
-  removeCreateRequests,
+  removeRequests,
 } from './data';
 
 /**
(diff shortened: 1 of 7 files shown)


src/redux/model.js
```diff
@@ -283,8 +283,8 @@ class BaseModel {
   /**
    * Action creator for removing create requests for specific model's resource
```
Godlevskyi: Please correct the comment

PR #751, opened 2019-03-01
UIEH-586: Assign tags to package record

Godlevskyi:
## Purpose
Purpose: To display, assign, unassign, and output tags associated with individual Package records within the FOLIO eholdings app.

As a staff person
I want to be able to assign, unassign, view, and output tags in a FOLIO eholdings app's package record
So that I can categorize records for filtering and reporting purposes.

Background
Tags are meant to be labels that can be assigned to individual data records in various FOLIO apps. Tags will be used as a visual indicator in the record, to gather records in some way, and to facilitate reporting. This is a cross-app “helper app.” Github repo: https://github.com/folio-org/ui-tags

## Approach
tags functionality in the eholdings app was developed separately from the tags implemented in the stripes-smart-components as in eholdings we not used the search and sort component that integrates the tags.

In order to add the tags the tags the new Tags model was added.
Tags to populate the multi-select dropdown we receive from the direct api call to the mod-tags.

In case the new custom tag is added the tag is also added to the mod-tags.

### DIFF
--- a/src/components/package/show/package-show.js
+++ b/src/components/package/show/package-show.js
@@ -2,6 +2,7 @@ import React, { Component, Fragment } from 'react';
 import PropTypes from 'prop-types';
 import update from 'lodash/fp/update';
 import set from 'lodash/fp/set';
+import get from 'lodash/get';
 
 import {
   Accordion,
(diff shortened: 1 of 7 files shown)


src/components/package/show/package-show.js
```diff
@@ -267,6 +286,33 @@ class PackageShow extends Component {
...
+                  onToggle={this.handleSectionToggle}
+                  displayWhenClosed={
+                    <Badge sixe='small'>
+                      <span data-test-eholdings-provider-tags-bage>
+                        <FormattedNumber value={this.getEntityTags().length} />
+                      </span>
+                    </Badge>
+                  }
+                >
+                  {(!tagsModel.request.isResolved || model.isLoading)
```
MaksymDryha: I think these parenthesis are not needed

PR #752, opened 2019-03-04
UIEH-656: fix bug on adding a contributor with default type

MaksymDryha:
## Purpose
Currently, when we edit a title, click `Add new contributor` and don't change the type of the contributor, the default displayed type (which is `Author`) won't be sent to the back-end. The `Author` value is displayed in the select, but it isn't actually a part of the form's state.

## Approach
I added `addNewField` method to `contributor-field` component, which is responsible for setting the value of a just created select to the form's state. I also adjusted related tests and tests related to the bug described below.

This PR also fixes another bug which I found. It can be reproduced as follows: 
1. Go to title edit
2. Add some contributors of different type to the title. Save edited title
3. Go to editing the title again
4. Notice that all contributor type selects display 'Author' option, even though you set the them to some other types

The reason for the bug was just a misspelling

### DIFF
--- a/src/components/title/_fields/contributor/contributor-field.js
+++ b/src/components/title/_fields/contributor/contributor-field.js
@@ -71,7 +71,7 @@ export default class ContributorField extends Component {
   renderContributorTypeOptions() {
     return contributorTypeOptions.map(({ value, translationIdEnding }) => (
       <FormattedMessage id={`ui-eholdings.label.${translationIdEnding}`}>
-        {(message) => <option value={value}>{message}</option>}
+        {message => <option value={value}>{message}</option>}
       </FormattedMessage>
     ));
   }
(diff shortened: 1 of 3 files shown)


src/components/title/_fields/contributor/contributor-field.js
```diff
@@ -12,6 +12,21 @@ import {
 import { FormattedMessage } from 'react-intl';
 import styles from './contributor-field.css';
 
+const contributorTypeOptions = [
+  {
+    value: 'Author',
```
Godlevskyi: Please move hardcoded strings to constants

src/components/title/_fields/contributor/contributor-field.js
```diff
@@ -27,74 +42,101 @@ export default class ContributorField extends Component {
...
+              ? <FormattedMessage id="ui-eholdings.title.contributor.notSet" />
+              : '';
+
+            return (
+              <RepeatableField
+                addLabel={addLabel}
+                emptyMessage={emptyMessage}
+                fields={fields}
+                legend={legend}
+                onAdd={() => this.addNewField(fields)}
```
Yurii-Danylenko: Please, do not create new functions in render

PR #761, opened 2019-03-11
UIEH-669: Make UI permissions effective 

MaksymDryha:
## Purpose
We want to not let a user do an action if they don't have the permission to do the action.
## Approach
`IfPermission` component from `stripes-core` was used to render content depending on current user's permissions. I also had to inject stripes into `ResourceShow` component to use stripes's `hasPerm` function. This is needed for conditional rendering.

View permission was deleted from `package.json`, as it was decided to not use it at all. 

One more thing which I've done is removing the selection permission from the edit's sub-permissions (it's not needed there)

### DIFF
--- a/src/components/package/show/package-show.js
+++ b/src/components/package/show/package-show.js
@@ -11,6 +11,7 @@ import set from 'lodash/fp/set';
 import hasIn from 'lodash/fp/hasIn';
 
 import {
+  withStripes,
   Pluggable,
   IfPermission,
 } from '@folio/stripes-core';
@@ -63,6 +64,9 @@ class PackageShow extends Component {
     provider: PropTypes.object.isRequired,
     proxyTypes: PropTypes.object.isRequired,
     searchModal: PropTypes.node,
+    stripes: PropTypes.shape({
+      hasPerm: PropTypes.func.isRequired,
+    }).isRequired,
     tagsModel: PropTypes.object,
     toggleSelected: PropTypes.func.isRequired,
     updateEntityTags: PropTypes.func.isRequired,
@@ -134,6 +138,7 @@ class PackageShow extends Component {
 
   getActionMenu = ({ onToggle }) => {
     const {
+      stripes,
       onEdit,
       onFullView,
       model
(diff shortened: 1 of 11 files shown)


carolegodfrey:
@MaksymDryha -- this may be covered in another PR -- but just wanted to make sure that the  backend `kb-ebsco.all` is enabled through a front end permission. This could be done by adding `kb-ebsco.all` as a subPermission to `module.eholdings.enabled`

src/components/title/show/title-show.js
```diff
@@ -99,22 +109,24 @@ class TitleShow extends Component {
...
-              ariaLabel={ariaLabel}
-              onClick={onEdit}
-            />
-          )}
-        </FormattedMessage>
+        <IfPermission perm="ui-eholdings.records.edit">
+          <FormattedMessage
+            id="ui-eholdings.title.editCustomTitle"
+            values={{
+              id: '',
```
Yurii-Danylenko: Please, check if it is necessary to pass an empty string for id.

src/components/package/show/package-show.js
```diff
@@ -126,57 +134,113 @@ class PackageShow extends Component {
...
       <Fragment>
-        <Button
-          buttonStyle="dropdownItem fullWidth"
-          onClick={onEdit}
-        >
-          <FormattedMessage id="ui-eholdings.actionMenu.edit" />
-        </Button>
-
-        {onFullView && (
+        <IfPermission perm="ui-eholdings.records.edit">
```
Godlevskyi: Why did you chabnge the order of buttons in the dropdown? was it according to  acceptance criteria?

MaksymDryha: I didn't change it. I think git shows the diffs in a confusing way

MaksymDryha:
> @MaksymDryha -- this may be covered in another PR -- but just wanted to make sure that the backend `kb-ebsco.all` is enabled through a front end permission. This could be done by adding `kb-ebsco.all` as a subPermission to `module.eholdings.enabled`

Done! Thank you @carolegodfrey !

PR #771, opened 2019-03-15
Agreements for resources

Yurii-Danylenko:
## Purpose
Fix agreement attach request and extract agreements component

## TODO
Learn more about testing of RxJS epics
https://issues.folio.org/browse/UIEH-672


src/components/package/show/package-show.js
```diff
@@ -517,18 +481,12 @@ class PackageShow extends Component {
           }
         </Accordion>
 
-        <Accordion
+        <AgreementsSection
```
Godlevskyi: The component is called Accordion section, but it is accordion. please rename to AgreementsAccordion

PR #775, opened 2019-03-20
UIEH-641: Added the ability to assign/unassign tags for providers wit…

Godlevskyi:
## Purpose

The purposes of this PR is to unblock assigning/un Tags on the provider records when no selected packages.


src/components/provider/show/provider-show.js
```diff
@@ -227,36 +227,32 @@ class ProviderShow extends Component {
...
-          )
-        }
+              <FormattedMessage id="ui-eholdings.tags" />
+            </Headline>
+          )}
+          open={sections.providerShowTags}
+          id="providerShowTags"
+          onToggle={this.handleSectionToggle}
+          displayWhenClosed={
+            <Badge sixe='small'>
```
MaksymDryha: Probably it's "size"?

PR #779, opened 2019-03-21
Make prefer-const rule effective

MaksymDryha:
In this small PR, I made the `prefer-const` rule effective again. All variable declarations which violated the rule were automatically changed to use `const` keyword. `eslint --fix` was used for it.


.eslintrc.js
```diff
@@ -23,7 +23,6 @@ module.exports = {
     "import/no-extraneous-dependencies": "off",
     "jsx-quotes": "off",
     "no-console": "warn",
-    "prefer-const": "off",
```
MaksymDryha: Here's the key change

PR #780, opened 2019-03-22
UIEH-677: fix code smells

MaksymDryha:
In this PR, I fixed SonarCloud code smells. I also fixed some spelling mistakes which I was able to find


src/components/scroll-view/scroll-view.js
```diff
@@ -146,7 +146,7 @@ export default class ScrollView extends Component {
   render() {
     // strip all other props to pass along the rest to the div
     // eslint-disable-next-line no-unused-vars
-    let { items, length, itemHeight, offset: _, onUpdate, scrollable, fullWidth, ...props } = this.props;
+    let { items, length, itemHeight, onUpdate, scrollable, fullWidth, ...props } = this.props;
```
Godlevskyi: new lines, const

src/redux/data.js
```diff
@@ -614,7 +604,7 @@ const parseResponseBody = (response) => {
  * @param {Object} state - data store state leaf
  * @param {Object} action - redux action being dispatched
  */
-export function reducer(state = {}, action) {
+export function reducer(state, action) {
```
Yurii-Danylenko: Please, define initial value for the state `{}` inside function body

MaksymDryha: @Yurii-Danylenko, the default value for state is already set and passed to this function (it's called `dataReducer`) [here](https://github.com/folio-org/ui-eholdings/blob/master/src/redux/index.js#L46).

PR #784, opened 2019-03-29
UIEH-629, UIEH-631: Add New button to package and resource show pages…

Yurii-Danylenko:
UIEH-629, UIEH-631

## Purpose
Provide a possibility to create new agreement from e-holdings app

## Approach
- Add "New" button to package and resource show pages
- Add tests

### DIFF
--- a/test/bigtest/tests/package-show-test.js
+++ b/test/bigtest/tests/package-show-test.js
@@ -3,12 +3,13 @@ import { expect } from 'chai';
 
 import setupApplication from '../helpers/setup-application';
 import PackageShowPage from '../interactors/package-show';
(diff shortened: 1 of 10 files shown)


test/bigtest/tests/package-show-test.js
```diff
@@ -3,12 +3,13 @@ import { expect } from 'chai';
...
 import PackageShowPage from '../interactors/package-show';
+import { entityAuthorityTypes } from '../../../src/constants';
 
 describe('PackageShow', () => {
   setupApplication();
-  let provider,
-    providerPackage,
-    resources;
+  let provider;
+  let providerPackage;
```
Godlevskyi: As I remember we decided to use const

PR #789, opened 2019-04-05
UIEH-651: Fixed translation for capitalized strings

Godlevskyi:
## Purpose

According to the UIEH-651, 
lodash's capitalize function is used throughout the app to display type-indicators such as "titles", "packages", or "providers" as "Titles" etc. This usage conflates a type (as used in code) from the type's label (as used for display) and, crucially, makes these values impossible to translate. All uses of _.capitalize() that render text to the screen must be replaced with <FormattedMessage> to allow for translation. Uses that render markup are probably permissible but should be investigated.


translations/ui-eholdings/en.json
```diff
@@ -326,5 +331,10 @@
     "hidden": "Hidden",
     "name.isRequired": "Name *",
     "resultCount": "{count, number} {count, plural, one {search result} other {search results}}",
-    "over": "Over"
+    "over": "Over",
+    "contributorType.Author": "{count, plural, one {Author} other {Authors}}",
```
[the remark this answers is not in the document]
Godlevskyi: Good catch!

src/components/search-form/search-form.js
```diff
@@ -75,11 +79,11 @@ class SearchForm extends Component {
...
   getFiltersComponent = (searchType) => {
-    if (searchType === 'titles') {
+    if (searchType === searchTypes.TITLES) {
       return TitleSearchFilters;
-    } else if (searchType === 'packages') {
+    } else if (searchType === searchTypes.PACKAGES) {
       return PackageSearchFilters;
-    } else if (searchType === 'providers') {
+    } else if (searchType === searchTypes.PROVIDERS) {
       return ProviderSearchFilters;
```
MaksymDryha: Maybe it would be cleaner to make it like this:
```javascript
const filterComponents = {
  [searchTypes.TITLES]: TitleSearchFilters,
  [searchTypes.PACKAGES]: PackageSearchFilters,
  [searchTypes.PROVIDERS]: ProviderSearchFilters,
}
```

and then we could write 
```js
const filterComponent = filterComponents[searchType] || null;
```

And I think that probably `null` is not needed there which means it can be even more clean:
```js
const filterComponent = filterComponents[searchType];
```

src/components/details-view/accordion-list-header.js
```diff
@@ -3,12 +3,13 @@ import PropTypes from 'prop-types';
...
 import { DefaultAccordionHeader, KeyValue } from '@folio/stripes/components';
 import styles from './accordion-list-header.css';
+import { listTypes } from '../../constants';
 
 function AccordionListHeader(props) {
   // RM API does not return exact number of results when count is over 10K
   // For title lists, resultsLength of 10000 indicates this.
   // For other lists (package and provider) resultsLength of 10001 indicates this.
-  let overCount = props.listType === 'titles' ? 10000 : 10001;
+  let overCount = props.listType === listTypes.TITLES ? 10000 : 10001;
```
MaksymDryha: please, format it

PR #790, opened 2019-04-08
UIEH-684: Fix agreements attach functionality

Yurii-Danylenko:
https://issues.folio.org/browse/UIEH-684

## Purpose
Fix attach agreement functionality


src/features/agreements-accordion/model/agreement.js
```diff
@@ -1,11 +1,14 @@
...
-  authority = 'EKB';
+  authority = '';
   reference = '';
   label = '';
 
-  constructor({ reference, label }) {
-    this.reference = reference;
+  constructor({ id, refId, authorityType, label }) {
+    this.id = id;
+    this.reference = refId;
```
MaksymDryha: You can do it outside of the constructor. Like this:
```js
export default class Agreement {
  id = this.props.id;
  type = 'external';
  authority = this.props.authorityType;

```

PR #801, opened 2019-05-21
UIEH-685 Title: Remove Tags functionality

carolegodfrey:
## Purpose
Per https://issues.folio.org/browse/UIEH-685 we are removing tags from title records.

Tags will be supported on resource records (aka title+package records). When performing a filter by tags from Search&Filter for Titles, we will lookup tags associated with resources to support this.
https://issues.folio.org/browse/UIEH-701

## Approach
- Remove tags accordion and associated methods from Title Show Page Component
- Remove tag methods and model information from Title Show Route
- Remove/Update related tests
- Update Redux title (removing taglist)

#### TODOS and Open Questions
- [ ] Decided to remove tags from Title Redux layer (no longer used by UI although still available in the back end)

## Screenshots

Before
Screen Shot 2019-05-21 at 4 46 49 PM

After
Screen Shot 2019-05-21 at 4 45 15 PM

### DIFF
--- a/src/components/title/show/title-show.js
+++ b/src/components/title/show/title-show.js
@@ -32,7 +32,6 @@ import IdentifiersList from '../../identifiers-list';
 import ContributorsList from '../../contributors-list';
 import AddTitleToPackage from '../_field-groups/add-title-to-package';
 import Toaster from '../../toaster';
-import TagsAccordion from '../../tags';
 import KeyValueColumns from '../../key-value-columns';
 import styles from './title-show.css';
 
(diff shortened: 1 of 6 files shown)


carolegodfrey:
> LGTM @carolegodfrey. Do you think we should create a story to cleanup in backend as well? I know there is quite a bit to deal with tags at title level. Unsure if Khalilah will want this back again at some point.

Good point -- probably should add a placeholder story for this removal

src/components/title/show/title-show.js
```diff
@@ -218,15 +211,6 @@ class TitleShow extends Component {
           lastMenu={this.lastMenu}
           bodyContent={(
             <Fragment>
-              <TagsAccordion
```
MaksymDryha: I guess the `<Fragment>` is no longer needed since the `<TagsAccordion />` was removed

carolegodfrey: Thanks @MaksymDryha -- removed it!

PR #803, opened 2019-05-29
Note view

Godlevskyi:
## Purpose
In this PR we add the note view route.


src/routes/note-edit.js
```diff
@@ -59,7 +57,7 @@ export default class NoteEditRoute extends Component {
 
     const { id } = match.params;
 
-    const referredEntityData = location.state && this.getReferredEntityData();
+    const referredEntityData = formatNoteReferrerEntityData(location.state);
```
MaksymDryha: I guess with this change, `getReferredEntityData` method should be removed

src/routes/note-view.js
```diff
@@ -0,0 +1,79 @@
...
+import {
+  entityTypeTranslationKeys,
+  entityTypePluralizedTranslationKeys,
+  APP_ICON_NAME,
+} from '../constants';
+
+class NoteViewRoute extends Component {
+  static propTypes = {
+    history: PropTypes.shape({
+      goBack: PropTypes.func.isRequired,
```
MaksymDryha: Why don't you use `ReactRouterPropTypes`?

PR #818, opened 2019-06-24
Test filtering by tags

Yurii-Danylenko:
https://issues.folio.org/browse/UIEH-730

## Purpose
Test filtering by tags resources inside a package and packages inside a provider.

### DIFF
--- a/test/bigtest/network/config.js
+++ b/test/bigtest/network/config.js
@@ -118,7 +118,7 @@ export default function config() {
     }]
   }]);
   // tags endpoint
-  this.get('/tags', {
+  this.get('/tags', ({
     tags: [
       {
         id: '1',
@@ -131,7 +131,7 @@ export default function config() {
         description: 'Requires not urgent attention',
       },
     ]
-  });
+  }));
 
   this.get('/erm/sas', [
     {
--- a/test/bigtest/tests/package-show-titles-search-test.js
+++ b/test/bigtest/tests/package-show-titles-search-test.js
@@ -43,10 +43,19 @@ describe('Package Show Title Search', () => {
       isSelected: true,
     });
 
+    resources[0].title.update({
+      tags: {
+        tagList: ['urgent', 'not urgent']
+      }
+    });
+
     resources[1].title.update({
       name: 'My Title 2',
       publicationType: 'book',
-      publisherName: 'The Frontside'
+      publisherName: 'The Frontside',
+      tags: {
+        tagList: ['urgent']
+      }
     });
 
     resources[1].update({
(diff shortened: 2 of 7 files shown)


test/bigtest/tests/package-show-titles-search-test.js
```diff
@@ -175,6 +187,91 @@ describe('Package Show Title Search', () => {
         });
       });
     });
+
+    describe('filtering resources by tags', () => {
+      beforeEach(async () => {
+        await PackageShowPage.clickListSearch();
+      });
+
+      it('shows the accordion with tags filter', () => {
```
MaksymDryha: since we discussed, what about
```
describe(when ...)
  describe(and...)
    describe(and...)
      it(should...)
```

test/bigtest/tests/package-show-titles-search-test.js
```diff
@@ -175,6 +187,91 @@ describe('Package Show Title Search', () => {
...
+        beforeEach(async () => {
+          await PackageShowPage.searchModal.tagsSection.clickTagHeader();
+        });
+
+        it('should expand the accordion', () => {
+          expect(PackageShowPage.searchModal.tagsSection.tagsAccordion.isOpen).to.be.true;
+        });
+      });
+
+      describe('after doing a couple of clicks on accordion header', () => {
```
MaksymDryha: since we discussed, what about
```
describe(when ...)
  describe(and...)
    describe(and...)
      it(should...)
```

test/bigtest/network/config.js
```diff
@@ -190,7 +190,7 @@ export default function config() {
     }]
   }]);
   // tags endpoint
-  this.get('/tags', {
+  this.get('/tags', ({
```
MaksymDryha: I would prefer to configure models, factories etc so that mirage can support tags. This way you won't have to hardcode tag names in tests. Also tags functionality would become available in mirage mode, now it's not

test/bigtest/tests/provider-show-package-search-test.js
```diff
@@ -92,6 +107,87 @@ describe('ProviderShow package search', () => {
...
+
+        it('should expand the accordion', () => {
+          expect(ProviderShowPage.searchModal.tagsSection.tagsAccordion.isOpen).to.be.true;
+        });
+      });
+
+      describe('after doing a couple of clicks on accordion header', () => {
+        beforeEach(async () => {
+          await ProviderShowPage.searchModal.tagsSection.clickTagHeader();
+          await ProviderShowPage.searchModal.tagsSection.clickTagHeader();
```
carolegodfrey: Just curious why a few clicks are needed - just behavior of interactor?

PR #825, opened 2019-06-26
UIEH-687: removed convertion to lower case of contributor type

Godlevskyi:
## Purpose
Per https://issues.folio.org/browse/UIEH-687 we we receive the contributor in lower case in the payload and no need to lowercase it.


test/bigtest/interactors/title-create.js
```diff
@@ -25,12 +26,12 @@ import {
...
-  addContributor(type, name) {
-    return this
-      .click('[data-test-eholdings-contributors-fields] [data-test-repeatable-field-add-item-button]')
-      .fill('[data-test-eholdings-contributor-type] select', type)
-      .fill('[data-test-eholdings-contributor-contributor] input', name);
-  }
+  clickAddContributor = clickable('[data-test-eholdings-contributors-fields] [data-test-repeatable-field-add-item-button]');
+  contributorsRowList = collection('[data-test-eholdings-contributors-fields] li', {
+    type: fillable('[data-test-eholdings-contributor-type] select'),
+    contributor: fillable('[data-test-eholdings-contributor-contributor] input'),
```
MaksymDryha: maybe it's better to name them like enterType and enterContributor?

src/components/contributors-list.js
```diff
@@ -14,11 +14,10 @@ export default function ContributorsList({ data }) {
   return (
     <div>
       {Object.keys(contributorsByType).map((contibutorType) => {
-        const contributorTypeTranslation = contibutorType.toLowerCase();
         const names = contributorsByType[contibutorType];
         const label = (
           <FormattedMessage
-            id={`ui-eholdings.contributorType.${contributorTypeTranslation}`}
+            id={`ui-eholdings.contributorType.${contibutorType}`}
```
sduvvuri-ebsco: a typo on "contributorType"

test/bigtest/interactors/title-create.js
sduvvuri-ebsco: data-test-eholdings-contributor-contributor seems redundant but I can't think of a better name to suggest

PR #827, opened 2019-06-27
ERM-253: Update Agreements URLs

doytch:
## Purpose
The URLs to view and create agreements have changed with ERM-253 implemented in https://github.com/folio-org/ui-agreements/pull/219 This PR updates eHoldings to use those new URLs.

Feel free to push to this branch if I missed anything.

Also, there's [big underlying changes to the find-agreement plugin](https://github.com/folio-org/ui-plugin-find-agreement/pull/41), but it looks like eHoldings is completely decoupled from them since the public API has stayed the same and eHoldings doesn't declare a dependency on a particular version of ui-plugin-find-agreement (which makes sense).

### DIFF
--- a/src/components/agreements-list/agreements-list.js
+++ b/src/components/agreements-list/agreements-list.js
@@ -39,7 +39,7 @@ export default class AgreementsList extends React.Component {
       labelStrings,
     } = row;
 
-    const ermAgreementUrl = `/erm/agreements/view/${id}`;
+    const ermAgreementUrl = `/erm/agreements/${id}`;
     const ariaLabel = labelStrings && labelStrings.join('...');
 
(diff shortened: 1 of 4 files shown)


doytch:
Pretty sure those errors aren't related to this... 🤔

doytch:
@carolegodfrey @sduvvuri-ebsco , I'm not sure whether to keep rerunning these tests if they're just flaky or if something under the covers has broken. 😬

sduvvuri-ebsco:
Thanks @doytch . We are on it. Please do not waste your time. We are monitoring and its flakey tests. Each time a different test fails. We will merge it once we get the tests to pass.

PR #828, opened 2019-06-28
Update jenkins file to not run tests

sduvvuri-ebsco:
There are few flaky tests in eHoldings and we want to get this build going. So disabling running them for now.


MaksymDryha:
> Sorry, but I think it's not a good idea to disable all tests because of few failed. What about skipping just flaky tests?

I

MaksymDryha:
Sorry, accidentally missclicked and closed the PR

Yurii-Danylenko:
As situation with tests in Stripes are not stable, it was decided to disable tests until solution will be found

PR #831, opened 2019-07-03
UIEH-737: make search filters collapsed by default

MaksymDryha:
The purpose of this PR is to make all search filters throughout the application closed by default.
All necessary tests adjustments were done to reflect this change.


MaksymDryha:
> Looks good @MaksymDryha -- only minor comment would be possibly adding tests that check accordion state is collapsed by default (although you toggle them open to access filters)

This makes sense, thank you! I added some more tests!

PR #834, opened 2019-07-10
rxjs

Yurii-Danylenko:
https://issues.folio.org/browse/UIEH-672

## Purpose
Cover epics tied to agreements with unit tests.

## Approach
- Extract API calls from epics tied to agreements.
- Bump up RxJS version to 6.5.2
- Bump up versions of RxJS dependencies(Redux, Redux-observable)
- Cover agreements epics with unit tests with a help of marble diagrams

## Learning
https://redux-observable.js.org/docs/recipes/WritingTests.html

### DIFF
--- a/src/api/common/parse-response-body.js
+++ b/src/api/common/parse-response-body.js
@@ -1,0 +1,5 @@
+export default function parseResponseBody(response) {
+  return response.text().then((text) => {
+    try { return JSON.parse(text); } catch (e) { return text; }
+  });
+}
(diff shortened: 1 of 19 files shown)


src/api/common/parse-response-body.js
```diff
@@ -0,0 +1,5 @@
+export default function parseResponseBody(response) {
+  return response.text().then((text) => {
+    try { return JSON.parse(text); } catch (e) { return text; }
```
Godlevskyi: Please remove non-used catch argument

PR #842, opened 2019-07-26
UIEH-712, UIEH-713: Disable filtering by tags if user enters a search term or selects a filter

MaksymDryha:
In this PR, I implemented toggling between searching by search term and searching by tags so that it's impossible to combine these types of search.

### DIFF
--- a/src/routes/search.js
+++ b/src/routes/search.js
@@ -244,15 +244,8 @@ class SearchRoute extends Component {
    * Update the url to match new search params
    * @param {Object} params - query param object
    */
-  updateURLParams = () => {
+  updateURLParams = (params) => {
     const { location, history } = this.props;
-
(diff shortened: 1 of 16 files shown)


src/routes/search.js
```diff
@@ -101,17 +123,64 @@ class SearchRoute extends Component {
...
+    const {
+      searchByTagsEnabled,
+      draftSearchFilters,
+    } = this.state;
 
-  handleFilterChange = (sort, searchFilter) => {
-    this.setState({ sort, searchFilter }, this.handleSearch);
-  }
+    this.setState(currentState => ({
+      searchByTagsEnabled: !currentState.searchByTagsEnabled
```
Godlevskyi: Please move up the setState from callback

PR #846, opened 2019-08-05
UIEH-745, UIEH-747, UIEH-748: Update tags endpoints

Godlevskyi:
## Purpose
The purpose of this PR is to change the mechanism for the tags update.
Before the tags update was done updating the resource.
Currently tags updated separately from the resource and this significantly  improved performance. 
## Approach
Updating the tag we create "fake" update request for the resource and update only tags in the associated resource record in case of the success or update request with errors in case of error.


src/components/tags/tags.js
```diff
@@ -29,26 +29,49 @@ export default class Tags extends React.Component {
     this.saveTags(tags);
   }
 
-  onRemove(tag) {
+  formatTagUpdatePayload(model) {
+    const data = {
+      payload: {
```
Yurii-Danylenko: Please, get rid of unnecessary wrapper property "payload".

src/components/tags/tags.js
```diff
@@ -29,26 +29,49 @@ export default class Tags extends React.Component {
     this.saveTags(tags);
   }
 
-  onRemove(tag) {
+  formatTagUpdatePayload(model) {
+    const data = {
```
Yurii-Danylenko: Consider to rename it to `tagsData`

src/redux/actions/update-entity-tags.js
```diff
@@ -0,0 +1,14 @@
...
+export default function updateEntityTags(type, payload, path) {
+  return {
+    type: entityTagsActionTypes.UPDATE_ENTITY_TAGS,
+    data: {
+      type,
+      path,
+      params: { id: payload.id },
+      timestamp: Date.now()
+    },
+    payload: payload.payload
```
Yurii-Danylenko: Should be refactored after changes

src/redux/constants/entityTagsActionTypes.js
```diff
@@ -0,0 +1,8 @@
+const entityTagsActionTypes = {
+  // Entity tags
```
Yurii-Danylenko: Please, get rid of comment

PR #849, opened 2019-08-07
UIEH-770: Updated react-final-form and related dependencies

Godlevskyi:
## Purpose
1) react-final-form and related libraries was updated to the latest versions
2) Fixed validators in resource edit form
3) Fixed  coverage dated handling (set datepicker not to correct the date accorting to the timezone)
4) Fixed tests


test/bigtest/tests/settings-root-proxy-test.js
```diff
@@ -19,61 +20,56 @@ describe('With list of root proxies available to a customer', () => {
...
-      this.server.put('/root-proxy', {
+    beforeEach(async function () {
+      await this.server.put('/root-proxy', {
         errors: [{
           title: 'There was an error'
         }]
       }, 500);
 
-      this.visit('/settings/eholdings/root-proxy');
+      await this.visit('/settings/eholdings/root-proxy');
```
MaksymDryha: is `await` needed? I don't think `visit` returns a promise

PR #868, opened 2019-09-13
Release v2.0.1

Yurii-Danylenko:
## Purpose
Regenerate yarn.lock with pointing the packages to npm-folio registry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change history for ui-eholdings
 
+## [2.0.1](https://github.com/folio-org/ui-eholdings/tree/v2.0.1) (2019-09-16)
+[Full Changelog](https://github.com/folio-org/ui-eholdings/compare/v2.0.0...v2.0.1)
+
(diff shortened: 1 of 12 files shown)


zburke:
FYI, typically the only changes in a release PR are in `package.json` (bumping the version) and CHANGELOG.md (cleaning up the list changes and adding the release details headline). That said, there is nothing technically wrong with bundling additional changes into a release PR. 

As to the stated purpose of this PR, "Regenerate yarn.lock with pointing the packages to npm-folio registry", in fact the {{yarn.lock}} file should simply be removed. This module will be built in the context of a platform, and thus its dependencies resolved in the context of that platform. This means the {{yarn.lock}} here will be ignored. 

The only context in which the local {{yarn.lock}} for a UI app or stripes library is relevant is when building it in isolation for the purposes of unit testing. You could argue that this is a good reason to keep the file around, but given the other expectations people have about {{yarn.lock}} which do not apply here, it provides something of a false sense of security.

PR #875, opened 2019-09-25
UIEH-792: Remove yarn.lock from the repository

Godlevskyi:
## Purpose
The Frontside historically pushed the yarn.lock to the repo so that on CI new yarn.lock file is not generated. Across all the FOLIO apps such approach is not used.
Taking into account that eholdings now depends more on the other modules it make sense not to push the yarn.lock file

So the yarn.lock removed from the repository and added to the gitignore file so that the registry is switched on CI depending on the build.


Godlevskyi:
Yarn lock is deleted within the release v2.0.2

PR #880, opened 2019-10-01
Support erm interface v2.0

doytch:
## Purpose
`mod-agreements` needs to update the version on the `erm` interface to 2.0 because of the breaking changes to the API to [_set_ agreement dates/periods](https://issues.folio.org/browse/ERM-466). eHoldings needs to support the new interface version for the nightly builds to build.

## Approach
The breaking change is only to the process of editing the start, end or cancellation date of an agreement. eHoldings does none of these so a simple bump to the version in `okapiInterfaces` is all that's needed.

Note that as of this PR being opened, `mod-agreements` has not yet been updated with the new interface version. I'm opening this PR ahead of time so that this can be merged before then.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "okapiInterfaces": {
       "eholdings": "2.0",
       "tags": "1.0",
-      "erm": "1.0"
+      "erm": "1.0 2.0"
     },
     "permissionSets": [
       {


doytch:
@Godlevskyi, it looks like the failed CI checks are unrelated to this, can you confirm if you think that's the case? If so, can you go ahead and merge this if you've got that power? https://github.com/folio-org/mod-agreements/pull/185 has been merged so this will need merging today.

Godlevskyi:
> @Godlevskyi, it looks like the failed CI checks are unrelated to this, can you confirm if you think that's the case? If so, can you go ahead and merge this if you've got that power? [folio-org/mod-agreements#185](https://github.com/folio-org/mod-agreements/pull/185) has been merged so this will need merging today.

@doytch , thanks for merging. Tests errors are not related to your PR. We believe that the problem with it that we have too many tests and the CI environment does not have enough capacity. Re-running the tests usually helps))

PR #935, opened 2019-12-26
UIEH-805: Display custom labels on view resource page

Vladyslav-Velytskyi:
## Purpose

- Create a Custom Labels Accordion
- Implement GET query to backend for Custom Labels
- Add dash value for empty fields with screen readers support

## Link
https://issues.folio.org/browse/UIEH-805

## Screenshots
image

## Test Coverage
image


src/components/utilities.js
```diff
@@ -112,6 +113,18 @@ export const getTagLabelsArr = (tagsModel) => {
   return Object.values(tagRecords).map((tag) => tag.attributes);
 };
 
+export const getCustomLabelsArr = (customLabelsModel) => {
+  const customLabelsRecords = get(customLabelsModel, ['resolver', 'state', 'customLabel', 'records'], {});
```
MaksymDryha: Shouldn't the default value be `[]` instead of `{}`?

src/constants/dashValue.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const dashValue = (
+  <span
+    /* eslint-disable-next-line jsx-a11y/aria-role */
+    role="text"
+    aria-label="no value set"
```
Godlevskyi: 1) Aria-label should be internationalized. 
2) Please capitalize it when added to translations en.json (No value set.)

Godlevskyi: Also please ask in the stripes channel whether there is already a "Folio way" pattern/component for such no value stuff. As if the label will be "No value set." It may make sense to add it to the stripes components to have internationalization strings in one plase

src/components/utilities.js
Vladyslav-Velytskyi: No, `records`  returns an object anyway, in the next string of this function we get values of this object
` return Object.values(customLabelsRecords).map(customLabel => customLabel.attributes);`

src/components/utilities.js
```diff
@@ -112,6 +113,18 @@ export const getTagLabelsArr = (tagsModel) => {
...
+
+  return Object.values(customLabelsRecords).map(customLabel => customLabel.attributes);
+};
+
+export const getUserDefinedFields = (model) => {
+  const attributes = get(model, ['data', 'attributes']);
+
+  return pick(attributes, ['userDefinedField1', 'userDefinedField2', 'userDefinedField3', 'userDefinedField4', 'userDefinedField5']);
```
Godlevskyi: Is there a limitation to have only 5 custom labels on the resource?

src/redux/custom-label.js
```diff
@@ -0,0 +1,29 @@
...
+  displayOnPublicationFinder = false;
+
+  serialize() {
+    const data = {
+      type: this.type,
+      attributes: {},
+    };
+
+    data.attributes = {
+      id: this.id,
```
MaksymDryha: I don't think `id` should be inside `data.attributes`. It should be placed next to `type` and `attributes` fields of `data`

Vladyslav-Velytskyi: I'm not sure about it, according to API documentation `id` present just in attributes and does not exist in `data`

doc: https://s3.amazonaws.com/foliodocs/api/mod-kb-ebsco-java/p/custom-labels.html#eholdings_custom_labels_get

src/components/resource/resource-show.js
```diff
@@ -213,6 +214,9 @@ class ResourceShow extends Component {
       sections
     } = this.state;
 
+    const userDefinedFields = getUserDefinedFields(model);
+    const isCustomLabelsViewed = model.isTitleCustom || (model.isSelected && !model.isTitleCustom);
```
Godlevskyi: 1) Please rename it to showCustomLables or similar as as viewed means that someone already saw it
2) I believe model.isTitleCustom || (model.isSelected && !model.isTitleCustom) could be rewritten to 
model.isTitleCustom || model.isSelected

src/constants/dashValue.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+const dashValue = (
```
Godlevskyi: please rename it to noValue

src/redux/reducers/customLabels.js
```diff
@@ -0,0 +1,48 @@
...
+  },
+  [GET_CUSTOM_LABELS_FAILURE]: (state, action) => {
+    const {
+      payload: { errors },
+    } = action;
+
+    return {
+      ...state,
+      isLoading: false,
+      errors: formatErrors(errors),
```
Godlevskyi: Please check the format of errors, returned from the cutomLables endpoint to be sure that it is properly hanled by  formatErrors func

src/api/custom-labels.js
```diff
@@ -0,0 +1,40 @@
...
+import 'rxjs/add/observable/throw';
+
+import {
+  parseResponseBody,
+  getHeaders,
+} from './common';
+
+const API_URL = '/eholdings/custom-labels';
+
+const createUrl = (baseUrl) => {
```
MaksymDryha: We have a function with the same purpose in `api/agreements.js`. I would suggest to generilize it and move to `api/common`

MaksymDryha: We can try using currying here

src/redux/epics/get-custom-labels.js
```diff
@@ -0,0 +1,25 @@
...
+  } = store;
+
+  const state = getState();
+
+  return action$
+    .filter(action => action.type === GET_CUSTOM_LABELS)
+    .mergeMap(() => customLabelsApi
+      .getAll(state.okapi)
+      .map(response => getCustomLabelsSuccess(response))
+      .catch(errors => Observable.of(getCustomLabelsFailure({ errors }))));
```
MaksymDryha: I assume it's possible to write it like this:
```js
.mergeMap(() => customLabelsApi.getAll(state.okapi))
.map(response => getCustomLabelsSuccess(response))
.catch(errors => Observable.of(getCustomLabelsFailure({ errors })));
```

PR #964, opened 2020-02-19
UIEH-828: Create Settings Access status types page

Vladyslav-Velytskyi:
## Purpose
- Create Settings Access status types page
- Create API connection for access types
- Validation implementation on UI side
- New code test coverage

## Link
https://issues.folio.org/browse/UIEH-828

## Screenshots
image

## Test Coverage
image


src/components/settings/settings-access-status-types/settings-access-status-types.js
```diff
@@ -0,0 +1,163 @@
...
+      />;
+    } else {
+      return null;
+    }
+  };
+
+  return (
+    <Pane
+      paneTitle={<FormattedMessage id="ui-eholdings.settings.accessStatusTypes" />}
+      data-test-settings-access-status-types
```
annamelnyk: Please, try to put `data-` attributes in the top on the first place

src/components/settings/settings-access-status-types/settings-access-status-types.js
```diff
@@ -0,0 +1,163 @@
...
+              to="/settings/eholdings"
+            />
+          )}
+        </FormattedMessage>
+      )}
+    >
+      <IntlConsumer>
+        {intl => (
+          <EditableList
+            actionProps={accessTypesData && accessTypesData.length >= 15 ? { create: () => ({ disabled: true }) } : {}}
```
annamelnyk: I would recommend you to bring current expression into separate function `getActionProps` or smth like that.
And also would be better to use `isEmpty()` lodash method instead `accessTypesData.length`

src/redux/epics/attach-access-type.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default ({ accessTypesApi }) => (action$, store) => {
+  const { getState } = store;
+  const state = getState();
+
+  return action$
+    .filter(action => action.type === ATTACH_ACCESS_TYPE)
+    .mergeMap(action => {
+      const { payload: accessType } = action;
+
```
annamelnyk: Current variable could be destructuring  in an arguments above `({ payload: accessType }) => {...}`
And the same can be done in simular cases below, just FYI

src/redux/epics/attach-access-type.js
```diff
@@ -0,0 +1,30 @@
...
+  attachAccessTypeFailure,
+  addAccessType,
+} from '../actions';
+
+export default ({ accessTypesApi }) => (action$, store) => {
+  const { getState } = store;
+  const state = getState();
+
+  return action$
+    .filter(action => action.type === ATTACH_ACCESS_TYPE)
```
annamelnyk:
```suggestion
    .filter(({ type }) => type === ATTACH_ACCESS_TYPE)
```

src/components/settings/settings-access-status-types/settings-access-status-types.js
```diff
@@ -0,0 +1,163 @@
...
+      updatedByUserId: PropTypes.string,
+      updatedDate: PropTypes.string,
+    }),
+    type: PropTypes.string.isRequired,
+    updater: PropTypes.objectOf(PropTypes.string),
+  })),
+  onCreate: PropTypes.func.isRequired,
+  onDelete: PropTypes.func.isRequired,
+  onUpdate: PropTypes.func.isRequired,
+};
```
annamelnyk: If you have `props `which are not required, I would recommend to add them to `defaultProps`

PR #973, opened 2020-03-04
Bump to version 3 with stripes v3

Godlevskyi:
## Purpose
Bumped stripes to v3

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@folio/eholdings",
-  "version": "2.1.0",
+  "version": "3.0.0",
   "description": "FOLIO UI module for eHoldings",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,6 +38,7 @@
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-redux": "^5.0.7",
+    "react-router-dom": "^4.3.1",
```
BogdanDenis: I though that `react-router` also needed to be moved to `devDependencies`?

Godlevskyi: added it according to this doc: https://github.com/folio-org/stripes/blob/stripes-v3/doc/upgrade-v3.md

PR #990, opened 2020-03-24
UIEH-854: Edit Package Record | Proxy selection is not honored and Unsaved Changes modal does not display when you edit a record for the first time

BogdanDenis:
## Purpose

- Fix Unsaved Changes Modal not appearing after changing selected status of package.
- Fix need to click 'X' twice to return after changing selected status of resource and visiting Edit page.

## Issue

[UIEH-854](https://issues.folio.org/browse/UIEH-854)

## Coverage

image

### DIFF
--- a/src/components/package/edit-custom/custom-package-edit.js
+++ b/src/components/package/edit-custom/custom-package-edit.js
@@ -36,7 +36,7 @@ import styles from './custom-package-edit.css';
 const focusOnErrors = createFocusDecorator();
 
 class CustomPackageEdit extends Component {
-  static getInitialValues(model) {
+  static getInitialValues(model, proxyTypes) {
     const {
       name,
       contentType,
@@ -46,6 +46,9 @@ class CustomPackageEdit extends Component {
       visibilityData,
     } = model;
 
+    const proxyTypesRecords = proxyTypes.resolver.state.proxyTypes.records;
+    const matchingProxy = Object.values(proxyTypesRecords).find(proxyType => proxyType.id.toLowerCase() === proxy.id);
+
     return {
       name,
       contentType,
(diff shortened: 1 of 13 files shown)


src/components/package/edit-custom/custom-package-edit.js
```diff
@@ -82,21 +91,30 @@ class CustomPackageEdit extends Component {
...
 
     if (selectionStatusChanged) {
       stateUpdates = {
-        initialValues: CustomPackageEdit.getInitialValues(nextProps.model),
+        initialValues: CustomPackageEdit.getInitialValues(nextProps.model, proxyTypes),
         packageSelected: isSelected
       };
     }
 
+    if (isProxyTypesLoaded && !proxyTypesLoaded) {
```
MaksymDryha: I think it's better to name `proxyTypesLoaded` something like `wasProxyTypesLoaded`, because `isProxyTypesLoaded && !proxyTypesLoaded` is very confusing. Or we could just not destructure `proxyTypesLoaded` and write `isProxyTypesLoaded && !prevState.proxyTypesLoaded`

src/components/package/edit-managed/managed-package-edit.js
```diff
@@ -80,25 +89,34 @@ class ManagedPackageEdit extends Component {
...
 
     if (selectionStatusChanged || providerTokenWasLoaded) {
       stateUpdates = {
-        initialValues: ManagedPackageEdit.getInitialValues(nextProps.model, nextProps.provider),
+        initialValues: ManagedPackageEdit.getInitialValues(nextProps.model, nextProps.provider, proxyTypes),
         packageSelected: isSelected
       };
     }
 
+    if (isProxyTypesLoaded && !proxyTypesLoaded) {
```
MaksymDryha: same here

PR #994, opened 2020-03-31
UIEH-835: implement filtering packaes by access status types

MaksymDryha:
This PR adds the possibility to filter packages by access status types

### DIFF
--- a/src/components/search-form/search-form.js
+++ b/src/components/search-form/search-form.js
@@ -49,9 +49,11 @@ class SearchForm extends Component {
     searchByTagsEnabled: PropTypes.bool.isRequired,
     searchField: PropTypes.string,
     searchFilter: PropTypes.shape({
+      'access-type': PropTypes.string,
       filter: PropTypes.object,
       q: PropTypes.string,
       searchfield: PropTypes.string,
+      tags: PropTypes.string,
     }),
     searchString: PropTypes.string,
     searchType: PropTypes.oneOf(validSearchTypes).isRequired,
@@ -68,6 +70,7 @@ class SearchForm extends Component {
     displaySearchTypeSwitcher: true,
     displaySearchButton: true,
     searchString: '',
+    searchFilter: {},
   };
 
   constructor(props) {
@@ -262,7 +265,7 @@ class SearchForm extends Component {
     const {
       accessTypesStoreData,
       searchByAccessTypesEnabled,
-      searchFilter = {},
+      searchFilter,
     } = this.props;
 
     const {
--- a/src/components/search-form/search-form.js
+++ b/src/components/search-form/search-form.js
@@ -193,7 +193,7 @@ class SearchForm extends Component {
           </span>
         )}
         onClick={onStandaloneFilterToggle('access-type')}
-        data-test-tags-checkbox
+        data-test-toggle-access-types-filter-checkbox
       />
     );
   }
@@ -286,7 +286,7 @@ class SearchForm extends Component {
       : accessStatusTypesExist && (
         <div
           className={styles['search-filters']}
-          data-test-eholdings-tag-filter
+          data-test-eholdings-access-types-filter
         >
           <Accordion
             label={<FormattedMessage id="ui-eholdings.settings.accessStatusTypes" />}
@@ -304,7 +304,7 @@ class SearchForm extends Component {
               {
                 label => (
                   <MultiSelectionFilter
-                    id="accessTypeFilter"
+                    id="accessTypeFilterSelect"
                     ariaLabel={label}
                     dataOptions={this.getAccessTypesDataOptions()}
                     name="access-type"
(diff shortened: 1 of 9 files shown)


src/components/search-form/search-form.js
```diff
@@ -106,11 +112,11 @@ class SearchForm extends Component {
     this.props.onFilterChange(sort, searchFilter);
   };
 
-  handleUpdateTagFilter = (filter) => {
-    const tagFilter = {
-      tags: filter.values.length > 0 ? filter.values.join(',') : undefined
+  handleStandaloneFilterChange = filter => {
+    const formattedFilter = {
+      [filter.name]: filter.values.length > 0 ? filter.values.join(',') : undefined
```
Vladyslav-Velytskyi: I think here could be used chaining operator
```suggestion
      [filter.name]: filter?.values.join(',') || undefined
```
or without `undefined` if empty string is possible

src/components/search-form/search-form.js
```diff
@@ -229,6 +258,77 @@ class SearchForm extends Component {
...
+      'access-type': accessTypes = ''
+    } = searchFilter;
+
+    const {
+      sections,
+    } = this.state;
+
+    const accessTypesList = accessTypes ? accessTypes.split(',').map(accessType => {
+      return accessType.toLowerCase();
+    }) : [];
```
Vladyslav-Velytskyi: You can remove `return` from here
```suggestion
    const accessTypesList = accessTypes?.split(',').map(accessType => accessType.toLowerCase()) || [];
```

src/components/search-form/search-form.js
```diff
@@ -229,6 +258,77 @@ class SearchForm extends Component {
...
+
+    accessTypesList.sort();
+
+    const accessStatusTypesExist = !!accessTypesStoreData?.items?.data?.length;
+
+    return accessTypesStoreData?.isLoading
+      ? <Icon icon="spinner-ellipsis" />
+      : accessStatusTypesExist && (
```
Vladyslav-Velytskyi: Please change `&&` to ternary operator with null. If `accessStatusTypesExist` is `false` we will render `0`

src/components/search-form/search-form.js
```diff
@@ -229,6 +258,77 @@ class SearchForm extends Component {
...
+          data-test-eholdings-tag-filter
+        >
+          <Accordion
+            label={<FormattedMessage id="ui-eholdings.settings.accessStatusTypes" />}
+            id="accessTypesFilter"
+            separator={false}
+            open={sections.accessTypesFilter}
+            closedByDefault
+            header={FilterAccordionHeader}
+            displayClearButton={accessTypesList.length > 0}
```
Vladyslav-Velytskyi: Please remove `0`
```suggestion
            displayClearButton={accessTypesList.length}
```

src/components/search-form/search-form.js
```diff
@@ -229,6 +258,77 @@ class SearchForm extends Component {
...
+        </div>
+      );
+  }
+
+  getAccessTypesDataOptions() {
+    const { accessTypesStoreData } = this.props;
+
+    return accessTypesStoreData.items.data.map(accessType => ({
+      value: accessType.attributes.name.toLowerCase(),
+      label: accessType.attributes.name.toLowerCase(),
```
Vladyslav-Velytskyi: Does value equal to label?

src/components/search-form/search-form.js
```diff
@@ -339,10 +440,11 @@ class SearchForm extends Component {
           {Filters && (
             <div role="tablist">
               {this.renderTagFilter()}
+              {searchType === searchTypes.PACKAGES && this.renderAccessTypesFilter()}
               <Filters
                 activeFilters={combinedFilters}
                 onUpdate={this.handleUpdateFilter}
-                disabled={searchByTagsEnabled}
+                disabled={searchByTagsEnabled || searchByAccessTypesEnabled}
```
Vladyslav-Velytskyi: Cover `searchByTagsEnabled || searchByAccessTypesEnabled` in some variable cause you use it in this file three times

src/components/search-form/search-form.js
MaksymDryha: `false` won't be rendered as `0`

src/components/search-form/search-form.js
MaksymDryha: we need explicit `undefined` here, but to write `filter.values.join(',') || undefined` is a nice suggestion

src/components/search-form/search-form.js
MaksymDryha: yes. back-end filters packages using access status types' labels, not ids

src/components/search-form/search-form.js
Vladyslav-Velytskyi: By same implementation we have this [bug](https://issues.folio.org/browse/UIEH-859)

MaksymDryha: @Vladyslav-Velytskyi, we probably have it because we forgot to transform a number value into a boolean value (`{something.length && ...}` should be `{!!something.length && ...}`). `false` is not rendered as `0`. Also, I'm not able to reproduce this bug.

PR #1014, opened 2020-04-12
UIEH-856: Ability to create multiple Knowledge base configurations

Vladyslav-Velytskyi:
## Purpose
 - Add possibility to create Knowledge base configuration
 - Add possibility to create KB name

## Issues
[UIEH-855](https://issues.folio.org/browse/UIEH-855)
[UIEH-856](https://issues.folio.org/browse/UIEH-856)

## Screenshots
image


test/bigtest/network/config.js
```diff
@@ -670,7 +671,7 @@ export default function config() {
     return {};
   });
 
-  this.get('/custom-labels', {
+  this.get('/kb-credentials/:id/custom-labels', {
```
BogdanDenis, line 674: there are already GET and PUT `/kb-credentials/:id/custom-labels` routes defined below
we need both `/custom-labels` and `/kb-credentials/:id/custom-labels` routes because first one is used in e-holdings to fetch custom labels for a KB that current user has access to and second one is to get custom labels for a particular KB configuration

test/bigtest/network/config.js
```diff
@@ -706,7 +707,7 @@ export default function config() {
     }],
   });
 
-  this.put('/custom-labels', (schema, request) => request.requestBody);
+  this.put('/kb-credentials/:id/custom-labels', (schema, request) => request.requestBody);
```
BogdanDenis, line 710: same here

src/redux/actions/delete-kb-credentials-failure.js
```diff
@@ -1,6 +1,6 @@
 export const DELETE_KB_CREDENTIALS_FAILURE = 'DELETE_KB_CREDENTIALS_FAILURE';
 
-export const deleteKBCredentialsFailure = ({ errors }) => ({
+export const deleteKBCredentialsFailure = error => ({
```
BogdanDenis, line 3: Please check if error messages are displayed correctly.

In `/epics/delete-kb-credentials.js` we're passing an object with `errors` property to action as payload
```
.catch(errors => Observable.of(deleteKBCredentialsFailure({ errors })));
```
so in either an action or a reducer we should destructure this object to actually get an array of errors. If we don't then in store we'll have structure like so:
```
isLoaded: false,
isFailed: true,
...
errors: {
  errors: [...]
}
```

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
```diff
@@ -148,9 +235,19 @@ export default class SettingsKnowledgeBase extends Component {
...
+              />
+            )}
           </SettingsForm>
         )}
       />
     );
   }
 }
+
+export default connect()(injectIntl(SettingsKnowledgeBase));
```
BogdanDenis, line 253: why do we need an empty `connect` here?

PR #1023, opened 2020-04-17
UIEH-857: implement settings assigned users page

MaksymDryha:
## Approach
* configure redux for POSTing (assignment) KB users
* configure redux for GETting patron groups names to display in the users' list
* create `SettingsAssignedUsersRoute ` connected component responsible for fetching assigned users and user groups
* create `SettingsAssignedUsers` component that is responsible for user interaction logic and displaying of assigned users


src/redux/epics/post-kb-credentials-user.js
```diff
@@ -0,0 +1,27 @@
...
+
+  return action$
+    .filter(action => action.type === POST_KB_CREDENTIALS_USER)
+    .mergeMap(({ payload }) => {
+      const { credentialsId, userData } = payload;
+
+      return kbCredentialsUsersApi
+        .assignUser(state.okapi, credentialsId, userData)
+        .map(postKBCredentialsUserSuccess)
+        .catch(errors => Observable.of(postKBCredentialsUserFailure(errors)));
```
BogdanDenis, line 25: I think here should be `{ errors }` because in `postKBCredentialsUserFailure` action argument is desctructured

src/redux/reducers/kb-credentials-users.js
```diff
@@ -39,7 +42,24 @@ const handlers = {
     hasFailed: true,
     errors: formatErrors(action.payload),
   }),
-
+  [POST_KB_CREDENTIALS_USER]: state => ({
+    ...state,
+    isLoading: true,
```
BogdanDenis, line 47: Please update `hasLoaded` here and in `[POST_KB_CREDENTIALS_USER_SUCCESS]`

PR #1035, opened 2020-05-05
UIEH-834: FE - Filter Titles by Access Status Types

BogdanDenis:
## Purpose
- Add ability to filter titles by Access Status Types

## Issue

[UIEH-834](https://issues.folio.org/browse/UIEH-834)

## Coverage

image


test/bigtest/network/helpers.js
```diff
@@ -158,3 +158,16 @@ export function nestedResourceRouteFor(foreignKey, resourceType, filter = () =>
...
+ * in format 'filter[access-type]=Trial&filter[access-type]=Subscription
+ * @param {String} query - query part of url without '?'
+ * @returns {Array} array of Access Types
+ */
+export function getAccessTypesFromFilterQuery(query = '') {
+  return query
+    .split('&')
+    .filter(param => param.split('=')[0] === 'filter[access-type]')
+    .map(param => param.split('=')[1]);
+}
```
Vladyslav-Velytskyi, line 173: I think, in this case, it's better to use `'qs'` module, it helps to parse string without consiquence of array methods

```
import queryString from 'qs';

export function getAccessTypesFromFilterQuery(query = '') {
  const parsedQuery = queryString.parse(query, { ignoreQueryPrefix: true });

  return parsedQuery['filter[access-type']; 
}
```

PR #1044, opened 2020-05-13
UIEH-861: Provider Record Packages UI update

Vladyslav-Velytskyi:
## Description
UI updates for Provider record list of Packages

## Issue
[UIEH-861](https://issues.folio.org/browse/UIEH-861)

## Screenshot
image

### DIFF
--- a/src/components/search-package-list-item/search-package-list-item.js
+++ b/src/components/search-package-list-item/search-package-list-item.js
@@ -18,7 +18,7 @@ function SearchPackageListItem({
   packageName,
   onClick,
   headingLevel,
-  isVisibleProviderName,
+  showProviderName,
 }) {
   return !item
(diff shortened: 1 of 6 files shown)


src/components/search-package-list-item/search-package-list-item.js
```diff
@@ -108,6 +112,10 @@ SearchPackageListItem.propTypes = {
   packageName: PropTypes.string,
 };
 
+SearchPackageListItem.propTypes = {
```
BogdanDenis, line 115: I believe this should be `defaultProps`, right?

Vladyslav-Velytskyi: Yep, you're right, misprint

PR #1063, opened 2020-06-02
UIEH-899: Settings > eholdings > KB credentials page | UI updates

BogdanDenis:
## Purpose
Change save KB form button text

Related PR in `stripes-components`: https://github.com/folio-org/stripes-components/pull/1281


zburke:
Alternatively, use `stripes-core.button.save`, which already exists. There isn't firm guidance about which repo these common values should live in.

PR #1076, opened 2020-06-12
UIEH-864: Fix Access status types settings permissions

BogdanDenis:
## Purpose
- Remove Access Types `collection.get` and `item.get` permissions from Settings permissions, so that everyone has view permissions in eHoldings records. To see Settings related permission set (`Settings (eholdings): Can view access status types`) is still required.


BogdanDenis:
Issue not reproducible

PR #1087, opened 2020-07-01
UIEH-916: reset router state when visiting eholdings

MaksymDryha:
We have routing issues caused by preserving the router state when navigating outside of the eholdings app.  This PR fixes these issues by resetting the router state after the eholdings root component is mounted (if eholdings app is being mounted, than it is safe to say that previously we were in another application and thus, the eholdings app router state should be reset).

[Ticket](https://issues.folio.org/browse/UIEH-916)


src/index.js
```diff
@@ -31,6 +31,13 @@ import SettingsAssignedUsersRoute from './routes/settings-assigned-users-route';
 
 class EHoldings extends Component {
   static propTypes = {
+    history: PropTypes.shape({
```
BogdanDenis, line 34: I think we should use `react-router-prop-types` here to handle the shape of `history`, `location` and `match` props

src/components/details-view/details-view.js
```diff
@@ -208,8 +208,15 @@ class DetailsView extends Component {
       location,
     } = this.props;
 
+    const previousEHoldingsLocationKnown = location.search.includes('searchType');
+
     if (location.state && location.state.eholdings) {
       history.goBack();
+    } else if (previousEHoldingsLocationKnown) {
```
BogdanDenis, line 215: Please add a test to cover this case

MaksymDryha:
> Please, update changelog

@Vladyslav-Velytskyi , thanks, I'm always forgetting about changelog🤦

src/components/details-view/details-view.js
MaksymDryha: @BogdanDenis , done

PR #1096, opened 2020-07-10
UIEH-915: Fix permission error in devtools console

Vladyslav-Velytskyi:
Add translations for permission names in the scope of the [UIEH-915](https://issues.folio.org/browse/UIEH-915).

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,19 +102,34 @@
       {
         "permissionName": "module.eholdings.enabled",
         "displayName": "UI: eHoldings module is enabled",
-        "visible": true,
+        "visible": false,
         "subPermissions": [
           "kb-ebsco.all",
           "tags.all"
         ]
       },
       {
+        "permissionName": "ui-eholdings.app.enabled",
+        "subPermissions": [
+          "module.eholdings.enabled"
+        ],
+        "visible": true
+      },
+      {
         "permissionName": "settings.eholdings.enabled",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -116,6 +123,13 @@
           "settings.enabled"
         ]
       },
```
MaksymDryha, line 125: Shouldn't this permission be hidden too since now we have `ui-eholdings.settings.enabled` which is meant to be shown in the permissions list?

package.json
```diff
@@ -102,12 +102,19 @@
...
+      {
+        "permissionName": "ui-eholdings.app.enabled",
+        "subPermissions": [
+          "module.eholdings.enabled"
+        ],
+        "visible": true
+      },
       {
         "permissionName": "settings.eholdings.enabled",
         "displayName": "Settings (eHoldings): display list of settings pages",
```
MaksymDryha, line 120: Shouldn't this permission be hidden too since now we have `ui-eholdings.settings.enabled` which is meant to be shown in the permissions list?

Vladyslav-Velytskyi: By the default `visible` gets false, but you are right, it will better to set this param, to make it more readable and not confuse other developers

MaksymDryha: My bad. It seemed to me that ‘visible: true’ was written there

PR #1106, opened 2020-07-28
UIIH-907: implement package selection confirmation modal

MaksymDryha:
With these changes, when a user wants to select a package, they will be asked to confirm this action:
image

Also, the selection button content was changed from Add to holdings to Add package to holdings (when there are no titles selected) and from Add all to holdings to Add all titles to holdings (when there is at least one title selected):
image

image


MaksymDryha:
> Please, update CHANGELOG

Thanks, done

PR #1107, opened 2020-08-04
UIEH-909: Add possibility to delete KB credentials

MaksymDryha:
The goal of this PR is to implement the functionality of deleting KB credentials
## Things worth mentioning
* To display notifications, I used `CalloutContext` provided by `stripes-core`, not eholdings app's `<Toaster>`. After credentials are removed, the user is redirected to `/settings/eholdings` page. In this case, if  eholdings `Toaster` was used, it would be unmounted right after the user was redirected, resulting in no notification visible. `CalloutContext` gives access to the global `Callout` component instance which is available regardless of the current route and app.
* Because I used `Callout` from stripes in eholdings, now we are using 2 different components to display notifications. These components have different styling which is bad in terms of design. During a discussion with PO, it was decided to apply the same styling that is used for stripes `Callout` to eholdings `Toaster`. Next steps could be refactoring eholdings to actually use stripes `Callout` instead of just copying its styling
## Screens
image
image
image


src/routes/settings-knowledge-base.js
```diff
@@ -99,39 +143,59 @@ class SettingsKnowledgeBaseRoute extends Component {
...
-                  isCreateMode={this.state.isCreateMode}
-                />
-              </TitleManager>
-            )}
-          </FormattedMessage>
-        )}
-      </FormattedMessage>
+      <CalloutContext.Consumer>
+        {({ sendCallout }) => {
+          this.sendDeleteSuccessCallout = sendCallout;
```
MaksymDryha, line 159: `sendCallout` needs to be called inside `didUpdate`. The only way to make it available inside `didUpdate` that I managed to come up with is saving this function to `this`. Any thoughts about how to make it better?

src/components/settings/settings-form/settings-form.js
```diff
@@ -21,6 +21,7 @@ export default class SettingsForm extends Component {
       invalid: PropTypes.bool,
       pristine: PropTypes.bool,
     }).isRequired,
+    lastMenu: PropTypes.element,
```
Vladyslav-Velytskyi, line 24: Why this prop isn't required?

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
```diff
@@ -174,85 +241,95 @@ class SettingsKnowledgeBase extends Component {
...
+                    <FormattedMessage id="ui-eholdings.settings.kb.customerId">
+                      {label => (
+                        <Field
+                          label={label}
+                          name="customerId"
+                          component={TextField}
+                          type="text"
+                          autoComplete="off"
+                          validate={value => (
+                            value ? undefined : <FormattedMessage id="ui-eholdings.validate.errors.settings.customerId" />
```
Vladyslav-Velytskyi, line 300: As I remember if value is valid validation function should return null (maybe it doesn't change your implementation but it will look like other parts of the application)

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
```diff
@@ -174,85 +241,95 @@ class SettingsKnowledgeBase extends Component {
...
+                          )}
+                          required
+                          aria-label={label}
+                        />
+
+                      )}
+                    </FormattedMessage>
+                  </div>
+
+                  <p><FormattedMessage id="ui-eholdings.settings.kb.url.ebsco.customer.message" /></p>
```
Vladyslav-Velytskyi, line 330: But lease, split each tag in separate lines

src/redux/actions/confirm-delete-kb-credentials.js
```diff
@@ -0,0 +1,6 @@
+export const CONFIRM_DELETE_KB_CREDENTIALS = 'CONFIRM_DELETE_KB_CREDENTIALS';
+
+
```
Vladyslav-Velytskyi, line 3: Please, delete this empty line

src/routes/settings-knowledge-base.js
```diff
@@ -49,11 +63,41 @@ class SettingsKnowledgeBaseRoute extends Component {
...
+      this.setState({ currentKBName: this.getCurrentKBName() });
     }
+
+    if (kbCredentials.hasDeleted) {
+      history.replace('/settings/eholdings');
+      this.sendDeleteSuccessCallout({
+        type: 'success',
+        message: (
+          <span data-test-kb-deleted-notification>
+            <FormattedMessage id="ui-eholdings.settings.kb.delete.toast" values={{ kbName: currentKBName }} />
```
Vladyslav-Velytskyi, line 82: Please, split each props in separate lines

src/routes/settings-knowledge-base.js
```diff
@@ -99,39 +143,59 @@ class SettingsKnowledgeBaseRoute extends Component {
...
+        {({ sendCallout }) => {
+          this.sendDeleteSuccessCallout = sendCallout;
+
+          return (
+            <FormattedMessage id="ui-eholdings.label.settings">
+              {pageLabel => (
+                <FormattedMessage id="ui-eholdings.settings.kb">
+                  {recordLabel => (
+                    <TitleManager
+                      page={pageLabel}
```
BogdanDenis, line 167: Can we use `intl.formatMessage` here instead? It'll simplify the structure a lot

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
```diff
@@ -174,85 +241,95 @@ class SettingsKnowledgeBase extends Component {
...
-                <p><FormattedMessage id="ui-eholdings.settings.kb.url.ebsco.customer.message" /></p>
-              </>
-            )}
+            {kbCredentials.isLoading
+              ? <Icon icon="spinner-ellipsis" />
+              : (
+                <>
+                  <div data-test-eholdings-settings-kb-name>
+                    <FormattedMessage id="ui-eholdings.name">
+                      {label => (
```
BogdanDenis, line 267: Here we could also use `int.formatMessage`

src/components/settings/settings-form/settings-form.js
MaksymDryha: @Vladyslav-Velytskyi , because there are cases when last menu should not be rendered

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
MaksymDryha: Good point! Other fields (customer id and api key) can also be simplified by using `intl.formatMessage`

PR #1121, opened 2020-08-18
UIEH-905 No providers/packages/titles found does not show the search term entered.

yelyzavetakhokhlova:
[UIEH-905](https://issues.folio.org/browse/UIEH-905)  No providers/packages/titles found does not show the search term entered.
## Purpose
No Provider/Package/Title results should display the following message: [ No providers/packages/titles found for "<<show the search term entered in search box>>." Please check your spelling and filters. ]

## Screenshots
image

## Tests
image

### DIFF
--- a/src/components/no-results-message/noResultsMessage.css
+++ b/src/components/no-results-message/noResultsMessage.css
@@ -1,0 +1,33 @@
+@import '@folio/stripes-components/lib/variables';
+
+.noResultsMessage {
+  padding: 15px;
+  color: var(--color-text-p2);
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  background: var(--color-fill);
+  flex-direction: column;
+}
+
+.noResultsMessage,
+.noResultsMessageLabelWrap {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.noResultsMessageLabelWrap {
+  font-size: var(--font-size-large);
+}
+
+.noResultsMessageIcon,
+.noResultsMessageLabel {
+  margin: 0 4px;
+}
+
+.noResultsMessageButton {
+  margin-top: 15px;
+}
--- a/src/components/no-results-message/noResultsMessage.css
+++ b/src/components/no-results-message/noResultsMessage.css
@@ -5,11 +5,9 @@
   color: var(--color-text-p2);
   position: absolute;
   top: 0;
-  bottom: 0;
-  left: 0;
-  right: 0;
+  width: 100%;
+  height: 100%;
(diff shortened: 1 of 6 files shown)


src/components/no-results-message/noResultsMessage.js
```diff
@@ -0,0 +1,25 @@
...
+import css from './noResultsMessage.css';
+
+const propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+const NoResultsMessage = ({ children, ...rest }) => {
+  return (
+    <div className={css.noResultsMessage} {...rest}>
+      <div className={css.noResultsMessageLabelWrap}>
```
Vladyslav-Velytskyi, line 15: @BogdanDenis, @MaksymDryha Should we replace this tag with `<p>` like in `SearchAndSort` component?

src/components/no-results-message/noResultsMessage.css
```diff
@@ -0,0 +1,33 @@
+@import '@folio/stripes-components/lib/variables';
+
+.noResultsMessage {
+  padding: 15px;
+  color: var(--color-text-p2);
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
```
Vladyslav-Velytskyi, line 10: `top`, `bottom`, `left`, `right`, properties are used for positioning, not for size of element. In my opinion is better to use `height` and `width` for it

```suggestion
  top: 0;
  width: 100%;
  height: 100%;
```

src/components/no-results-message/noResultsMessage.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.noResultsMessageLabelWrap {
+  font-size: var(--font-size-large);
+}
+
+.noResultsMessageIcon,
+.noResultsMessageLabel {
+  margin: 0 4px;
+}
```
Vladyslav-Velytskyi, line 29: Just suggestion, if you need space between two elements, better to rewrite it as in example below
```suggestion
.noResultsMessageLabel {
  margin-left: 8px;
}
```
Pros of this variant that we don't need for special class for the icon

src/components/no-results-message/noResultsMessage.js
MaksymDryha: @Vladyslav-Velytskyi , makes sense

src/components/no-results-message/noResultsMessage.css
MaksymDryha: @Vladyslav-Velytskyi, considering all this stuff with dynamic list height calculations + absolute positioning, I'm not sure if that will work, but we can try.

PR #1175, opened 2020-10-26
 UIEH-968: Settings - Usage consolidation - Usage consolidation ID

Vladyslav-Velytskyi:
## Purpose
Implement new Settings tab - usage Consolidation
Form has only one field - Usage Consolidation ID

https://issues.folio.org/browse/UIEH-968

## Screenshots
Page View
image
Validation for empty field
image
Validation if usage consolidation ID is invalid
image


src/api/usage-consolidation.js
```diff
@@ -0,0 +1,48 @@
+import {
+  getHeaders,
+  doRequest,
+  createUrl,
+} from './common';
+
+const API_URL = '/eholdings/kb-credentials';
```
BogdanDenis, line 7: I think we also need to add `/uc` at the end here

src/redux/epics/post-usage-consolidation.js
```diff
@@ -0,0 +1,23 @@
...
+export default ({ usageConsolidationApi }) => (action$, store) => {
+  return action$
+    .filter(action => action.type === POST_USAGE_CONSOLIDATION)
+    .mergeMap(({ payload }) => {
+      const { credentialsId, data } = payload;
+
+      return usageConsolidationApi
+        .postUsageConsolidation(store.getState().okapi, credentialsId, { data })
+        .map(postUsageConsolidationSuccess)
+        .catch(errors => Observable.of(postUsageConsolidationFailure(errors)));
```
BogdanDenis, line 21: I think you should wrap `errors` in an object here

src/redux/reducers/usageConsolidation.js
```diff
@@ -0,0 +1,72 @@
...
+      ...state,
+      isLoading: false,
+      data,
+    };
+  },
+  [GET_USAGE_CONSOLIDATION_FAILURE]: handleError,
+  [POST_USAGE_CONSOLIDATION_FAILURE]: handleError,
+  [PATCH_USAGE_CONSOLIDATION_FAILURE]: handleError,
+  [POST_USAGE_CONSOLIDATION_SUCCESS]: (state, { payload }) => {
+    const { data } = payload;
```
BogdanDenis, line 37: I think the response from back-end looks like this:
```
{
  type: 'ucSettings',
  id: 'some-id',
  attributes: {
    name: 'uc name',
    startMonth: 'jan',
    currency: 'USD',
  }
}
```

src/api/usage-consolidation.js
Vladyslav-Velytskyi: End point `/eholdings/kb-credentials/uc` will be unvalid and wrong for our backend

src/redux/epics/post-usage-consolidation.js
Vladyslav-Velytskyi: No, It works correctly, but thank you to watching on it because we need to rewrite by the same way patch epic

BogdanDenis (approved):

PR #1179, opened 2020-10-28
UIEH-969: Settings | EBSCO Usage Consolidation (UC) + FOLIO integration | Show usage statistics default display Field

BogdanDenis:
## Description
Add Usage Consolidation Platform Type field

## Issue
[UIEH-969](https://issues.folio.org/browse/UIEH-969)


src/components/settings/settings-usage-consolidation/settings-usage-consolidation.js
```diff
@@ -64,6 +66,21 @@ const SettingsUsageConsolidation = ({
...
+    {
+      label: formatMessage({ id: `ui-eholdings.settings.usageConsolidation.platformType.${platformTypes.PUBLISHER_ONLY}` }),
+      value: platformTypes.PUBLISHER_ONLY,
+    },
+    {
+      label: formatMessage({ id: `ui-eholdings.settings.usageConsolidation.platformType.${platformTypes.NON_PUBLISHER_ONLY}` }),
+      value: platformTypes.NON_PUBLISHER_ONLY,
+    },
+  ];
+
```
Vladyslav-Velytskyi, line 83:
```suggestion
  const platformTypesDataOptions = platformTypes.map(paltformType => ({
     label: formatMessage({ id: `ui-eholdings.settings.usageConsolidation.platformType.${platformType}` }),
     value: platformType,
   });
```

src/routes/settings-usage-consolidation-route.js
```diff
@@ -56,12 +56,19 @@ const SettingsUsageConsolidationRoute = ({
       ...updatedData
     } = params;
 
+    const attributes = {
+      currency: 'USD',
+      platformType: updatedData.platformType,
+      startMonth: updatedData.startMonth,
+    };
```
Vladyslav-Velytskyi, line 63: You missed `customerKey` value

BogdanDenis: It's added later. This is because back-end sends this field as '*' characters. And if we include this field always even if it's not changed then the requests would fail because it's not a valid value.

Vladyslav-Velytskyi: Sorry, I forgot about it, you're right

Vladyslav-Velytskyi (approved):

PR #1181, opened 2020-11-02
UIEH-965: Settings - Usage Consolidation - Currency field

Vladyslav-Velytskyi:

### DIFF
--- a/src/components/settings/settings-usage-consolidation/settings-usage-consolidation.js
+++ b/src/components/settings/settings-usage-consolidation/settings-usage-consolidation.js
@@ -19,6 +19,11 @@ import SettingsForm from '../settings-form';
 
 const propTypes = {
   clearUsageConsolidationErrors: PropTypes.func.isRequired,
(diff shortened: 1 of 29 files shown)


src/components/settings/settings-usage-consolidation/settings-usage-consolidation.js
```diff
@@ -66,6 +79,27 @@ const SettingsUsageConsolidation = ({
...
+  const parseUsageConsolidationId = value => {
+    if (customerKeyIsInvalid) {
+      clearUsageConsolidationErrors();
+    }
+
+    return value;
+  };
+
+  const defaultCurrency = {
+    value: '',
```
BogdanDenis, line 91: Do we not need to set value to 'USD' here?

Vladyslav-Velytskyi: That was timely value, because currency is required field

BogdanDenis (approved):

PR #1191, opened 2020-11-11
UIEH-941: Package Record: Display Usage & Analysis Accordion 

BogdanDenis:
## Purpose
- Add Usage Consolidation accordion to Package View page
- Add permissions to get Usage Consolidation data for packages, titles and resources

## Issue
[UIEH-941](https://issues.folio.org/browse/UIEH-941)


src/features/usage-consolidation-accordion/usage-consolidation-accordion.js
```diff
@@ -0,0 +1,102 @@
...
+    );
+  };
+
+  if (usageConsolidation.isFailed || canViewUsageConsolidation) {
+    return null;
+  }
+
+  return (
+    <>
+      {usageConsolidation.data &&
```
Vladyslav-Velytskyi, line 75: `usageConsolidation.data` is an object and in all cases it will be true, so you need to check if `data` has `credentialsId` property
```suggestion
      {usageConsolidation.data?.credentialsId &&
```

src/features/usage-consolidation-accordion/usage-consolidation-accordion.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+const UsageConsolidationAccordion = ({
+  getUsageConsolidation,
+  headerProps,
+  id,
+  isOpen = true,
+  onToggle,
+  usageConsolidation,
+  section: Section,
```
Vladyslav-Velytskyi, line 37: I'm not sure that we need this prop. By the logic of this feature(and as other eholdings features), component must to generate it's own content and it is not a part of this story. You can just leave this accordion empty. YAGNI.

src/features/usage-consolidation-accordion/usage-consolidation-accordion.js
```diff
@@ -0,0 +1,102 @@
...
+  getUsageConsolidation: PropTypes.func.isRequired,
+  headerProps: PropTypes.object,
+  id: PropTypes.string.isRequired,
+  isOpen: PropTypes.bool,
+  onToggle: PropTypes.func.isRequired,
+  section: PropTypes.node.isRequired,
+  usageConsolidation: PropTypes.shape({
+    data: PropTypes.object,
+    errors: PropTypes.array,
+    isFailed: PropTypes.bool.isRequired,
```
Vladyslav-Velytskyi, line 26: I see you update usage consolidation reducer, please update usageConsolidation shape with `isLoaded` and `isLoading`. Also I think it will be great to create shape for this prop and use it in other files.

P.S. `data` and `errors` are required props too.

Vladyslav-Velytskyi (changes requested):

src/features/usage-consolidation-accordion/usage-consolidation-accordion.js
BogdanDenis: I added this prop here, because I think that this component can be reused for Title and Resources pages, but I haven't added any content here, just a simple placeholder for now

src/components/details-view/details-view.js
```diff
@@ -161,11 +169,13 @@ class DetailsView extends Component {
 
       // don't do these calculations when not scrolling the container
     } else if (e.currentTarget === e.target) {
+      const usageConsolidationHeight = this.$usageConsolidationContainer?.offsetHeight || 0;
+
       const top = e.currentTarget.scrollTop;
       const height = e.currentTarget.offsetHeight;
-      const scrollHeight = e.currentTarget.scrollHeight;
+      const scrollHeight = e.currentTarget.scrollHeight - usageConsolidationHeight;
```
BogdanDenis, line 176: We need to subtract usage consolidation accordion height here, because we need to know where titles list end to know that we scrolled down to it

src/components/details-view/details-view.js
```diff
@@ -174,6 +184,7 @@ class DetailsView extends Component {
       // if bottoming out, enable isSticky
       if (bottomedOut && !isSticky && isListScrollbarNeeded) {
         this.setState({ isSticky: true });
+        this.$sticky.scrollIntoView();
```
BogdanDenis, line 187: I added this to scroll up to titles list accordion so it always stays fully in view when a user has scrolled down to it. Without this there are cases when a user scrolls down very fast and a part of accordion is not visible

src/features/usage-consolidation-accordion/usage-consolidation-accordion.js
Vladyslav-Velytskyi: You are right, we will use this component for Resources too. But accordion for packages and accordion for resources will be equal. In my opinion we will only pass usage consolidation data to this accordion and it will create all tables inside, we don't need any placeholder here right now

BogdanDenis: It's just that data structure of costperuse for packages and resources is different:
`/eholdings/packages/{packageId}/costperuse`:
```
{
  "data": [
    {
      "type": "packageCostPerUse",
      "attributes": {
        "platforms": [
          {
            "name": "InjentaConnect"
          },
          {
            "name": "Proquest"
          },
          {
            "name": "Willey Interscience"
          },
          {
            "name": "Willey Online Library"
          }
        ],
        "analysis": {
          "publisherPlatforms": {
            "cost": 5175,
            "usage": 6149,
            "costPerUse": 0.8416002602
          }
        }
      }
    }
  ]
}
```
and `/eholdings/resources/{resourceId}/costperuse`:
```
{
  "data": [
    {
      "type": "resourceCostPerUse",
      "attributes": {
        "platforms": [
          {
            "name": "InjentaConnect",
            "counts": [60, 55, 50, 55, 65, 50, 50, 100, 50, 130, 68, 56],
            "total": 789
          },
          {
            "name": "Willey Interscience",
            "counts": [75, 80, 80, 105, 85, 75, 75, 75, 75, 100, 75, 75],
            "total": 975
          },
          {
            "name": "Willey Online Library",
            "counts": [40, 40, 40, 90, 50, 25, 22, 30, 24, 60, 40, 20],
            "total": 481
          }
        ],
        "analysis": {
          "publisherPlatforms": {
            "cost": 800.00,
            "usage": 2225,
            "costPerUse": 0.359550561797753
          }
        }
      }
    }
  ]
}
```

Because of this I there's going to be a lot of conditional logic in this component to decide what to render and I think it'll be a complicated solution.

Vladyslav-Velytskyi: I see that this two objects are identically, but values ​​for packages property `platforms` are expanded and for my opinion it is no more one additional condition inside this component instead of` section` prop. It will be great if all logic for usage & analysis will be in one place.
If we will make as you purpose we need at least to rename this prop to `platformsInfo` for example because, `section` is not quite clear name. It is just one difference between this two responses. Following this logic, expanded platform info will be only for resources and we don't need it for packages, so this prop is not required. We can create for it default prop `() => null`. In this case in will much better. But, I'm repeat, I'm not happy with this solution before implementation costperuse for one of package or resourses.

BogdanDenis: What if we just pass costperuse data and some prop to identify if it's for a package title or a resource, and inside the accordion, based on this prop we'll conditionally render imported views?
Something like this:
```
...
if (props.recordType === 'package') {
  return <PackageUsageConsolidationContent />
} else if (props.recordType === 'title') {
  return <TitleUsageConsolidationContent />
} else {
  return <ResourceUsageConsolidationContent />
}
```

This will give us the ability to only pass data to the accordion, but there wouldn't be too much code inside this single component, since it'll be split into several different components

Vladyslav-Velytskyi: Sounds better. For each field create component and combine them inside these components. But for Titles and Resources could be one component because that have equal mockups

Vladyslav-Velytskyi: Also we can just pass costperuse data, it is already has type (package, title, or resourse)

Vladyslav-Velytskyi (approved):

PR #1194, opened 2020-11-18
UIEH-945: FE | Package record | Display Summary Table

BogdanDenis:
## Description
- Add Summary table to Package Usage Consolidation accordion

## Screenshots
image

## Issue
[UIEH-945](https://issues.folio.org/browse/UIEH-945)


src/components/details-view/details-view.js
```diff
@@ -346,7 +346,7 @@ class DetailsView extends Component {
                     onToggle={onListToggle}
                     listType={listType}
                   >
-                    {renderList(isSticky)}
+                    {isListAccordionOpen ? renderList(isSticky) : null}
```
BogdanDenis, line 349: No need to render list if the accordion is closed

test/bigtest/tests/unit/epics/get-package-cost-per-use-test.js
```diff
@@ -0,0 +1,116 @@
...
+        okapi: {
+          url: 'https://folio-snapshot',
+          tenant: 'diku',
+          token: 'token',
+        },
+      };
+    },
+  };
+
+
```
Vladyslav-Velytskyi, line 34:
```suggestion
```

src/features/usage-consolidation-accordion/usage-consolidation-content-package.js
```diff
@@ -0,0 +1,142 @@
...
+  const currencySymbol = getSymbolFromCurrency(currency);
+
+  const {
+    cost,
+    costPerUse,
+    usage,
+  } = data?.attributes?.analysis;
+  const noCostPerUseAvailable = !cost && !costPerUse && !usage;
+
+  const formatCost = (value) => `${currencySymbol || ''}${value} (${currency})`;
```
Vladyslav-Velytskyi, line 57: Just suggestion
```suggestion
  const currencySymbol = getSymbolFromCurrency(currency) || '';

  const {
    cost,
    costPerUse,
    usage,
  } = data?.attributes?.analysis;
  const noCostPerUseAvailable = !cost && !costPerUse && !usage;

  const formatCost = (value) => `${currencySymbol}${value} (${currency})`;
```
It makes `formatCost` little bit more readable

src/features/usage-consolidation-accordion/usage-consolidation-content-package.js
```diff
@@ -0,0 +1,142 @@
...
+  const formatValue = (value, formatter) => {
+    if (!noCostPerUseAvailable && !value) {
+      return '-';
```
Vladyslav-Velytskyi, line 60: Better to use `<NoValue />` component from `stripes` instead of dash

src/features/usage-consolidation-accordion/usage-consolidation-content-package.js
```diff
@@ -0,0 +1,142 @@
...
+            record: 'package',
+            year,
+          }}
+        />
+      </div>
+    );
+  }
+
+  return (
+    <>
```
Vladyslav-Velytskyi, line 81: React.Fragment unnecessary here

src/features/usage-consolidation-accordion/usage-consolidation-content-package.js
```diff
@@ -0,0 +1,142 @@
...
+      <MultiColumnList
+        id="packageUsageConsolidationSummary"
+        onRowClick={() => {}}
+        contentData={[{ cost, costPerUse, usage }]}
+        visibleColumns={['cost', 'usage', 'costPerUse', 'ucActions']}
+        formatter={{
+          cost: (rowData) => formatValue(rowData.cost, formatCost),
+          costPerUse: (rowData) => formatValue(rowData.costPerUse, formatCost),
+          usage: (rowData) => formatValue(rowData.usage),
```
Vladyslav-Velytskyi, line 90: Why do we need `formatValue` here? Without formatter this function returns value or dash so we can just write
```suggestion
          usage: (rowData) => rowData.usage || <NoValue />,
```
And if we do it `formatValue` function could loose condition in return

src/api/costperuse.js
```diff
@@ -0,0 +1,33 @@
...
+    platform: filterData.platformType,
+    fiscalYear: filterData.year,
+  });
+};
+
+const getPackageCostPerUseUrl = (packageId, filterData) => {
+  return `/eholdings/packages/${packageId}/costperuse?${formatParametersForBackend(filterData)}`;
+};
+
+
```
Vladyslav-Velytskyi, line 20:
```suggestion
```

Vladyslav-Velytskyi (changes requested):

src/features/usage-consolidation-accordion/usage-consolidation-content-package.js
BogdanDenis: I've missed a bug here, `formatValue` should check that the value is falsy, but not equal to 0 (added a test case for this).
If we check `!value && value !== 0` in `formatValue` then we can't use `||` to decide what to return here

Vladyslav-Velytskyi: So, I clearly understand that we could use just `rowData.usage` here or it could be expect cases where we need `formatValue` function?

Vladyslav-Velytskyi (approved):

src/features/usage-consolidation-accordion/usage-consolidation-content-package.js
BogdanDenis: For usage we also need to show either `0` or `-` depending on what value we get from back-end. And `formatValue` implement logic that decides what to show

PR #1226, opened 2021-01-08
UIEH-1017: Fix failing eHoldings tests due to accessibility issues

BogdanDenis:
## Description
Some of the tests were failing due to a11y issues in `<Modal>` and some other components. This PR fixes these issues.

### Related PRs
https://github.com/folio-org/stripes-components/pull/1483
https://github.com/folio-org/stripes-smart-components/pull/979

## Issues
[UIEH-1017](https://issues.folio.org/browse/UIEH-1017)


src/redux/selectors/select-prop-from-data.js
```diff
@@ -1,3 +1,5 @@
+import { get } from 'lodash';
+
 export default function selectPropFromData(store, type) {
-  return store.eholdings.data[type];
+  return get(store, `eholdings.data.${type}`, {});
```
Vladyslav-Velytskyi, line 4: Just suggestion, it would be better to omit using of lodash, it could be looks like
```suggestion
  return store?.eholdings?.data?[type] || {};
```
but, all is fine, it has the right to live

Vladyslav-Velytskyi (approved):
Nice work!!

PR #1261, opened 2021-02-05
UIEH-1033: Title Record | Add notes helper app 

BogdanDenis:
## Description
Add Notes accordion to Title Show page so that users can store information on Titles records even after they are removed from holdings

## Screenshots
image

## Issues
[UIEH-1033](https://issues.folio.org/browse/UIEH-1033)

### DIFF
--- a/src/components/title/show/title-show.js
+++ b/src/components/title/show/title-show.js
@@ -6,13 +6,10 @@ import {
 } from 'react-intl';
 import { withRouter } from 'react-router';
 import ReactRouterPropTypes from 'react-router-prop-types';
-
 import { Form } from 'react-final-form';
 import createFocusDecorator from 'final-form-focus';
-
 import update from 'lodash/fp/update';
 import set from 'lodash/fp/set';
-
 import qs from 'qs';
 
 import {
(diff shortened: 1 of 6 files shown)


Yuliia-Dovhal (approved):

src/components/title/show/title-show.js
```diff
@@ -250,86 +255,98 @@ class TitleShow extends Component {
...
-                        {model.publisherName}
+            <>
+              <Accordion
+                label={<Headline size="large" tag="h3"><FormattedMessage id="ui-eholdings.title.titleInformation" /></Headline>}
+                open={sections.titleShowTitleInformation}
+                id="titleShowTitleInformation"
+                onToggle={this.handleSectionToggle}
+              >
+                <KeyValueColumns>
+                  <div>
```
Vladyslav-Velytskyi, line 266: This `<div>` could be removed. If we need `<div>` to split some blocks, better to use `<Row>` and `<Col>`

src/components/title/show/title-show.js
```diff
@@ -250,86 +255,98 @@ class TitleShow extends Component {
...
+                    <ContributorsList data={model.contributors} />
+
+                    {model.edition && (
+                      <KeyValue label={<FormattedMessage id="ui-eholdings.title.edition" />}>
+                        <div data-test-eholdings-title-show-edition>
```
Vladyslav-Velytskyi, line 271: It is a lit bit hardcoded. `<div>` with data-attribute can be removed. This data-attribute is used for a interactor, but it could be written more simply. `<KeyValue>` component has an interacotr itself in which we can catch the entered value.
This point should not be changed right now, because bigtest will be removed soon, but we must not forgot about it.

Vladyslav-Velytskyi (approved):
There were just suggestions, but thanks for fixing it!

PR #1270, opened 2021-02-15
UIEH-979: Add FOLIO keyboard shortcuts to eholdings

Yuliia-Dovhal:
[UIEH-979](https://issues.folio.org/browse/UIEH-979) Add FOLIO keyboard shortcuts to eholdings

## Purpose

- Add FOLIO keyboard shortcuts to eholdings
- Fixed translation for Provider save changes toast message

#### TODOS and Open Questions
Have to write tests for keyboard shortcuts on react-testing-library. 
I created a Jiira issue for that: [UIEH-1043](https://issues.folio.org/browse/UIEH-1043)


src/components/provider/show/provider-show.js
```diff
@@ -280,6 +285,53 @@ class ProviderShow extends Component {
...
 
+  openEditProvider = () => {
+    const {
+      stripes,
+      onEdit,
+    } = this.props;
+
+    const editPermission = stripes.hasPerm(this.editPermissionName);
+
+    if (editPermission) {
```
Vladyslav-Velytskyi, line 296: If this value returns the boolean so better to name it as `hasEditPermission` or `isEditPermission`

src/components/provider/show/provider-show.js
```diff
@@ -73,6 +76,8 @@ class ProviderShow extends Component {
     };
   }
 
+  editPermissionName = 'ui-eholdings.records.edit';
```
Vladyslav-Velytskyi, line 79: Same problem with naming
```suggestion
  RECORDS_EDIT_PERMISSION_NAME = 'ui-eholdings.records.edit';
```

src/components/package/show/package-show.js
```diff
@@ -679,6 +692,48 @@ class PackageShow extends Component {
...
+    if (this.hasEditPermission()) {
+      onEdit();
+    }
+  };
+
+  toggleAllSections = (expand) => {
+    this.setState((curState) => {
+      const newSections = expandAllFunction(curState.sections, expand);
+      return {
+        sections: newSections
```
Vladyslav-Velytskyi, line 707:
```suggestion
        sections: newSections,
```

Vladyslav-Velytskyi (changes requested):

src/components/search-paneset/search-paneset.js
```diff
@@ -48,13 +57,41 @@ export default class SearchPaneset extends Component {
     this.state = {};
   }
 
+  createDeletePermissionName = 'ui-eholdings.titles-packages.create-delete';
+
+  isPackageOrTitle = (resultsType) => !!(resultsType === searchTypes.PACKAGES || resultsType === searchTypes.TITLES);
```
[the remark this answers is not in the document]
Yuliia-Dovhal: I can agree with the last notice, that it will be more readable to convert into `resultsType !== searchTypes.PROVIDERS`. Good point, thanks!

But what about `!!` -> when you are using logic operator `||` it returns `true` or the last `false` expression. That's why it's better to convert into boolean.
For example: 
``` javascript
false || -5 // returns -5
```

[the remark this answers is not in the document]
Yuliia-Dovhal: I can agree with the last notice, that it will be more readable to convert into resultsType !== searchTypes.PROVIDERS. Good point, thanks!

But what about !! -> when you are using logic operator || it returns true or the last false expression. That's why it's better to convert into boolean.
For example:
``` javascript
false || 0 // returns 0
!!(false || 0) / returns false
```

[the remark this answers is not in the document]
Vladyslav-Velytskyi: Yep, but here you have just identity operator, it will return bool anyway
```
resultType === searchType.PACKAGES // bool
retultType === searchType.TITLES // bool 
bool || bool // bool
```

BogdanDenis (approved):

src/components/with-key-shortcuts/with-key-shortcuts.js
```diff
@@ -0,0 +1,117 @@
...
+import PropTypes from 'prop-types';
+
+import {
+  checkScope,
+  HasCommand,
+} from '@folio/stripes-components';
+
+import { handleSaveKeyFormSubmit } from '../utilities';
+
+const WithKeyShortcuts = ({
```
Vladyslav-Velytskyi, line 14: I'm not sure that this is a correct naming of this component, because naming started with word `with`  according throw naming [convention](https://airbnb.io/javascript/react/#naming) it should be a HOC, but here you have just component, so it could be better to name it `KeyShortcutsContaner`

Yuliia-Dovhal: Okay, I can rename it. But I like more `KeyShortcutsWrapper`. Is it ok for you?

Vladyslav-Velytskyi: Yeah, sure, that was just suggestion

Vladyslav-Velytskyi (approved):

PR #1276, opened 2021-02-25
UIEH-1041: Fix error message for costperuse in record

Vladyslav-Velytskyi:
## Purpose
Fix of this [PR](https://github.com/folio-org/ui-eholdings/pull/1272)
Now if no costPerUse data available it should show full text request usage table if usage data present

Issue: [UIEH-1041](https://issues.folio.org/browse/UIEH-57)

## Screenshots
image

### DIFF
--- a/src/features/usage-consolidation-accordion/usage-consolidation-content-title.js
+++ b/src/features/usage-consolidation-accordion/usage-consolidation-content-title.js
@@ -13,6 +13,7 @@ import {
 
 import FullTextRequestUsageTable from './full-text-request-usage-table';
 import SummaryTable from './summary-table';
+import NoCostPerUseAvailable from './no-cost-per-use-available';
 import {
   formatCoverageYear,
   formatCoverageFullDate,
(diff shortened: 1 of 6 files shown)


BogdanDenis (approved):

src/features/usage-consolidation-accordion/usage-consolidation-content-title.js
```diff
@@ -163,7 +164,12 @@ const UsageConsolidationContentTitle = ({
     formatter,
   };
 
-  return (
+  return noCostPerUseAvailable ? (
```
BogdanDenis, line 167: Do we also need to check here that platforms are not empty?

Vladyslav-Velytskyi: Don't know, haven't seen description of the same behavior, but thanks, I'll add it, to omit same bug in future

Yuliia-Dovhal (approved):

PR #1283, opened 2021-03-09
UIEH-1106: Fix settings assign/unassign user permission

Vladyslav-Velytskyi:
## Purpose
Remove unnecessary navigation component

Issue: [UIEH-1106](https://issues.folio.org/browse/UIEH-1106)


src/components/settings/settings.js
```diff
@@ -128,9 +128,6 @@ class Settings extends Component {
                 <FormattedMessage id="ui-eholdings.settings.accessStatusTypes" />
               </NavListItem>
             </IfPermission>
-            <NavListItem to={`/settings/eholdings/${configuration.id}/users`}>
```
BogdanDenis, line 131: It looks like we have two problems:

1. We're missing `usergroups.collection.get` permission to receive user groups in our `ui-eholdings.settings.assignedUser`.
2. `ui-eholdings.settings.assignedUser` permission is not used in `settings-assigned-users` to hide the button if user doesn't have it.

BogdanDenis (changes requested):

src/components/settings/settings.js
Vladyslav-Velytskyi: About second. True, but we hide the button in `settings`, is it not enough?

BogdanDenis: From description in ticket I got that we need to show the users list even without this permission, we only need to hide the `Assign users` button

BogdanDenis: Let's ask Khalilah what's the exact expected behavior and which permission should grant the ability to view assigned users

Vladyslav-Velytskyi: Oh, I understand where is my mistake. By the way, in your opinion it is normal that I delete this component? We have the similar wrapped with permissions below.

BogdanDenis: Well, if we need to show the assigned users to anyone then we don't need to use this permission for route

BogdanDenis (approved):

PR #1294, opened 2021-03-17
UIEH-1107: Five Colleges | Page continues to wobble when viewing list of packages or titles

BogdanDenis:
## Description
Fix issue with scroll when the page is zoomed out a lot. This is caused by comparison of scroll container and scrollable list heights. With 100% zoom they are exactly equal, but when zooming out to 67% or lower - heights start to have small difference (about 0.01px, but I assume that this value can change from person to person). And because of this `>=` comparison failed.
Suggested solution is to allow some small difference to exist between these heights to still count as if they're equal.

## Screenshots
With bug
eHoldings - Journal of Fish Biology - FOLIO (1)

Without bug
eHoldings - Journal of Fish Biology - FOLIO

## Issues
[UIEH-1107](https://issues.folio.org/browse/UIEH-1107)


Vladyslav-Velytskyi (approved):
Please, update CHANGELOG.md too

Yuliia-Dovhal (approved):

PR #1300, opened 2021-03-25
UIEH-1108: Add app context menu dropdown

Vladyslav-Velytskyi:
## Purpose
- Implement an app context menu dropdown
- Add `eholdings.css` because in future refactoring class `EHoldings` from `index.js` should be removed to separate file

Issue: [UIEH-1108](https://issues.folio.org/browse/UIEH-1108)


Yuliia-Dovhal (approved):

src/index.js
```diff
@@ -112,28 +122,71 @@ class EHoldings extends Component {
...
+                <NavListSection>
+                  <NavListItem
+                    id="eholdings-app-item"
+                    to="eholdings"
+                  >
+                    <FormattedMessage id="ui-eholdings.navigation.app" />
+                  </NavListItem>
+                  <NavListItem
+                    id="content-item"
+                    href="http://www.tfaforms.com/306425"
```
BogdanDenis, line 138: There's a https version of this site. Let's ask Khalilah if we can use it instead of http

src/index.js
```diff
@@ -112,28 +122,71 @@ class EHoldings extends Component {
...
-                <Route path={`${rootPath}/notes/new`} exact component={NoteCreate} />
-                <Route path={`${rootPath}/notes/:noteId`} exact component={NoteView} />
-                <Route path={`${rootPath}/notes/:id/edit`} exact component={NoteEdit} />
-                <Route path={`${rootPath}/`} exact component={SearchRoute} />
-              </Switch>
-            </Route>
-          </KeyShortcutsWrapper>
-        </CommandList>
+        <>
+          <AppContextMenu>
```
BogdanDenis, line 126: I think we should move this to a separate component like `EHoldingsAppContext` to not make root component too large

BogdanDenis (changes requested):

src/index.js
Vladyslav-Velytskyi: Nice catch, fixed it

BogdanDenis (approved):

PR #1308, opened 2021-03-31
UIEH-1110: Usage & analysis accordion focus issues

Yuliia-Dovhal:
[UIEH-1110](https://issues.folio.org/browse/UIEH-1110)

## Purpose
Fix couple issues with focus and one bug with scroll

## Approach
For fixing scroll issue for the last accordion, that is closed by default and after discussion with John Coburn, I put focus on the first select of Usage consolidation filter.
Maybe should be shown such approach and discussed with Khalilah Gambrell.


src/features/usage-consolidation-accordion/summary-table/summary-table.js
```diff
@@ -42,18 +47,27 @@ const SummaryTable = ({
...
     ? rest.contentData || [data?.attributes?.analysis]
     : [];
 
+  useEffect(() => {
+    if (summaryMCLRef.current && isLoaded) {
+      getMCLFirstDataRow(summaryMCLRef.current).focus();
+    }
+  }, [summaryMCLRef, isLoaded]);
+
+  if (!data) {
```
BogdanDenis, line 67: Please move this block higher up just below `const summaryMCLRef = useRef(null);`. We don't need to call `useEffect` if we're not going to render anything

BogdanDenis (changes requested):

Vladyslav-Velytskyi (approved):

src/features/usage-consolidation-accordion/summary-table/summary-table.js
Yuliia-Dovhal: I also noticed, that this data check is the same as in `usage-consolidation-[entity-type].js`.

We have this part of code in that files: 
``` javascript
  const data = costPerUseData.data[costPerUseTypes.SOME_TYPE];

  if (!data) {
    return null;
  }
```
And the same check is in `SummaryTable`:

```javascript
const data = costPerUseData.data[costPerUseType];

  if (!data) {
    return null;
  }
```
`costPerUseType` is the same as in parent component, so it seems, like in `SummaryTable` data always will be. So I can get rid of this if statement in `SummaryTable`.

What do you think?

BogdanDenis: Ok, let's remove that check in `<SummaryTable`>

BogdanDenis (approved):

PR #1335, opened 2021-05-04
UIEH-1004: View Packages Record > Titles Accordion | Apply Load more results feature

Vladyslav-Velytskyi:
## Purpose
- Add load more button for loading next page for package titles
- Change page size for loading package titles from 25 items to 100
- Add tests

Issue: [UIEH-1004](https://issues.folio.org/browse/UIEH-1004)

## Approach
- It was decided to abandon the use of the model for obtaining package titles in favor of the reducer. This makes the code more maintainable and excludes the `State` from the library `impagination`.

## TO DO
- Add tests for `PackageShow` component and `PackageShow` route

## Screenshots
MxFHqoYDK6


src/components/query-search-list/query-search-list.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  const getLoadMoreButton = () => {
+    const isUnloadedPagesPresent = (totalLength !== length) && !(length % PAGE_SIZE);
+
+    if (isUnloadedPagesPresent) {
+      return (
+        <div className={styles['button-wrapper']}>
+          <Button
+            style={{ width: '75%', margin: '1rem auto' }}
```
BogdanDenis, line 64: Maybe add styles for this button in css?

BogdanDenis (changes requested):

src/components/query-search-list/query-search-list.js
Vladyslav-Velytskyi: I tried but class doesn't not present in layout. The same button in `MultiColumnList` component use style instead of class too.   Probably this is a issue of `Button` component.

BogdanDenis: Ok, in this case it's fine to keep like this

BogdanDenis (approved):

PR #1339, opened 2021-05-12
UIEH-1112: Add ability to configure APIGEE from UI

Vladyslav-Velytskyi:
## Purpose
* Add new fields for settings usage consolidation page.
* Add RTL/Jest tests for `SettingsUsageConsolidation` and `SettingsUsageConsolidationRoute` components.

Issue: [UIEH-1112](https://issues.folio.org/browse/UIEH-1112)

## Approach
When component get `stripes` from `useStripes()` developer cannot change stripes configuration. Because of this point we can't to test permissions in functional components. 

#### TODOS and Open Questions
We need to think about how to test `stripes` when component use `useState()`
Also route component does not get `match` prop from `withRoutes` HOC. Maybe it's need to be mocked to?

## Learning
If we want to test actions from `mapDispatchToProps` we need to mock them by 
```
jest.mock('../pathToActions', () => { action: mockAction });
```
And after pass this mock to the render function

## Screenshots
image


src/routes/settings-usage-consolidation-route.test.js
```diff
@@ -0,0 +1,293 @@
...
+          },
+        },
+      });
+
+      expect(container.querySelector('.icon-spinner-ellipsis')).toBeDefined();
+    });
+  });
+
+  describe('when usage consolidation is loaded', () => {
+    it('should handle getUsageConsolidationKey', async () => {
```
BogdanDenis, line 149: Previous and this test have same props, but are supposed to test different cases? Am I missing something?

BogdanDenis (approved):

BogdanDenis:
I don't get your comment about mocking `stripes` and `useStripes`. I see that you've added a mock to `stripesCore.mock` and you should also be able to do that from a single test file and configure stripes object however you want.
And which component did you mean when talking about `withRouter`?

src/routes/settings-usage-consolidation-route.test.js
Vladyslav-Velytskyi: In first test we use prop `isLoading` to see is spinner displayed and in the second we use `isLoaded` prop, when data loaded and we handle other action

Vladyslav-Velytskyi: In first test we use prop `isLoading` to see is spinner displayed and in the second we use `isLoaded` prop, when data loaded and we handle other action

Vladyslav-Velytskyi:
@BogdanDenis `withRoute` we use in route components (ex. `SettingsUsageConsolidationRoute`)

src/routes/settings-usage-consolidation-route.test.js
BogdanDenis: You're right, sorry, I didn't read that correctly

PR #1358, opened 2021-06-10
UIEH-1050: Jest+RTL > Test SearchPackageListItem, RootProxySelectField, TagsLabel, AddTitleToPackage

mariiaschv:
## Purpose
Added tests for SearchPackageListItem, RootProxySelectField, TagsLabel, AddTitleToPackage

Issue: [UIEH-1050](https://issues.folio.org/browse/UIEH-1050)

### DIFF
--- a/src/components/search-package-list-item/search-package-list-item.js
+++ b/src/components/search-package-list-item/search-package-list-item.js
@@ -112,7 +112,7 @@ SearchPackageListItem.propTypes = {
   link: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.object
-  ]),
+  ]).isRequired,
   onClick: PropTypes.func,
   packageName: PropTypes.string,
   showProviderName: PropTypes.bool,
--- a/src/components/search-package-list-item/search-package-list-item.js
+++ b/src/components/search-package-list-item/search-package-list-item.js
@@ -18,7 +18,7 @@ import styles from './search-package-list-item.css';
 
 const cx = classNames.bind(styles);
 
-function SearchPackageListItem({
+const SearchPackageListItem = ({
   item,
   link,
   active,
@@ -29,11 +29,12 @@ function SearchPackageListItem({
   showTitleCount,
   showSelectedCount,
   showTags,
-}) {
+}) => {
   return !item
     ? (
       <div
         className={cx('skeleton', 'is-provider-name-visible')}
+        data-testid="skeleton-element"
       />
     )
     : (
@@ -49,6 +50,7 @@ function SearchPackageListItem({
             onClick();
           }
         }}
+        data-testid="search-package-list-item-link"
       >
         <Headline
           data-test-eholdings-package-list-item-name
@@ -94,7 +96,19 @@ function SearchPackageListItem({
 SearchPackageListItem.propTypes = {
   active: PropTypes.bool,
   headingLevel: PropTypes.string,
-  item: PropTypes.object,
+  item: PropTypes.shape({
+    name: PropTypes.string,
+    providerName: PropTypes.string,
+    isSelected: PropTypes.bool,
+    selectedCount: PropTypes.number,
+    titleCount: PropTypes.number,
+    visibilityData: PropTypes.shape({
+      isHidden: PropTypes.bool,
+    }),
+    tags: PropTypes.shape({
+      tagList: PropTypes.arrayOf(PropTypes.string),
+    }),
+  }),
   link: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.object
--- a/src/components/search-package-list-item/search-package-list-item.js
+++ b/src/components/search-package-list-item/search-package-list-item.js
@@ -91,22 +91,22 @@ const SearchPackageListItem = ({
         </div>
       </InternalLink>
     );
-}
+};
 
 SearchPackageListItem.propTypes = {
   active: PropTypes.bool,
   headingLevel: PropTypes.string,
   item: PropTypes.shape({
+    isSelected: PropTypes.bool,
     name: PropTypes.string,
     providerName: PropTypes.string,
-    isSelected: PropTypes.bool,
     selectedCount: PropTypes.number,
+    tags: PropTypes.shape({
+      tagList: PropTypes.arrayOf(PropTypes.string),
+    }),
     titleCount: PropTypes.number,
     visibilityData: PropTypes.shape({
       isHidden: PropTypes.bool,
-    }),
-    tags: PropTypes.shape({
-      tagList: PropTypes.arrayOf(PropTypes.string),
     }),
   }),
   link: PropTypes.oneOfType([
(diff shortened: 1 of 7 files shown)


src/components/search-package-list-item/search-package-list-item.test.js
```diff
@@ -0,0 +1,118 @@
+import React from 'react';
+import { screen, render, fireEvent } from '@testing-library/react';
```
BogdanDenis, line 1: Please write each imported item on a new line

src/components/search-package-list-item/search-package-list-item.test.js
```diff
@@ -0,0 +1,118 @@
...
+      isHidden: true,
+    }
+  },
+  showProviderName: true,
+  showTitleCount: true,
+  showTags: true,
+};
+
+describe('Given SearchPackageListItem', () => {
+  const handleClick = jest.fn();
```
BogdanDenis, line 47: Please rename to `mockOnClick`

src/components/search-package-list-item/search-package-list-item.test.js
```diff
@@ -0,0 +1,118 @@
...
+  const renderSearchPackageListItem = ({ ...props }) => render(
+    <Router history={routerHistory}>
+      <SearchPackageListItem {...props} />
+    </Router>
+  );
+
+  it('should render skeleton when item is not provided', () => {
+    renderSearchPackageListItem(testEmptyProps);
+
+    const isSkeleton = document
```
BogdanDenis, line 58: Instead of querySelector you should use returned selectors from `renderSearchPackageListItem` and select items by their text or `data-testid` attribute
Here's docs about using RTL queries https://testing-library.com/docs/queries/about/

BogdanDenis (changes requested):

src/components/search-package-list-item/search-package-list-item.test.js
```diff
@@ -0,0 +1,117 @@
...
+
+    expect(isSkeleton).toBeDefined();
+  });
+
+  it('should render SearchPackageListItem link', () => {
+    renderSearchPackageListItem(testBasicProps);
+
+    const isSearchPackageListItemAttribute = document
+      .querySelector('.item')
+      .getAttribute('data-test-eholdings-package-list-item');
```
Vladyslav-Velytskyi, line 69: As @BogdanDenis mentioned, we should use `data-testid` for getting testing elements. Template `data-test-...` was used for testing with `bigtest` library. We don't delete this attributes yet, but we will possibly do it in future with removing `bigtest`

src/components/search-package-list-item/search-package-list-item.test.js
Vladyslav-Velytskyi: Please, take a look how tests wrote for other components and try to write them in the same template.

Vladyslav-Velytskyi (changes requested):
Also in test tickets we trying to refactoring components.
Could you, please, update some items
1. For `<AddTitleToPackage />` change `PropTypes` to
```
PropTypes.arrayOf(PropTypes.shape({
    disabled: PropTypes.bool.isRequired,
    key: PropTypes.string.isRequired,
    label: PropTypes.string.isRequired,
  })),
```
2. For `<RootProxySelectField />` add `ariaLabel` prop, it is equal for `label` but it need to wrote throw `intl.formatMessage`. Use `useIntl` hook for this. And update propTypes for it too.(they could be different as in `AddTitleToPackage`)
3. And the last one. Please, look on propTypes of `SeachPackageList`, probably the need updates too(ex. `item` props is not described)

src/components/settings/settings-root-proxy/_fields/root-proxy-select/root-proxy-select-field.js
```diff
@@ -30,16 +35,25 @@ function RootProxySelectField({ proxyTypes }) {
...
         {options}
       </Field>
     </div>
   );
-}
+};
 
 RootProxySelectField.propTypes = {
-  proxyTypes: PropTypes.object.isRequired
+  proxyTypes: PropTypes.shape({
```
Vladyslav-Velytskyi, line 49: This proptype is required

src/components/title/_field-groups/add-title-to-package.js
```diff
@@ -13,10 +13,14 @@ function AddTitleToPackage({ packageOptions }) {
...
-}
+};
 
 AddTitleToPackage.propTypes = {
-  packageOptions: PropTypes.array.isRequired
+  packageOptions: PropTypes.arrayOf(PropTypes.shape({
+    disabled: PropTypes.bool.isRequired,
+    key: PropTypes.string.isRequired,
+    label: PropTypes.string.isRequired,
+  })),
```
Vladyslav-Velytskyi, line 23:
```suggestion
  packageOptions: PropTypes.arrayOf(PropTypes.shape({
    disabled: PropTypes.bool.isRequired,
    key: PropTypes.string.isRequired,
    label: PropTypes.string.isRequired,
  })).isRequired,
```

src/components/search-package-list-item/search-package-list-item.js
```diff
@@ -89,12 +91,24 @@ function SearchPackageListItem({
...
+    providerName: PropTypes.string,
+    selectedCount: PropTypes.number,
+    tags: PropTypes.shape({
+      tagList: PropTypes.arrayOf(PropTypes.string),
+    }),
+    titleCount: PropTypes.number,
+    visibilityData: PropTypes.shape({
+      isHidden: PropTypes.bool,
+    }),
+  }),
```
Vladyslav-Velytskyi, line 111:
```suggestion
  item: PropTypes.shape({
    isSelected: PropTypes.bool.isRequired,
    name: PropTypes.string.isRequired,
    providerName: PropTypes.string,
    selectedCount: PropTypes.number.isRequired,
    tags: PropTypes.shape({
      tagList: PropTypes.arrayOf(PropTypes.string),
    }).isRequired,
    titleCount: PropTypes.number.isRequired,
    visibilityData: PropTypes.shape({
      isHidden: PropTypes.bool.isRequired,
    }).isRequired,
  }).isRequired,
```
Also, `link` property should be required

Vladyslav-Velytskyi (changes requested):

src/components/search-package-list-item/search-package-list-item.js
mariiaschv: Correct me if I'm wrong please, but how can `item` property be required if we check for its presence on line 33 and if it's not there we just render skeleton instead?

Vladyslav-Velytskyi: hm.. you are right, sorry, do not fix it. But it's strange, in all places where we are call this component we pull this prop and it's look like required. Probably we need to deep inside into this moment, but this is an another story.

Vladyslav-Velytskyi (approved):

PR #1370, opened 2021-07-02
UIEH-1140: View/Edit Managed title - Custom Package Record: Custom Package URL field does not display

BogdanDenis:
## Description
Display Managed url of a Resource only when a package is managed

## Issues
[UIEH-1140](https://issues.folio.org/browse/UIEH-1140)


Vladyslav-Velytskyi (approved):

mariiaschv (approved):

Vladyslav-Velytskyi:
@Denis, could you also add a test for this issue. We already have test file for this component, so, it will be better to cover it

PR #1386, opened 2021-07-16
UIEH-1143: Fix bug with root proxy selection

Vladyslav-Velytskyi:
# UIEH-1143: Root proxy selection is not saved

## Purpose
- Fix bug by removing case check
- Move `<ProxyDisplay>` component to separate folder
- Add tests for it
- Refactoring
- Add propTypes
- Change `proxyTypes` prop to `proxyTypesRecords` to omit a huge bunch of unnecessary values
- Fix tests for other components
 
Issue:  [UIEH-1143](https://issues.folio.org/browse/UIEH-1143)


mariiaschv (approved):

src/components/proxy-display/proxy-display.js
```diff
@@ -0,0 +1,62 @@
...
+  KeyValue,
+  Icon,
+} from '@folio/stripes/components';
+
+const ProxyDisplay = ({
+  proxy,
+  proxyTypesRecords,
+  inheritedProxyId,
+}) => {
+  if (proxyTypesRecords && proxy?.id && inheritedProxyId) {
```
BogdanDenis, line 14: I'd prefer to refactor this a bit more to use fail-fast
```
const showProxyInfo = proxyTypesRecords && proxy?.id && inheritedProxyId;

if (!showProxyInfo) {
  return <Icon icon="spinner-ellipsis" />;
}

// the rest goes here
```

It reduces nesting of important logic by one level and makes the component a bit easier to read, in my opinion

BogdanDenis (approved):

PR #1389, opened 2021-07-20
UIEH-1060: Jest+RTL > Test ProviderListItem

mariiaschv:
## Purpose

- Add tests for ProviderListItem component
- Refactoring

Issue: [UIEH-1060](https://issues.folio.org/browse/UIEH-1060)

### DIFF
--- a/src/components/provider-list-item/provider-list-item.js
+++ b/src/components/provider-list-item/provider-list-item.js
@@ -11,58 +11,70 @@ import styles from './provider-list-item.css';
 
 const cx = classNames.bind(styles);
 
-function ProviderListItem({ item, link, active, onClick, headingLevel }) {
-  return !item ? (
-    <div className={styles.skeleton} />
-  ) : (
-    <InternalLink
-      data-test-eholdings-provider-list-item
-      to={link}
-      className={cx('item', {
-        'is-selected': active
-      })}
-      onClick={(e) => {
-        if (onClick) {
-          e.preventDefault();
-          onClick();
-        }
-      }}
-    >
-      <Headline
-        data-test-eholdings-provider-list-item-name
-        margin="none"
-        size="medium"
-        tag={headingLevel || 'h3'}
-      >
-        {item.name}
-      </Headline>
-
+const ProviderListItem = ({
+  item,
+  link,
+  active,
+  onClick,
+  headingLevel,
+}) => {
+  return !item
+    ? (
       <div
-        data-test-eholdings-provider-list-item-selections
-        className={cx('itemMetadata')}
+        className={styles.skeleton}
+        data-testid="skeleton-element"
+      />
+    )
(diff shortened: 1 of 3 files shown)


src/components/provider-list-item/provider-list-item.js
```diff
@@ -11,68 +11,84 @@ import styles from './provider-list-item.css';
...
+            selectedCount={item.packagesSelected}
+            showSelectedCount
           />
-        </span>
-      </div>
-    </InternalLink>
-  );
-}
+
+          <span>
```
Vladyslav-Velytskyi, line 62: Do we need this `span`?

src/components/provider-list-item/provider-list-item.js
```diff
@@ -11,68 +11,84 @@ import styles from './provider-list-item.css';
...
-    PropTypes.object
+    PropTypes.object,
   ]),
-  onClick: PropTypes.func
+  onClick: PropTypes.func,
+};
+
+ProviderListItem.defaultProps = {
+  active: false,
 };
```
Vladyslav-Velytskyi, line 92: This component used just in `<ProviderSearchList>`, so it looks that almost all props are required
```
ProviderListItem.propTypes = {
  active: PropTypes.bool.isRequired,
  headingLevel: PropTypes.string,
  item: PropTypes.object,
  link: PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.shape({
      pathname: PropTypes.string,
  ]).isRequired,
  onClick: PropTypes.func.isRequired,
};
```

src/components/provider-list-item/provider-list-item.js
```diff
@@ -11,68 +11,84 @@ import styles from './provider-list-item.css';
...
-        <SelectedLabel
-          isSelected={item.packagesSelected}
-          selectedCount={item.packagesSelected}
-          showSelectedCount
-        />
+        <Headline
+          data-test-eholdings-provider-list-item-name
+          margin="none"
+          size="medium"
+          tag={headingLevel || 'h3'}
```
Vladyslav-Velytskyi, line 47: I think that it's better to add `headingLevel` to default props with value `h3` and you can left here just prop

src/components/provider-list-item/provider-list-item.js
mariiaschv: We can get rid of this `span` only if we also remove `span` around `FormattedNumber` a bit lower. Otherwise it breaks the display of the total packages label. So, should I remove both spans or just leave as it is?

Vladyslav-Velytskyi: Not now, because we need data-attribute here. Thank you for checking and describing it.

Vladyslav-Velytskyi (approved):

PR #1392, opened 2021-07-23
UIEH-1153: Add tests for EHoldings component

Vladyslav-Velytskyi:

### DIFF
--- a/src/components/eholdings/eholdings.test.js
+++ b/src/components/eholdings/eholdings.test.js
@@ -1,0 +1,63 @@
+import { render } from '@testing-library/react';
+import { noop } from 'lodash';
+
+import EHoldings from './eholdings';
+import Harness from '../../../test/jest/helpers/harness';
+
+jest.mock('../../routes/application', () => () => <span>ApplicationRoute</span>);
+jest.mock('../../routes/settings-route', () => () => <span>SettingsRoute</span>);
+jest.mock('../../routes/search', () => () => <span>SearchRoute</span>);
+jest.mock('../../routes/provider-show', () => () => <span>ProviderShow</span>);
+jest.mock('../../routes/provider-edit', () => () => <span>ProviderEdit</span>);
+jest.mock('../../routes/package-show', () => () => <span>PackageShow</span>);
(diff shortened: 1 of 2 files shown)


src/components/eholdings/eholdings.test.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+/*
+  describe('when visiting settings route', () => {
+    it('should not show EholdingsAppContext', () => {
+      const { queryByText } = renderEHoldings({ showSettings: true });
+
+      expect(queryByText('EHoldingsAppContext')).toBeNull();
+    });
+  });
+*/
```
mariiaschv, line 71: I believe that commented code should be removed, shouldn't it?

src/components/eholdings/eholdings.js
```diff
@@ -98,6 +98,7 @@ class EHoldings extends Component {
       match: { path: rootPath },
     } = this.props;
 
+
```
mariiaschv, line 101: Why do we need two empty lines here?

mariiaschv (changes requested):
Also, the changelog.md should probably be updated

src/components/eholdings/eholdings.js
Vladyslav-Velytskyi: It' need just to see on sonar coverage percentage. I'll remove it

src/components/eholdings/eholdings.test.js
Vladyslav-Velytskyi: You're right. I just don't understand why this simple test is failing. Without it all works well, but we got only 61% coverage. Not great, not terrible.

mariiaschv: Yeah, I got similar results when tried to test this component in scope of other ticket. It's not clear why tests on this component are failing. But still, we shouldn't leave commented pieces of code.

Vladyslav-Velytskyi: And it will be, because this is draft PR right now. I didn't find any reason why it's not working for sonar. Locally it works well.

mariiaschv (approved):

PR #1396, opened 2021-07-27
UIEH-1067: Jest+RTL > Test IdentifiersField

mariiaschv:
## Purpose

- Add tests for IdentifiersFields component
- Refactoring

Issue: [UIEH-1067](https://issues.folio.org/browse/UIEH-1067)


src/components/title/_fields/identifiers/identifiers-fields.js
```diff
@@ -13,96 +12,101 @@ import {
...
+      >
+        <FormattedMessage id="ui-eholdings.id">
+          {(fieldName) => (
+            <Field
+              name={`${identifier}.id`}
+              type="text"
+              component={TextField}
+              label={fieldName}
+              validate={validateId}
+              ariaLabel={fieldName}
```
Vladyslav-Velytskyi, line 73: Please use here `useIntl` to omit child function here

src/components/title/_fields/identifiers/identifiers-fields.js
```diff
@@ -13,96 +12,101 @@ import {
...
-          </Field>
-        </Col>
-        <Col
-          md
-          xs={12}
-          data-test-eholdings-identifiers-fields-id
+  return errors;
+};
+
+const renderField = (identifier) => {
```
Vladyslav-Velytskyi, line 31: I think it's better to make this function as a part of `IdentifiersField` component

src/components/title/_fields/identifiers/identifiers-fields.js
```diff
@@ -13,96 +12,101 @@ import {
...
 
-export default class IdentifiersFields extends Component {
-  validateId(value) {
-    let errors;
+import { IDENTIFIERS_FIELDS_VALUE_MAX_LENGTH } from '../../../../constants';
 
-    if (!value) {
-      errors = <FormattedMessage id="ui-eholdings.validate.errors.identifiers.noBlank" />;
-    }
+const validateId = (value) => {
```
Vladyslav-Velytskyi, line 17: Why you move this function outside the component?

Vladyslav-Velytskyi (changes requested):

src/components/title/_fields/identifiers/identifiers-fields.js
```diff
@@ -13,96 +12,101 @@ import {
...
+  let errors;
```
Vladyslav-Velytskyi, line 18: Better to rename it in `error` because in this function variable `error` get just single component as a result

Vladyslav-Velytskyi (changes requested):

src/components/title/_fields/identifiers/identifiers-fields.js
mariiaschv: Because there's no consistency in code: in some components functions are outside, in some the are inside the component. It's fixed now and I'll keep this in mind for the future, thanks.

Vladyslav-Velytskyi (approved):

PR #1411, opened 2021-08-10
UIEH-1163: Fix items count per page for PrevNextButtons component

Vladyslav-Velytskyi:
Fix bug described in the [comment](https://issues.folio.org/browse/UIEH-1163?focusedCommentId=109227&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-109227)

## Approach
Because fetch calls in different places the main propose to keep page value in global state.

## Screenshot
JoVUtzxqM8

### DIFF
--- a/src/components/query-search-list/query-search-list.js
+++ b/src/components/query-search-list/query-search-list.js
@@ -24,7 +24,7 @@ const QuerySearchList = ({
 }) => {
   useEffect(() => {
     fetch(FIRST_PAGE);
-  }, [fetch]);
+  }, []);
 
   const {
     totalResults,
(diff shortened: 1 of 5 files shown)


mariiaschv (approved):

src/components/query-search-list/query-search-list.js
```diff
@@ -23,18 +22,17 @@ const QuerySearchList = ({
   scrollable,
   type,
 }) => {
-  const [page, setPage] = useState(FIRST_PAGE);
-
   useEffect(() => {
-    fetch(page);
-  }, [fetch, page]);
+    fetch(FIRST_PAGE);
+  }, [fetch]);
```
BogdanDenis, line 27: This can cause issues if reference to `fetch` prop changes. It's better to leave dependencies array empty

BogdanDenis (approved):

PR #1417, opened 2021-08-16
UIEH-1078: Jest+RTL > Test SettingsRootProxy

mariiaschv:
## Purpose

- Add tests for SettingsRootProxy component
- Refactoring

Issue: [UIEH-1078](https://issues.folio.org/browse/UIEH-1078)

### DIFF
--- a/src/components/settings/settings-root-proxy/settings-root-proxy.test.js
+++ b/src/components/settings/settings-root-proxy/settings-root-proxy.test.js
@@ -45,32 +45,32 @@ const componentToRender = props => (
   </Harness>
 );
 
-const renderSettings = (props = {}) => render(
+const renderSettingsRootProxy = (props = {}) => render(
   componentToRender(props),
 );
 
 describe('Given SettingsRootProxy', () => {
   it('should render form', () => {
-    const { getByTestId } = renderSettings();
+    const { getByTestId } = renderSettingsRootProxy();
 
     expect(getByTestId('settings-root-proxy-form')).toBeDefined();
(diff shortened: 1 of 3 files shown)


src/components/settings/settings-root-proxy/settings-root-proxy.test.js
```diff
@@ -0,0 +1,166 @@
...
+const componentToRender = props => (
+  <Harness>
+    <SettingsRootProxy
+      onSubmit={() => {}}
+      proxyTypes={proxyTypes}
+      rootProxy={rootProxy}
+      {...props}
+    />
+  </Harness>
+);
```
Vladyslav-Velytskyi, line 46: Why you can not use it inside `renderSettingsRootProxy`? By the way, if you want to create a separate function, so please, it's naming should starts with a verb (for ex. `getComponent`, `getTestComponent` etc.)

mariiaschv: I can't use it inside `renderSettingsRootProxy` because I need to rerender the component in the tests, and `renderSettingsRootProxy` contains `render` function, which doesn't seem to make sense to put inside the `rerender`. So I moved the component itself to a separate function, so that it could be reused in `rerender` just by passing needed props.
Regarding naming, yes, my bad, I didn't notice that I forgot to fix it, sorry.

Vladyslav-Velytskyi (approved):

BogdanDenis (approved):

PR #1429, opened 2021-08-30
UIEH-1089: Add tests for SettingsKnowledgeBase component

Vladyslav-Velytskyi:
# UIEH-1089: Jest+RTL > Test SettingsKnowledgeBase

## Purpose
- Refactoring of `SettingsKnowledgeBase` component
- Add tests

Issue: [UIEH-1089](https://issues.folio.org/browse/UIEH-1089)

### DIFF
--- a/src/components/settings/settings-knowledge-base/settings-knowledge-base.js
+++ b/src/components/settings/settings-knowledge-base/settings-knowledge-base.js
@@ -301,7 +301,7 @@ const SettingsKnowledgeBase = ({
                     component={Select}
                     label={<FormattedMessage id="ui-eholdings.settings.kb.rmapiBaseUrl" />}
                   >
-                    {apiEndpointsOptions.map(({ value, translationId } ) => (
+                    {apiEndpointsOptions.map(({ value, translationId }) => (
                       <FormattedMessage
                         id={translationId}
                         values={{ url: value }}
--- a/src/components/settings/settings-knowledge-base/settings-knowledge-base.js
+++ b/src/components/settings/settings-knowledge-base/settings-knowledge-base.js
@@ -1,13 +1,16 @@
-import { Component } from 'react';
+import {
+  useEffect,
+  useState,
+  useRef,
+} from 'react';
 import PropTypes from 'prop-types';
 import { Field, Form } from 'react-final-form';
 import createFocusDecorator from 'final-form-focus';
 import {
   FormattedMessage,
-  injectIntl,
+  useIntl,
 } from 'react-intl';
-import { withRouter } from 'react-router';
-import ReactRouterPropTypes from 'react-router-prop-types';
+import { useHistory } from 'react-router';
 
 import { IfPermission } from '@folio/stripes/core';
 import SafeHTMLMessage from '@folio/react-intl-safe-html';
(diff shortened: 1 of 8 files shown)


src/components/settings/settings-knowledge-base/settings-knowledge-base.js
```diff
@@ -25,106 +28,125 @@ import NavigationModal from '../../navigation-modal';
...
+  onDelete,
+  onSubmit,
+}) => {
+  const intl = useIntl();
+  const history = useHistory();
+
+  const [toasts, setToasts] = useState([]);
+  const [deleteConfirmationModalDisplayed, setDeleteConfirmationModalDisplayed] = useState(false);
+
+  useEffect(() => {
```
BogdanDenis, line 63: I think it's better to split a single `useEffect` into several smaller ones, and instead of passing whole `kbCredentials` object as a dependency pass only `kbCredentials.hasUpdated`. This will guard against cases when `kbCredentials` reference has changed, but the property that we are checking is still the same, but code inside `if` will run twice anyway.

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
```diff
@@ -25,106 +28,125 @@ import NavigationModal from '../../navigation-modal';
...
+              values={{ name: config.attributes.name }}
+            />
+          ),
+          type: 'success',
+        },
+      ]);
     }
+  }, [kbCredentials]);
+
+  const prevErrors = useRef(kbCredentials.errors).current;
```
BogdanDenis, line 108: Isn't just using `useEffect` with `kbCredentials.errors` dependency enough to know that errors changed?

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
```diff
@@ -25,106 +28,125 @@ import NavigationModal from '../../navigation-modal';
...
+
+    prevErrors.current = kbCredentials.errors;
+  }, [kbCredentials.errors]);
+
+  if (!config) {
+    return null;
   }
 
-  validateNameField = value => {
+  const validateNameField = value => {
```
BogdanDenis, line 129: Please use parentheses for this function argument

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
```diff
@@ -156,187 +178,176 @@ class SettingsKnowledgeBase extends Component {
...
+                    data-testid="kb-name-field"
+                  />
+                </div>
+                <div data-test-eholdings-settings-kb-url>
+                  <Field
+                    name="url"
+                    component={Select}
+                    label={<FormattedMessage id="ui-eholdings.settings.kb.rmapiBaseUrl" />}
+                  >
+                    <option value="https://sandbox.ebsco.io">Sandbox: https://sandbox.ebsco.io</option>
```
BogdanDenis, line 299: Let's move these two options into constants with label and value

BogdanDenis (changes requested):

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
Vladyslav-Velytskyi: No, it isn't. We need `prevProps` to make it work as expected. Without `useRef` and using just `useEffect` we will get this behavior. We could try to make another action to clear errors when component will destroyed, but I think that this is out of scope of this task. 
HhsOxkQ4s3

mariiaschv (approved):

src/components/settings/settings-knowledge-base/settings-knowledge-base.js
BogdanDenis: I see, ok, let's keep it this way

BogdanDenis (approved):

PR #1432, opened 2021-09-06
UIEH-1171: Package Detail Record: When changing holding status, list of titles accordion does not reflect new holdings status UNLESS one refreshes page

mariiaschv:
## Purpose
Fixed when changing holding status, list of titles accordion does not reflect new holdings status UNLESS one refreshes page.

### TODO
Tests will be done in scope of [UIEH-1188](https://issues.folio.org/browse/UIEH-1188)

## Screenshots
qW76jgpnTo

## Issues:
[UIEH-1171](https://issues.folio.org/browse/UIEH-1171)


src/routes/package-show.js
```diff
@@ -173,21 +181,80 @@ class PackageShowRoute extends Component {
...
+      return { isTitlesUpdating: true };
+    });
+
+    this.timeout = window.setTimeout(() => {
+      getPackageTitles({ packageId, params });
+
+      this.setState(() => {
+        return { isTitlesUpdating: false };
+      });
+    }, 6000);
```
Vladyslav-Velytskyi, line 222: It will be better to have  a constant, but why it is 6 seconds?

Vladyslav-Velytskyi (changes requested):

src/routes/package-show.js
mariiaschv: Because for the less time there are often cases when data in the database isn't updated yet and request returns old values.

BogdanDenis:
Looks good, just need some tests to cover this feature

mariiaschv:
> Looks good, just need some tests to cover this feature

I added tests for PackageShow to cover this and tests for PackageShowRoute are going to be written in scope of [UIEH-1188](https://issues.folio.org/browse/UIEH-1188)

BogdanDenis (approved):

Vladyslav-Velytskyi (approved):

PR #1433, opened 2021-09-06
UIEH-1175: eHoldings - Removing Impagination dependency.

BogdanDenis:
## Description
The aim of this PR is to remove `impagination` dependency from `ui-eholdings` by replacing it with a custom component that will replicate the behaviour.
Added `<ImpaginationReplacement>` component that provides the same API as `<Impagination>`. Still we need to refactor `<QueryList>` and `<QuerySearchList>` components and combine them into one, and move `<ImpaginationReplacement>` logic into a hook. But this needs to be additionally discussed


src/components/scroll-view/scroll-view.js
```diff
@@ -181,36 +160,16 @@ export default class ScrollView extends Component {
...
-        : PAGE_SIZE;
-    }
-
-    // list height should be at least enough for the offset
-    if (listHeight === 0) {
-      listHeight = offset + visibleItems;
-    }
-
-    listHeight *= itemHeight;
+    const listHeight = (length || items.length) * itemHeight;
```
Vladyslav-Velytskyi, line 166: What if `length` will be more then 100?  Isn't list will have empty space?

src/components/scroll-view/scroll-view.js
```diff
@@ -181,36 +160,16 @@ export default class ScrollView extends Component {
...
 
     return (
       <div
         ref={(n) => { this.$list = n; }}
         className={cx('list', { locked: !scrollable })}
-        onScroll={this.handleScroll}
+        // onScroll={this.handleScroll}
```
mariiaschv, line 172: Do we need this comment here?

src/components/scroll-view/scroll-view.js
BogdanDenis: This prop wasn't actually passed to the component from anywhere so I removed it

Vladyslav-Velytskyi: Ok, but if `items.length` will be more then 100?

Vladyslav-Velytskyi (approved):

mariiaschv (approved):

PR #1450, opened 2021-10-04
UIEH-1193: Add tests for SettingsAccessStatusTypesRoute component

Vladyslav-Velytskyi:
# UIEH-1193 Jest+RTL > Test SettingsAccessStatusTypesRoute

## Purpose
Add tests for `SettingsAccessStatusTypesRoute`

Issue: [UIEH-1193](https://issues.folio.org/browse/UIEH-1193)

## Screenshots
image


src/routes/settings-access-status-types-route/settings-access-status-types-route.js
```diff
@@ -48,38 +49,33 @@ const SettingsAccessStatusTypesRoute = ({
     getAccessTypes(params.kbId);
```
BogdanDenis, line 49: For some reason sonar shows that both `useEffect` are covered, but there are no tests. Please add two tests: to check that `getAccessTypes` is called only once when component is mounted and to check that `getAccessTypes` is called only when `params.kbId` changes

src/routes/settings-access-status-types-route/index.js
```diff
@@ -0,0 +1,25 @@
+
```
mariiaschv, line 1: Please, remove unnecessary empty line

mariiaschv (approved):

PR #1469, opened 2021-11-04
UIEH-1212: Refactor '.all' permissions

Vladyslav-Velytskyi:
# UIEH-1212 refactor psets away from backend ".all" permissions

## Purpose
Replace `.all` permissions

[UIEH-1212](https://issues.folio.org/browse/UIEH-1212)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Add tests for SettingsKnowledgeBaseRoute component. (UIEH-1196)
 * Add tests for TitleCreateRoute component. (UIEH-1199)
 * Add tests for TitleEditRoute component. (UIEH-1200)
-
+* Refactor '.all' permissions. (UIEH-1212)
 
 ## [7.0.1] (https://github.com/folio-org/ui-eholdings/tree/v7.0.1) (2021-10-19)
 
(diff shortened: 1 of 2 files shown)


mariiaschv (approved):

BogdanDenis:
Looks good!
I have only one thing to ask: it'll be a good idea to look at all endpoints in module descriptor and for each endpoint compare required permissions and if UI permissions meet the requirement.
I'd like to make sure that this is the case because recently we found out that we didn't have following permissions:
```
"kb-ebsco.package-tags.put",
"kb-ebsco.provider-tags.put",
"kb-ebsco.resource-tags.put"
```
and it makes me think maybe there's something else we're missing

Vladyslav-Velytskyi:
> Looks good! I have only one thing to ask: it'll be a good idea to look at all endpoints in module descriptor and for each endpoint compare required permissions and if UI permissions meet the requirement. I'd like to make sure that this is the case because recently we found out that we didn't have following permissions:
> 
> ```
> "kb-ebsco.package-tags.put",
> "kb-ebsco.provider-tags.put",
> "kb-ebsco.resource-tags.put"
> ```
> 
> and it makes me think maybe there's something else we're missing

I think that we need a separate task for it. It looks like that we don't have any permissions for a lot of endpoint, so it's need to check each endpoint with all permissions on environment. It will be more complicated task

BogdanDenis (approved):

PR #1473, opened 2021-11-10
Fix syntax errors in Korean translation file

BogdanDenis:
## Description
Fix syntax errors in Korean translation file. Specifically, some words in `{}` are keys that are used in UI code to pass parameters to `intl`. Users won't see them and they need to stay as they are.


Vladyslav-Velytskyi (approved):

mariiaschv (approved):

translations/ui-eholdings/ko.json
```diff
@@ -29,7 +29,7 @@
     "package.packageSettings": "패키지 설정",
     "package.packageSettings.notSelected": "사용자에게 표시되지 않음",
     "package.packageType": "패키지 유형",
-    "package.partiallySelected": "{ selectedCount } titleCount } 타이틀 중 { titleCount } 개 선택됨",
+    "package.partiallySelected": "{ selectedCount } { titleCount } 타이틀 중 { titleCount } 개 선택됨",
```
Vladyslav-Velytskyi, line 32: Between `selectedCount` and `titleCount` present 'of' word, it seems that it should be translated too. How it should be resolved?

Vladyslav-Velytskyi (approved):

translations/ui-eholdings/ko.json
BogdanDenis: @Vladyslav-Velytskyi I don't speak Korean unfortunately) I think I'll add it in English and someone can translate it later

PR #1477, opened 2021-11-24
UIEH-1184 Add Application tests

BogdanDenis:
## Description
Add tests for `<ApplicationRoute>` component

## Issues
[UIEH-1184](https://issues.folio.org/browse/UIEH-1184)


mariiaschv (approved):
Maybe it's better to rename `application.js` to `application-route.js` and  `application.test.js` to  `application-route.test.js` so that we get rid of one code smell?

Vladyslav-Velytskyi (approved):

PR #1486, opened 2022-01-04
UIEH-1186: Added tests for PackageCreateRoute

Svitlana-Korol:
https://issues.folio.org/browse/UIEH-1186


src/routes/package-create-route/package-create.test.js
```diff
@@ -0,0 +1,152 @@
...
+    mockGetAccessTypes.mockClear();
+    mockCreatePackage.mockClear();
+    mockRemoveCreateRequests.mockClear();
+    mockHistory.replace.mockClear();
+  });
+
+  afterEach(cleanup);
+
+  it('should handle getAccessTypes', async () => {
+    await act(async () => {
```
BogdanDenis, line 78: There's no need to add `act` everywhere. `render`, `fireEvent` and other utils are already using `act` under the hood.
Here's an article on this https://javascript.plainenglish.io/you-probably-dont-need-act-in-your-react-tests-2a0bcd2ad65c

BogdanDenis (changes requested):

src/routes/package-create-route/package-create.test.js
```diff
@@ -0,0 +1,152 @@
...
+      await renderPackageCreateRoute({
+        props:{ getAccessTypes: mockGetAccessTypes },
```
mariiaschv, line 80: Please, don't forget spaces after keys in objects:
```suggestion
        props: { getAccessTypes: mockGetAccessTypes },
```

mariiaschv (changes requested):

mariiaschv (changes requested):
When file exports only one function it should be named after that function. So, it's better to rename `package-create.js` file to `package-create-route.js`. The same applies to the test file.

Also, we should strive for at least 80% of code coverage, so please try to add more test cases if possible.

src/routes/package-create-route/package-create-route.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+      fireEvent.click(getByRole('button', { name: 'ui-eholdings.label.icon.closeX' }));
+
+      expect(mockHistory.goBack).toHaveBeenCalled();
+    });
+  });
+
+  describe('when submit form with some values and click save', () => {
+    it('should handle mockCreatePackage action', async () => {
+      const { getByRole, getByText, getByTestId } = renderPackageCreateRoute();
```
mariiaschv, line 142: When we destructure more than one element of an object or an array it's better to write each variable on the separate line so that the code would be more readable:
```suggestion
      const {
        getByRole,
        getByText,
        getByTestId,
      } = renderPackageCreateRoute();
```

src/routes/package-create-route/package-create-route.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+      const packageNameInput = getByRole('textbox', { name: 'ui-eholdings.label.name' });
+
+      fireEvent.change(packageNameInput, { target: { value: 'New package name' } });
+      fireEvent.blur(packageNameInput);
+
+      const addCoverageSettingsButton = getByText('ui-eholdings.package.coverage.addDateRange');
```
mariiaschv, line 149: Please, add an empty line after the `const` variable declaration.

src/routes/package-create-route/package-create-route.test.js
```diff
@@ -0,0 +1,162 @@
...
+      fireEvent.click(addCoverageSettingsButton);
+
+      fireEvent.change(getByTestId('begin-coverage-0'), { target: { value: '01/01/2022' } });
+      fireEvent.change(getByTestId('end-coverage-0'), { target: { value: '10/01/2022' } });
+      getByTestId('begin-coverage-0').blur();
+      getByTestId('end-coverage-0').blur();
```
mariiaschv, line 155: Maybe it's better to use `fireEvent.blur()` here, since you use it in the test above, to keep the code consistent?
```suggestion
      fireEvent.blur(getByTestId('begin-coverage-0'));
      fireEvent.blur(getByTestId('end-coverage-0'));
```

mariiaschv (approved):
Approved! Just some minor issues left to fix.

PR #1491, opened 2022-01-28
UIEH-1189: Add tests for ProviderEditRoute

Svitlana-Korol:
## Description
Add tests for ProviderEditRoute

## Issues
[UIEH-1189](https://issues.folio.org/browse/UIEH-1189)


src/routes/provider-edit-route/provider-edit-route.test.js
```diff
@@ -0,0 +1,269 @@
...
+      />
+        Page content
+    </Harness>
+  </MemoryRouter>
+);
+
+const renderProviderEditRoute = (props) => render(getProviderEditRoute(props));
+describe('Given ProviderEditRoute', () => {
+  beforeEach(() => {
+    mockGetProxyTypes.mockClear();
```
BogdanDenis, line 134: Tip: you can clear all mocks with just one function call - `jest.clearAllMocks()`

src/routes/provider-edit-route/provider-edit-route.test.js
```diff
@@ -0,0 +1,269 @@
...
+  afterEach(cleanup);
+
+  it('should render page', () => {
+    const { getByText } = renderProviderEditRoute({});
+
+    expect(getByText('Page content')).toBeDefined();
+  });
+
+  it('should handle getProvider with providerId', async () => {
+    await act(async () => {
```
BogdanDenis, line 150: There's no need to call `act` here because there isn't an asynchronous event handling

src/routes/provider-edit-route/provider-edit-route.test.js
```diff
@@ -0,0 +1,269 @@
...
+      await renderProviderEditRoute();
+    });
+
+    expect(mockGetProvider).toHaveBeenCalledWith('provider-id');
+  });
+
+  it('should handle getProvider', async () => {
+    await act(async () => {
```
BogdanDenis, line 158: And here too, please check all tests

src/routes/provider-edit-route/provider-edit-route.test.js
```diff
@@ -0,0 +1,269 @@
...
+      });
+    });
+
+    expect(mockGetProxyTypes).toHaveBeenCalled();
+  });
+
+  it('should handle GetProxyTypes', async () => {
+    await act(async () => {
+      await renderProviderEditRoute({
+        props: { proxyType : mockGetProxyTypes },
```
BogdanDenis, line 186: Here's two mistakes:
1. `renderProviderEditRoute` expects an object with properties, like: `{ proxyType, ... }`, but you're passing `{ props: { proxyType } }`, so this has no effect on results
2. there's no need to pass this at all because this property is already passed at line 114, and you're not overriding anything to test a specific case

src/routes/provider-edit-route/provider-edit-route.test.js
```diff
@@ -0,0 +1,269 @@
...
+          getProvider: mockGetProvider,
+          match: {
+            ...match,
+            params: { providerId: 'other-provider-id' },
+          },
+        },
+      });
+    });
+
+    expect(mockGetProvider).toHaveBeenCalled();
```
BogdanDenis, line 170: Please also test that this prop is called with correct arguments.
Sometimes it's very important to test this, for example when requesting data from backend we need to make sure that the right data is requested, or when a child calls a prop with some data to pass to parent - we also need to make sure that data is correct, or when redirecting to a new page
If you check `expect(mockGetProvider).toHaveBeenCalledWith('other-provider-id');` it'll fail because of how you've passed props to component - see my next comment about this

src/routes/provider-edit-route/provider-edit-route.test.js
```diff
@@ -0,0 +1,269 @@
...
+    });
+
+    expect(mockGetProxyTypes).toHaveBeenCalled();
+  });
+
+  it('should handle Cancel', async () => {
+    let getByRole;
+
+    await act(async () => {
+      getByRole = await renderProviderEditRoute({}).getByRole;
```
BogdanDenis, line 197: No need to pass `{}` here because you already have a default value for this argument

src/routes/provider-edit-route/provider-edit-route.test.js
```diff
@@ -0,0 +1,269 @@
...
+
+    await act(async () => {
+      getByRole = await renderProviderEditRoute().getByRole;
+    });
+
+    fireEvent.submit(getByRole('button', { name: 'stripes-components.saveAndClose' }));
+
+    expect(mockUpdateProvider).toHaveBeenCalled();
+  });
+  describe('when update request resolves', () => {
```
BogdanDenis, line 222: Please add empty lines between `describe`

BogdanDenis (changes requested):

src/routes/provider-edit-route/provider-edit-route.test.js
Svitlana-Korol: Good idea, thanks!

src/routes/provider-edit-route/provider-edit-route.test.js
Svitlana-Korol: You are right, I fixed it.

src/routes/provider-edit-route/provider-edit-route.test.js
Svitlana-Korol: You are right, I fixed it.

src/routes/provider-edit-route/provider-edit-route.test.js
Svitlana-Korol: Thanks, I fixed it.

src/routes/provider-edit-route/provider-edit-route.test.js
```diff
@@ -0,0 +1,231 @@
...
+        rootProxy={rootProxy}
+        updateProvider={mockUpdateProvider}
+        {...props}
+      />
+        Page content
+    </Harness>
+  </MemoryRouter>
+);
+
+const renderProviderEditRoute = (props) => render(getProviderEditRoute(props));
```
mariiaschv, line 130: There should be an empty line too.

src/routes/provider-edit-route/provider-edit-route.test.js
```diff
@@ -0,0 +1,231 @@
...
+  afterEach(cleanup);
+
+  it('should render page', () => {
+    const { getByText } = renderProviderEditRoute();
+
+    expect(getByText('Page content')).toBeDefined();
+  });
+
+  it('should handle getProvider with providerId', async () => {
+    await renderProviderEditRoute();
```
mariiaschv, line 145: Do we need to `await` rendering here? Please check this and other tests too, and if it's not needed then remove `async` & `await` keywords.

mariiaschv (changes requested):

BogdanDenis (approved):

mariiaschv (approved):

PR #1492, opened 2022-01-30
MODKBEKBJ-619: Fix 3 GET requests instead of 1 are made when searching Providers/Packages/Titles

mariiaschv:
## Purpose
Fix 3 GET requests instead of 1 are made when searching Providers/Packages/Titles.

## Approach
The reason 3 GET requests were made is `useImpagination` custom hook where two `useEffect`s were called not only when values in dependencies array changed but also on mount, which was not needed.

The solution refers to [React docs](https://reactjs.org/docs/hooks-faq.html#can-i-run-an-effect-only-on-updates) where it's suggested to add a mutable ref to check if this is the first mount or not.

## Screenshots
6S8K4WDThr

## Issues
[MODKBEKBJ-619](https://issues.folio.org/browse/MODKBEKBJ-619)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Add tests for PackageCreateRoute. (UIEH-1186)
 * Add tests for PackageEditRoute. (UIEH-1187)
 * Lock `faker` version.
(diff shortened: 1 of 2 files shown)


BogdanDenis (approved):
Good catch!

Svitlana-Korol (approved):

PR #1496, opened 2022-02-02
UIEH-1227: Different date format in eHoldings resource coverage dates

mariiaschv:
## Purpose
Fix datepicker date format for Custom coverage dates to match current user locale when edit resource.

## Approach
The date format was forced to be `MM/DD/YYYY` by passing `DATE_FORMAT` constant with this value to `<Datepicker>` component. Removing this prop allows to format date based on current locale.

### Date formatting in datepicker for en-US locale:
image

### Date formatting in datepicker for en-GB locale:
image

## Issues
[UIEH-1227](https://issues.folio.org/browse/UIEH-1227)


BogdanDenis (approved):
Looks good, please also apply these changes to date pickers in `<PackageCoverageFields>` component and we can merge it

Svitlana-Korol (approved):

PR #1498, opened 2022-02-08
UIEH-1183:Jest+RTL > Test useMultiColumnListSort

Svitlana-Korol:
##  Description
Add tests for useMultiColumnListSort

## Issues
[UIEH-1183](https://issues.folio.org/browse/UIEH-1183)


src/hooks/useMultiColumnListSort.test.js
```diff
@@ -0,0 +1,62 @@
...
+        sortedColumn: 'newName',
+      };
+      const { result } = renderHook(() => useMultiColumnListSort(currentParameters.sortOrder, currentParameters.sortedColumn));
+      const [sortParameters, onChange] = result.current;
+
+      act(() => {
+        onChange('', { name: 'newName' });
+      });
+
+      expect(sortParameters.sortOrder).toEqual(sortOrders.asc);
```
BogdanDenis, line 39: Looking at code this hook should change sort from `asc` to `desc` and vice versa if `onChanged` is called with same name parameter, but this test expects to get the same `sortOrder` as before `onChange` and it passes.
The issues is that after `onChange` we need to get updated data in `result`. This can be done by awaiting `waitForNextUpdate`
```
const { result, waitForNextUpdate } = renderHook(...);
...
act(() => {
  onChange(...);
});

await waitForNextUpdate();

expect(result.current[0].sortOrder).toEqual(sortOrders.desc);
```

src/hooks/useMultiColumnListSort.test.js
```diff
@@ -0,0 +1,62 @@
...
+      expect(sortParameters.sortedColumn).toBe('newName');
+    });
+
+    it('should return descending sortOrder and selected sortedColumn', () => {
```
BogdanDenis, line 43: There are two cases we should check:
1. `onChange` is called with the same `name` as `currentColumn`. Then `sortOrder` should switch from `asc` to `desc` or vice versa
2. `onChange` is called with a different `name` as `currentColumn`. Then `sortOrder` should be set to default value `asc` and `currentColumn` should be equal to `name`

This test is equivalent to the previous one - they both check the first case. Please change this test to call `onChange` with a different name and check that correct default `sortOrder` is set

BogdanDenis (changes requested):

BogdanDenis (approved):

mariiaschv (approved):

PR #1499, opened 2022-02-09
UIEH-1235: Multiple requests to /resources sent when viewing Titles list on Package Show page

mariiaschv:
## Purpose
Fix multiple requests to /resources sent when viewing Titles list on Package Show page.

## Approach
This issue was caused due to incorrect object comparison in `componentDidUpdate` lifecycle method. As objects are reference types they cannot be checked for equality with `===` operator. So instead `isEqual` function from Lodash library was used.

This incorrect comparison led to dispatching a `getPackageTitlesAction` when it was not really needed and that caused excess requests being sent.

## Screenshots
wJVOaaei45

## Issues
[UIEH-1235](https://issues.folio.org/browse/UIEH-1235)


Svitlana-Korol (approved):

BogdanDenis (approved):
Good catch!

PR #1501, opened 2022-02-11
UIEH-1190:Jest+RTL > Test ResourceEditRoute

Svitlana-Korol:
## Description
Added test  for ResourceEditRoute

## Issues
[[UIEH-1190](https://issues.folio.org/browse/UIEH-1190)

### DIFF
--- a/src/routes/resource-edit-route/resource-edit-route.test.js
+++ b/src/routes/resource-edit-route/resource-edit-route.test.js
@@ -1,0 +1,297 @@
+import { MemoryRouter } from 'react-router-dom';
+
+import {
+  render,
+  cleanup,
+  fireEvent,
+} from '@testing-library/react';
+
+import ResourceEditRoute from './resource-edit-route';
+import Harness from '../../../test/jest/helpers/harness';
+
+const history = {
+  replace: jest.fn(),
+};
+
+const location = {
+  pathname: 'pathname',
+  search: '',
+  hash: '',
+};
+
+const provider = {
+  id: 'providerid',
+  providerToken: {
+    value: 'provider-token',
+    prompt: '',
+  },
+  data: {
+    isLoaded: true,
+    isLoading: false,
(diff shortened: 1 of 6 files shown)


src/routes/resource-edit-route/resource-edit-route.test.js
```diff
@@ -0,0 +1,333 @@
...
+  },
+};
+
+const id = `${provider.id}-titleid`;
+
+const match = {
+  isExact: true,
+  params: {
+    id,
+  },
```
mariiaschv, line 43: It could be written in one line
```suggestion
  params: { id },
```

src/routes/resource-edit-route/resource-edit-route.test.js
```diff
@@ -0,0 +1,333 @@
...
+
+      expect(mockUpdateResource).toHaveBeenCalledWith({
+        ...model,
+        isSelected: true,
+      });
+    });
+  });
+
+  describe('when a managed package is deselected', () => {
+    it('should call destroyResource', () => {
```
mariiaschv, line 319: You have here 'should call **destroyResource**' but in `expect` you check if mock**UpdateResource** has been called. Please recheck and change it accordingly.

mariiaschv (changes requested):
Please add empty lines before/after constant variables declarations where needed (for instance, in `resource-edit-route.js` after lines 31 and 33, 58 and so on). And put destructured values on separate lines (take a look at lines 32, 40-41, 72 etc. in the same file).

src/routes/resource-edit-route/resource-edit-route.test.js
Svitlana-Korol: You are right, thanks!

mariiaschv (approved):

BogdanDenis (approved):

PR #1508, opened 2022-02-17
UIEH-1242 correctly specify list's proptypes

zburke:
This looks like a typo since proptypes were specified twice, and the
second (invalid) settings look like values, not value-types.

Refs [UIEH-1242](https://issues.folio.org/browse/UIEH-1242)


BogdanDenis (approved):
Thanks for fixing this!

mariiaschv (approved):

Svitlana-Korol (approved):

PR #1384, opened 2021-07-15
UIEH-1054: Jest+RTL > Test InternalLink, ProxyDisplay, shouldFocus, TokenDisplay, eHoldings

mariiaschv:
## Purpose

- Add tests for InternalLink, shouldFocus, TokenDisplay components
- Refactoring

Tests for `ProxyDisplay` were written in another [PR](https://github.com/folio-org/ui-eholdings/pull/1386)
For testing `EHoldings` component there was created a separate ticket - [UIEH-1153](https://issues.folio.org/browse/UIEH-1153)

Because of changes in `eholdings.js` and absence of the test file the total coverage is less than expected. The coverage on components other than EHoldings is showing 100%.

image

## Issue
[UIEH-1054](https://issues.folio.org/browse/UIEH-1054)

PR #1516, opened 2022-02-28
UIEH-1244: Turn on lazy loading in eholdings

mkuklis:
https://issues.folio.org/browse/UIEH-1244

@mariiaschv we are in the process of introducing lazy loading for all UI modules under https://github.com/folio-org/stripes-webpack/pull/42. 

Most modules work fine but we noticed that eholdings is not working correctly and it gets into an infinitive re-rendering loop because of the code in this PR.

It looks like it was added here: https://github.com/folio-org/ui-eholdings/commit/ddc7c645ea6a54d02456a8789f40e0f7bad470c6

Do you think we could safely remove it? It looks like `history.replace` just reloads the same path here.


zburke (approved):
Agreed, this LGTM but it would be really nice to have a better understanding of why this code was introduced in #1384 since it is a bit peculiar and there is no discussion about it either in the PR, comments, or ticket.

mariiaschv (approved):
@mkuklis sorry, that it took a long time for me to respond.
This piece of code actually wasn't added in [ddc7c64](https://github.com/folio-org/ui-eholdings/commit/ddc7c645ea6a54d02456a8789f40e0f7bad470c6), it existed before in file [src/index.js](https://github.com/folio-org/ui-eholdings/commit/ddc7c645ea6a54d02456a8789f40e0f7bad470c6#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556) and in that PR it was just moved to `src/components/eholdings/eholdings.js`. So I don't know for sure why it was introduced in the first place, my guess that it was needed to clear the `location.state`. But it seems to work fine without this code, so from my point of view it can be removed safely.

mkuklis:
@mariiaschv thank you so much for getting back to us and providing more details. I will run couple more tests but if the app is working as expected I plan to merge this PR because it's currently blocking the work related to lazy loading.

PR #1529, opened 2022-04-04
UIEH-1265: Replace or remove react-hot-loader

Svitlana-Korol:
## Description
Remove react-hot-loader

## Issues
[UIEH-1265](https://issues.folio.org/browse/UIEH-1265)

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,6 @@
     "prop-types": "^15.6.2",
     "react-final-form": "^6.3.0",
     "react-final-form-arrays": "^3.1.0",
-    "react-hot-loader": "^4.3.12",
     "react-router-prop-types": "^1.0.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -85,7 +85,6 @@
     "prop-types": "^15.6.2",
     "react-final-form": "^6.3.0",
     "react-final-form-arrays": "^3.1.0",
-    "react-hot-loader": "^4.3.12",
```
BogdanDenis, line 88: This dep is used in `src/index`, so we should also remove it's use there

BogdanDenis (changes requested):

BogdanDenis (approved):

mariiaschv (approved):

PR #1566, opened 2022-06-20
UIEH-1169: eholdings settings > Usage Consolidation > Add Show/Hide t…

ODanyor:
## Purpose
Add Show/Hide toggle into: Usage consolidation ID, Usage consolidation client ID, Usage consolidation API key.

## Approach
Field components are replaced with ShowHidePasswordField component.


BogdanDenis:
Right now if you click the "Show password" button it will show "***". This is a placeholder and we need to show actual values
For that you need to make two requests: `/eholdings/uc-credentials/clientId` and `/eholdings/uc-credentials/clientSecret` to get real values and pass them to form's `initialValues`
Then the `<ShowHidePasswordField>` should handle switching between showing a placeholder and a real value

Svitlana-Korol (approved):

PR #1602, opened 2022-10-12
UIEH-1332: Fix tests caused by third party library updates

Dmytro-Melnyshyn:
## Purpose
Fix tests caused by third party library updates

### DIFF
--- a/src/components/settings/settings-custom-labels/settings-custom-labels.test.js
+++ b/src/components/settings/settings-custom-labels/settings-custom-labels.test.js
@@ -153,11 +153,14 @@ describe('Given SettingsCustomLabels', () => {
         getByRole,
         getByTestId,
       } = renderSettingsCustomLabels();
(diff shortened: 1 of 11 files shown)


src/components/settings/settings-custom-labels/settings-custom-labels.test.js
```diff
@@ -153,11 +153,14 @@ describe('Given SettingsCustomLabels', () => {
         getByRole,
         getByTestId,
       } = renderSettingsCustomLabels();
+      const name = 'ui-eholdings.settings.customLabels.remove ' +
```
BogdanDenis, line 156: Why do we need to join three translations together? What element will it match?

Dmytro-Melnyshyn: `div` with `role='dialog'`, the debugger reveals this name.

BogdanDenis (approved):

PR #1626, opened 2023-01-03
UIEH-1219: SPIKE: Difficulty returning to the eholdings page after adding a new Agreement

BogdanDenis:
## Description
Handle redirects in Eholdings when switching between different apps

## Approach
Create `RouteHistoryContext` context that keeps track of visited pages and saves it to sessions storage to persist between app switches
Need to keep track of pages manually because `react-router` doesn't provide history
This context provides function `navigateBack` that will find last eHoldings page that was visited and use `history.go` to skip all non-eholdings pages

`useHistoryBack` created that will either use `history.goBack` if we're navigating eHoldings, or `navigateBack` from `RouteHistoryContext` if we're going from another app

## Screenshots

https://user-images.githubusercontent.com/19309423/210358429-0f47ef0f-98ef-4302-922e-6beb20fcec52.mp4

## Issues
[UIEH-1219](https://issues.folio.org/browse/UIEH-1219)


src/components/route-history/route-history.js
```diff
@@ -0,0 +1,74 @@
...
+  }, []);
+
+  const navigateBack = useCallback(() => {
+    const nonEholdingsPagesCount = routeHistory.current.findIndex((page) => page.pathname.includes('eholdings') && page.pathname !== location.pathname);
+
+    if (nonEholdingsPagesCount > 0) {
+      // go(-1) will return to previous page
+      // if previous page was non eholdings - need to call go(-2)
+      // and this extends to N non-eholdings pages - call go(-(N + 1))
+      history.go(-(nonEholdingsPagesCount + 1));
```
Dmytro-Melnyshyn, line 53: If we do many navigations on the non eholdings page, then the navigation process will be unexpected on the eholdings page. 

I think we need to take into account the number of navigations on a non eholdings page too. It might be fixed by passing `history.length` via `LastVisitedContext` that is used in [main navigation](https://github.com/folio-org/stripes-core/blob/bbdf2dd352fbbaeeb91ab44c13351fa5b55ca5b4/src/components/MainNav/MainNav.js#L200).


https://user-images.githubusercontent.com/77053927/210651360-4a676aad-4905-4f2a-a597-16f557b7f95d.mp4

Dmytro-Melnyshyn:
Navigation fires only after the second click

https://user-images.githubusercontent.com/77053927/210653475-c625f739-a68a-424a-9edd-a56d2adbefb5.mp4

BogdanDenis:
Nice cases@Dmytro-Melnyshyn, I think I got them all to work

https://user-images.githubusercontent.com/19309423/210764203-08700e34-27de-4236-91c6-6494f578ac29.mp4

Sometime closing a package returns to search page without parameters, but that's ok according to requirements in the user story

Dmytro-Melnyshyn (approved):

PR #1627, opened 2023-01-04
UIEH-1344: Error while accessing undefined properties in eHolidngs Ti…

ZazaManagadze:
…tles list


BogdanDenis (changes requested):
Looks good, please also add entry to CHANGELOG.md file to the current version

BogdanDenis (approved):

PR #1642, opened 2023-03-09
UIEH-1355: Revise the 'Settings (eHoldings): View settings' permission.

Dmytro-Melnyshyn:
## Purpose
Revise the `Settings (eHoldings): View settings` permission

## Issues
[UIEH-1355](https://issues.folio.org/browse/UIEH-1355)
## Screenshot
image


src/components/settings/settings.js
```diff
@@ -129,27 +114,29 @@ class Settings extends Component {
           activeLink={pathname}
         >
           <div className={css.listSectionContent}>
-            <IfPermission perm="ui-eholdings.settings.root-proxy">
+            <IfPermission perm="kb-ebsco.kb-credentials.root-proxy.get">
```
BogdanDenis, line 117: Why do we need to use back-end permissions here?

Dmytro-Melnyshyn: Otherwise users with only `Settings (eHoldings): View settings permission` will not be able to view the root proxy.

BogdanDenis: But `ui-eholdings.settings.root-proxy` already has `kb-ebsco.kb-credentials.root-proxy.get` as a sub-permission, so it should work the same

Dmytro-Melnyshyn: According to the requirements, users should see the root proxy if they only have `Settings (eHoldings): View settings`. With only `ui-eholdings.settings.root-proxy` they will not be able to.

BogdanDenis: In this case just checking for `Settings (eHoldings): View settings` should work because it already has `kb-ebsco.kb-credentials.root-proxy.get` as a sub-permission
I just prefer to use UI permission because we can limit it to give access to one specific part of application. But a back-end permission can be a part of any UI permission.
In case of proxy, if `kb-ebsco.kb-credentials.root-proxy.get` is a part of other permission that is not even related to settings and a user has it - user also can have access to settings page

Dmytro-Melnyshyn: If we set UI permission `Settings (eHoldings): View settings`, a user with **only** `ui-eholdings.settings.root-proxy` permission will not be able to see this button, which is odd because the `ui-eholdings.settings.root-proxy` permission is meant to have full access to root proxy.

BogdanDenis (approved):

PR #1656, opened 2023-04-14
UIEH-1350: Add a Packages facet to Titles tab

Dmytro-Melnyshyn:
## Purpose
Add a Packages facet to Titles tab

## Issues

[UIEH-1350](https://issues.folio.org/browse/UIEH-1350)

## Screencast

https://user-images.githubusercontent.com/77053927/232053392-86209845-e838-40ae-8933-4fc6151145e7.mp4


src/components/search-form/components/packages-filter-accordion/packages-filter-accordion.js
```diff
@@ -0,0 +1,83 @@
...
+  packagesList.sort();
+
+  const itemToString = option => option?.label || '';
+
+  return (
+    <div
+      role="tab"
+      data-testid="packagesFilter"
+    >
+      <FormattedMessage id="ui-eholdings.packages.filter">
```
BogdanDenis, line 46: It's better to just use `intl.formatMessage` instead of this syntax

src/components/search-form/components/packages-filter/packages-filter.test.js
```diff
@@ -0,0 +1,120 @@
...
+    packagesFilterMap={{}}
+    results={{}}
+    onUpdate={() => {}}
+    {...props}
+  />
+);
+
+const renderComponent = (props = {}) => render(getComponent(props));
+
+describe('Given PackagesFilter', () => {
```
BogdanDenis, line 41: Please add some tests to check that `onUpdate` is called with correct data when selecting options

src/components/search-form/components/packages-filter-accordion/packages-filter-accordion.js
```diff
@@ -0,0 +1,83 @@
...
+          <Accordion
+            label={label}
+            id="packagesFilter"
+            separator={false}
+            closedByDefault
+            header={FilterAccordionHeader}
+            displayClearButton={!!packagesList.length}
+            onClearFilter={() => onUpdate({ packageIds: undefined })}
+          >
+            {isLoading && !disabled
```
BogdanDenis, line 57: Instead of showing a spinner when facets are loading - can we disable the filter?
We usually hide elements that can't be accessed at all, like when a permission is missing etc.
But in our situation user can access the filter if they perform a search so I think just disabling it will be a better experience

BogdanDenis (changes requested):

Dmytro-Melnyshyn:
Thanks @BogdanDenis, I've fixed the comments.

BogdanDenis (approved):

PR #1671, opened 2023-06-14
ERM-2641: Upgrade to Grails 5 (including Hibernate 5.6.x) for Poppy

EthanFreestone:
build: erm 6.0

Added okapi interface dependency on new erm interface 6.0

refs ERM-2641

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "okapiInterfaces": {
       "eholdings": "3.0",
       "tags": "1.0",
-      "erm": "1.0 2.0 3.0 4.0 5.0"
+      "erm": "1.0 2.0 3.0 4.0 5.0 6.0"
     },
     "permissionSets": [
       {


BogdanDenis (approved):

zburke:
@EthanFreestone, code changes must have per-project Jira tickets attached to them so folks doing Jira archaeology have a way to find out which features of a given app are associated with which releases. By only attaching an ERM ticket instead of a UIEH ticket, a PO browsing through all the UIEH tickets associated with Poppy will not find this change, i.e. cannot answer the question, "In what release did this project begin supporting `erm` interface `v6`?"

IOW, there MUST be a 1::1 correspondence between GitHub repo and Jira project in order for things to line up correctly. This mirrors the way ERM projects handle things with a 1::1 correspondence between GitHub repo and `Component` so it's not really different than what you are already doing; it's the same kind of relationship, just listed in a different place.

PR #1672, opened 2023-06-20
UIEH-1380 align test babel config with stripes-webpack

zburke:
Align the babel config with that in `@folio/stripes-webpack` so that all plugins referenced in the config are actually present in the build.

Refs [UIEH-1380](https://issues.folio.org/browse/UIEH-1380), [STRWEB-87](https://issues.folio.org/browse/STRWEB-87)


BogdanDenis (approved):
Thanks @zburke !

PR #1697, opened 2023-11-03
UIEH-1397: The list of packages is no longer cut off in the modal window.

Dmytro-Melnyshyn:
## Purpose
The list of packages is no longer cut off in the modal window.

## Issues
[UIEH-1397](https://issues.folio.org/browse/UIEH-1397)

## Screencast

https://github.com/folio-org/ui-eholdings/assets/77053927/7cf0cb67-2ef6-4292-8b6d-f6f72c72fe92

### DIFF
--- a/src/components/title/_fields/package-select/package-select-field.js
+++ b/src/components/title/_fields/package-select/package-select-field.js
@@ -11,8 +11,6 @@ import {
   Icon,
 } from '@folio/stripes/components';
 
-import css from './package-select-field.css';
-
 function validate(value) {
   return value ? undefined : <FormattedMessage id="ui-eholdings.validate.errors.packageSelect.required" />;
 }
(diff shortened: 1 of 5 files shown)


src/components/title/_fields/package-select/package-select-field.js
```diff
@@ -35,13 +33,13 @@ const PackageSelectField = ({
...
-    <div
-      data-test-eholdings-package-select-field
-      className={css.packageSelectFieldContainer}
-    >
+    <div data-test-eholdings-package-select-field>
       <Field
         name="packageId"
         component={Selection}
+        popper={{
+          portal: document.getElementById('ModuleContainer'),
```
BogdanDenis, line 41: I think the root cause was the `overflow` on the modal, and the fix works when portal renders to `OverlayContainer`. So I'd prefer to use `OverlayContainer` because it's usually the element where we render such dropdowns
Good catch with the overflow!

BogdanDenis (changes requested):

src/components/title/_fields/package-select/package-select-field.js
Dmytro-Melnyshyn: Scrolling doesn't work for me with `OverlayContainer`.

BogdanDenis (approved):

PR #1706, opened 2024-01-05
replace moment with dayjs

JohnC-80:
The moment project has been sunset for a bit now.
In efforts to move the FOLIO platform to a supported library with as easy a transition as possible. we intend to make the move to dayJS.
It's practically a drop-in replacement for moment, with a VERY similar API.

Its locale information is split from the main bundle and it's very modular/plugin-based.
`stripes-components` will re-export/provide the import for dayjs with numerous useful plugins included as well as utilities for dynamically loading dayjs locale data.

More information is available at https://day.js.org/en/

Related PR's:
* stripes-components: https://github.com/folio-org/stripes-components/pull/2192
* stripes-core: https://github.com/folio-org/stripes-core/pull/1384


BogdanDenis (approved):
Hey @JohnC-80, the PR looks good. I've added one commit to fix some import formatting

src/components/resource/_fields/validate-date-range.js
```diff
@@ -109,16 +107,16 @@ const validateWithinPackageRange = (resourceDateRange, packageDateRange) => {
...
-    const endCoverageDate = resourceDateRange.endCoverage ? moment.utc(resourceDateRange.endCoverage) : present;
+    const beginCoverageDate = dayjs.utc(resourceDateRange.beginCoverage);
+    const endCoverageDate = resourceDateRange.endCoverage ? dayjs.utc(resourceDateRange.endCoverage) : present;
 
-    const packageBeginCoverageDate = moment.utc(packageBeginCoverage);
-    const packageEndCoverageDate = packageEndCoverage ? moment.utc(packageEndCoverage) : moment.utc();
-    const packageRange = moment.range(packageBeginCoverageDate, packageEndCoverageDate);
+    const packageBeginCoverageDate = dayjs.utc(packageBeginCoverage);
+    const packageEndCoverageDate = packageEndCoverage ? dayjs.utc(packageEndCoverage) : dayjs.utc();
+    const packageRange = new DayRange(packageBeginCoverageDate, packageEndCoverageDate);
```
BogdanDenis, line 119: Hey @JohnC-80 this is a very very belated review, apologies for that.
It looks like functions for `moment.contains` and `dayjs.isBetween` work differently.
`moment` returns `true` if we check dates that are either the beginning or end date, but `dayjs` will return false.
For example, for when checking if date `2021-01-01` is in range `2021-01-01` to `2021-01-31` `moment` will return true, but `dayjs` will return false.
The `isBetween` plugin has a parameter that defines if begin and end dates should or should not be included: https://day.js.org/docs/en/plugin/is-between

I think since we've been using `moment` and by default it includes start and end date then we should do the same with `dayjs`. Do you have any thoughts?

BogdanDenis: I am working on making this PR up-to-date so we can merge it, just need to confirm with you if we should make the date inclusivity a default behaviour or not

BogdanDenis: Created a PR to handle date ranges correctly https://github.com/folio-org/stripes-components/pull/2443

OleksandrHladchenko1 (approved):

PR #1736, opened 2024-10-07
UIEH-1431: Change permission `kb-ebsco.kb-credentials.item.get` to `kb-ebsco.kb-credentials.key.item.get` and `kb-ebsco.kb-credentials.uc.item.get` to `kb-ebsco.kb-credentials.uc.key.item.get`.

Dmytro-Melnyshyn:
## Description
BE changes some permissions in [mod-kb-ebsco-java](https://github.com/folio-org/mod-kb-ebsco-java/pull/548), so UI should also make the following changes:
1. `kb-ebsco.kb-credentials.item.get` to `kb-ebsco.kb-credentials.key.item.get`;
2. `kb-ebsco.kb-credentials.uc.item.get` to `kb-ebsco.kb-credentials.uc.key.item.get`.

Hey @zburke - should this permissions change be a breaking UI change if the BE side doesn't change the version?
It will be merged with https://github.com/folio-org/mod-kb-ebsco-java/pull/548.

## Issues
[UIEH-1431](https://folio-org.atlassian.net/browse/UIEH-1431)

### DIFF
--- a/.github/workflows/build-npm-release.yml
+++ b/.github/workflows/build-npm-release.yml
@@ -153,7 +153,7 @@ jobs:
           comment_title: Jest Unit Test Statistics
 
       - name: Publish Jest coverage report
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         if: always()
         with:
           name: jest-coverage-report
@@ -161,7 +161,7 @@ jobs:
           retention-days: 30
 
       - name: Publish yarn.lock
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         if: failure()
         with:
           name: yarn.lock
--- a/.github/workflows/build-npm.yml
+++ b/.github/workflows/build-npm.yml
@@ -96,7 +96,7 @@ jobs:
           comment_title: Jest Unit Test Statistics
 
       - name: Publish Jest coverage report
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         if: always()
         with:
           name: jest-coverage-report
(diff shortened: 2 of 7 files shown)


zburke:
>  if the BE side doesn't change the version

BE side must change the version. Otherwise they are in a bad spot: the claim of "our interface version has not changed" means existing implementations that are compatible with this interface-version will continue to work, but they will not because the permissions are no longer aligned. 

Once that's fixed, then, yes, this is a breaking change because after this PR, this module will no longer work with the previous version because (again) the permissions won't be aligned. How permissions relate to interface versions is super confusing and probably not well-documented/well-understood. It's something we should try to get @p01nd3xt3r's attention on so it can be noted in the developer docs. 

([Past conversations about permissions and API versions](https://github.com/folio-org/ui-users/pull/2067#issuecomment-1070788164), for reference.)

Dmytro-Melnyshyn:
> > if the BE side doesn't change the version
> 
> BE side must change the version. Otherwise they are in a bad spot: the claim of "our interface version has not changed" means existing implementations that are compatible with this interface-version will continue to work, but they will not because the permissions are no longer aligned.
> 
> Once that's fixed, then, yes, this is a breaking change because after this PR, this module will no longer work with the previous version because (again) the permissions won't be aligned. How permissions relate to interface versions is super confusing and probably not well-documented/well-understood. It's something we should try to get @p01nd3xt3r's attention on so it can be noted in the developer docs.
> 
> ([Past conversations about permissions and API versions](https://github.com/folio-org/ui-users/pull/2067#issuecomment-1070788164), for reference.)

I agree, I thought so too, thanks for confirming. I've aligned this with BE and made the necessary updates.

psmagin (approved):

UladzislauKutarkin (approved):

zburke (approved):

PR #1743, opened 2024-10-28
UIEH-1436: Rename permissions to match the naming convention.

Dmytro-Melnyshyn:
## Description
Rename permissions to match the naming [convention](https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/156368925/Permissions+naming+convention). Remove [permissions violations](https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/458752012/Permissions+violations) of this module.

Merge with https://github.com/folio-org/stripes-testing/pull/4401

## Issues
[UIEH-1436](https://folio-org.atlassian.net/browse/UIEH-1436)


package.json
```diff
@@ -222,7 +222,7 @@
         ]
       },
       {
-        "permissionName": "ui-eholdings.package-title.select-unselect",
+        "permissionName": "ui-eholdings.package-title.execute",
```
BogdanDenis, line 225: I think we should still have `select-unselect` because without it it's impossible to tell what we're executing
So, maybe call it `ui-eholdings.package-title.select-unselect.execute`

package.json
```diff
@@ -244,7 +244,7 @@
         ]
       },
       {
-        "permissionName": "ui-eholdings.titles-packages.create-delete",
+        "permissionName": "ui-eholdings.titles-packages.manage",
```
BogdanDenis, line 247: Maybe call it `title-packages.create-delete.execute` or `title-packages.create-delete.manage`?
Again, without `create-delete` it's impossible to tell what we are managing or executing.

mariia-aloshyna (approved):

Terala-Priyanka (approved):

BogdanDenis (approved):

PR #1757, opened 2025-01-09
UIEH-1440: Package Record - Title List Accordion - Improve Visibility of Search Within/Filter/Sort options.

Dmytro-Melnyshyn:
## Description
UI changes:
- replace the magnifying glass with a search box, search options box, and Actions button;
- remove accordion from all filters (they are shown in Actions dropdown);
- show a clear icon next to the filter label if the filter's option is selected;
- badge near Actions button shows hover over text;

Functionality changes:
- submit the search once the Enter keyboard key is hit, and once any filter option is selected;
- move focus to the filter label after hitting the clear icon or to the default option for filters with radio buttons.

Before:
image

image

After:
image

image

## Issues
[UIEH-1440](https://folio-org.atlassian.net/browse/UIEH-1440)

## Approach
- create a separate component `search-section` for a search box, search options, and Actions button;
- the UI version with modal uses accordion for the access type and tags filters, so extract filters without accordion to a separate component to reuse in both the modal and newly created search section. Modal is still in use in other places;
- add the ability to render filters with radio buttons with and without accordion. The `hasAccordion` prop is used.

Screencasts:

https://github.com/user-attachments/assets/fb806f93-24e4-4c4d-8247-37410f092690


src/components/search-form/search-filters.js
```diff
@@ -39,7 +44,64 @@ const SearchFilters = ({
...
+      [name]: undefined,
+    });
+  };
+
+  const handleClearButtonClick = (labelId, name) => {
+    handleClearFilter(name);
+
+    // focus on the default option
+    setTimeout(() => {
+      document.querySelector(`[aria-labelledby="${labelId}"] input[tabindex="0"]`)?.focus();
```
BogdanDenis, line 63: Can we use `autoFocus={isChecked}` on `<RadioButton>` instead of this call?

Dmytro-Melnyshyn: We need to move the focus after hitting the clear icon, this will happen on page update, and `autoFocus` is only for page mounting.

BogdanDenis (approved):

artem-blazhko (approved):

PR #1763, opened 2025-02-12
update stripes deps and react-intl for sunflower

BogdanDenis:
## Description
update stripes deps and react-intl for sunflower

## Issues
[UIEH-1454](https://folio-org.atlassian.net/browse/UIEH-1454)
[UIEH-1453](https://folio-org.atlassian.net/browse/UIEH-1453)


zburke (changes requested):
please update `version` in `package.json`

zburke (approved):

PR #1777, opened 2025-05-07
UIEH-1471 Show a loading indicator under Package Titles list when package titles are updating.

BogdanDenis:
## Description
Before switching to MCL Package Titles list showed a loading indicator when a Package was added/removed from holdings and package-titles records were being updated.
In this PR we're adding that functionality back to current implementation.

## Screenshots

https://github.com/user-attachments/assets/0a1b41a9-3c7a-4fc4-b5ba-799ea68623b4

## Issues
[UIEH-1471](https://folio-org.atlassian.net/browse/UIEH-1471)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [11.1.0] (IN PROGRESS)
 
 * Replace `moment` usage with `dayjs`. (UIEH-1407)
+* Show a loading indicator under Package Titles list when package titles are updating. (UIEH-1471)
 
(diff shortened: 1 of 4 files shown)


Dmitriy-Litvinenko (approved):

zburke (approved):
Functionally, this resolves the stated bug; good. But I don't understand why adding a separate condition and spinner is necessary. Can't we just do 
```
<MultiColumnList loading={isLoading || isTitlesUpdating} ... />
```

PR #1785, opened 2025-06-04
FOLIO-4308 make erm instance optional

NikitaSedyx:
## Purpose

agreements app is going to be split from platform-complete + erm is optional for eholdings functionality

## Approach

Move to optional

## Learning

Create a ticket to make code change to not fail when erm interface is not present


NikitaSedyx:
done in scope of https://github.com/folio-org/ui-eholdings/pull/1786

PR #1786, opened 2025-06-04
UIEH-1476 Make "erm" an optional interface in eHoldings.

BogdanDenis:
## Description
Agreements will become a separate application as part of Eureka platform and we need to make erm interface an optional dependency.

In scope of this ticket we need to hide functionality that’s related to Agreements when that interface is missing:
- Package Show page - Agreements accordion
- Package+Title Show page - Agreements accordion

## Issues
[UIEH-1476](https://folio-org.atlassian.net/browse/UIEH-1476)

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,9 @@
     ],
     "okapiInterfaces": {
       "eholdings": "4.0",
-      "tags": "1.0",
+      "tags": "1.0"
+    },
+    "optionalOkapiInterfaces": {
       "erm": "1.0 2.0 3.0 4.0 5.0 6.0 7.0"
     },
     "permissionSets": [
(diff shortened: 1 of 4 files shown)


NikitaSedyx (approved):

package.json
```diff
@@ -96,7 +96,9 @@
     ],
     "okapiInterfaces": {
       "eholdings": "4.0",
-      "tags": "1.0",
+      "tags": "1.0"
+    },
+    "optionalOkapiInterfaces": {
```
Dmytro-Melnyshyn, line 101: Changing required to optional is a breaking change.

NikitaSedyx: it shouldn't be breaking as it's still backward compatible

PR #1811, opened 2026-04-16
Release v11.1.0

BogdanDenis:


package.json
```diff
@@ -26,9 +26,9 @@
   "devDependencies": {
     "@folio/eslint-config-stripes": "^8.0.0",
     "@folio/jest-config-stripes": "^3.0.0",
-    "@folio/stripes": "^10.0.0",
+    "@folio/stripes": "^10.1.0",
     "@folio/stripes-cli": "^4.0.0",
-    "@folio/stripes-core": "^11.0.0",
+    "@folio/stripes-core": "^11.1.0",
```
zburke, line 31: I have high confidence the stripes-core import is cruft, a hold-over from when old BTOG tests directly imported its setup functions. At present, I don't see any such imports and tests run cleanly without it. Your call whether to remove this now or at a later date.

package.json
```diff
@@ -68,7 +68,7 @@
     "redux-observable": "^1.2.0"
   },
   "peerDependencies": {
-    "@folio/stripes": "^10.0.0",
+    "@folio/stripes": "^10.1.0",
```
zburke, line 71: Technically, raising the minimum version of a peer is a breaking change because you're stating a new external requirement. If you're confident you don't consume any new components or functions, leave this at ^10.0.0 and leave your package version at 11.1.0. If you're confident you do use new components or functions that are only present in 10.1.0, then this is the correct change and you should also bump your package version to 12.0.0. If you're not sure, I'm not gonna turn you in 🙈 .

zburke (approved):

package.json
BogdanDenis: Oh, this is good to know. We have done very little work in eholdings this release and haven't used any of the new stripes-core functionality, so I'll leave it as is, but will keep it in mind when releasing other modules

BogdanDenis: Also, the module is already released as `11.1.0` with `stripes v10.1.0`....this is just a PR to update these files in the master branch

PR #1815, opened 2026-05-28
UIEH-1493 Remove duplicate requests for KB list and Access Types when opening Settings.

BogdanDenis:
## Description
When opening eHoldings Settings we can see duplicate requests for:
- `/kb-credentials`
- `/access-types`

We can remove a call to `/kb-credentials` from `<SettingsRoute>` because it's also present in `<ApplicationRoute>` child component.
And there are two calls to `/access-types` in `<SettingsAccessStatusTypesRoute>` of which we can remove one.

## Issues
[UIEH-1493](https://folio-org.atlassian.net/browse/UIEH-1493)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [11.2.0] (IN PROGRESS)
 
 * Fix "Publication type" column is empty in "Titles" accordion of "Package" record. (UIEH-1487)
+* Remove duplicate requests for KB list and Access Types when opening Settings. (UIEH-1493)
 
 ## [11.1.0] (https://github.com/folio-org/ui-eholdings/tree/v11.1.0) (2026-04-16)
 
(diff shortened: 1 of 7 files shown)


zburke (changes requested):
Your PR description says "We can remove these requests from X because they are also present in Y". Alas, it appears we have tests for X that expect to see those requests. 

I think your work is correct, so we just need to update/remove the corresponding expectations from the tests.

BogdanDenis:
> Your PR description says "We can remove these requests from X because they are also present in Y". Alas, it appears we have tests for X that expect to see those requests.
> 
> I think your work is correct, so we just need to update/remove the corresponding expectations from the tests.

Yeah, missed the test - I removed that case where we expected a request

OleksandrHladchenko1 (approved):

zburke (approved):

PR #1824, opened 2026-07-08
UIEH-1452 Provider Record - Packages List view  - Change to MCL

BogdanDenis:
## Description
Change view of Packages list within a Provider view to MCL.

There is a large number of changed files in this PR, but most of them are files being deleted since they have code that is no longer needed (removing redux actions, reducers, epics, apis, imports and exports from index.js files).
The number of files where code was changed/added is actually < 10.

Refactored `<ProviderShowRoute>` component to be a functional component, so that we could use hooks and react-query to fetch required packages data. Fetching this data via react-query allowed us to delete a lot of redux boilerplate (~10 files!) and implement the same data fetching with just 50 lines of code.

New `<ProviderPackagesList>` component basically accepts provider's packages data and renders it in an MCL.
Unfortunately, we couldn't call the `useProviderPackages` hook in the component itself, because some data from that hook is also needed in `<ProviderShow>` component, so the data is fetched in `<ProviderShowRoute>`, which acts as a data-fetching and business logic container for the Provider Show page.

Also, added new Jest tests to verify the new components and hooks, and updated existing tests to work with these updates.

## Issues
[UIEH-1452](https://folio-org.atlassian.net/browse/UIEH-1452)

## Screenshots

https://github.com/user-attachments/assets/cd7fb6c1-7364-46e7-aa10-afa3610936d1


zburke (approved):
Maintenance? What is that?!? 🙌 ❤️

OleksandrHladchenko1 (approved):

PR #1828, opened 2026-07-23
UIEH-1502 refactor PackageShowRoute component

BogdanDenis:
## Description
Preparing Package Show-related components for new fields on Package Show page, which involves a lot of refactoring.
Again, turned a class component into a functional one, moved one some code into hooks, updated tests, verified everything works as expected.

One interesting finding: `<PackageShowRoute>`'s `model` prop was and is being mutated all over the place. Since it's been like that for almost 10 years, and the app works as expected - guess prop mutation isn't such a bad thing, eh? But we definitely need to use react-query, get rid of the model and the mutations.


zburke (approved):
Thanks for tackling these refactors in this very old application!

OleksandrHladchenko1 (approved):

PR #1838, opened 2026-08-13
UIEH-1501 refactor PackageShow route and view to use react-query to fetch packages

BogdanDenis:
## Description
Removed more of the old redux-based data fetching in favor of react-query.
Now packages in Package Show and Edit paged are fetched, deleted, and edited via react-query.

Main new hooks:
`use-provider` - fetches providers
`use-package` - fetches packages
`use-package-update` - updates packages
`use-package-delete` - deletes packages
`use-package-model` - this one joins all the other package hooks, exposes their functions (update, delete, create) and returns a `model`.
Before this change Packages CRUD was implemented via a `Model` class that used redux. These models have a very complex structure, and since they are prop drilled throughout eHoldings app - getting rid of this drilling would require much more effort and more changes in many components, so for now the shape of `model` prop is matched in returned `model` from `use-package-model`.
Also, Providers and Resources are still using this redux approach, so we need to keep compatibility with all record types. In future, when all record types are fetched via hooks - these records would just be strored in Context and retreived when needed


vashjs (approved):
Looks good to me. Please preffer object args for `useQuery()` as it will be only 1 option in future versions, so it will simplify migration. https://tanstack.com/query/v4/docs/eslint/prefer-query-object-syntax

OleksandrHladchenko1 (approved):

PR #1846, opened 2026-09-02
UIEH-1535 update custom alternate names field label

BogdanDenis:
## Description
update custom alternate names field label to say "Custom alternate names"

## Issues
[UIEH-1535](https://folio-org.atlassian.net/browse/UIEH-1535)


zburke (approved):
Or just change the value in `en.json` and then you only have to worry about one file instead of four 🤷 your call.

Dmytro-Melnyshyn (approved):

zepheiryan (approved):

BogdanDenis:
> Or just change the value in `en.json` and then you only have to worry about one file instead of four 🤷 your call.

Oh I'd rather update the label and code than worry whether the labels will be updated for the next build.
Lately translation updates have been a bit unpredictable like that

PR #1849, opened 2026-09-15
UIEH-1507: Package Search Results list display

zepheiryan:
## Purpose

Update package search results display to include package display name if set and more details about when the package is hidden / excluded.

## Approach

- Update shape of item to include `customDisplayName` and `visibility` array instead of `visibilityData`, include `customDisplayName` is result title
- Add a dedicated `PackageHiddenLabel` component to handle different combinations of visibility categories while leaving the original `HiddenLabel` for use with Title search result display
- LLM used for unit tests and reviewing

## Refs

https://folio-org.atlassian.net/browse/UIEH-1507

## Screenshots

Screenshot 2026-09-15 at 14 59 20


OleksandrHladchenko1 (approved):

src/components/package-hidden-label/package-hidden-label.js
```diff
@@ -0,0 +1,64 @@
...
+
+PackageHiddenLabel.propTypes = {
+  id: PropTypes.string.isRequired,
+  visibility: PropTypes.arrayOf(PropTypes.shape({
+    category: PropTypes.string,
+    hidden: PropTypes.bool,
+  })),
+};
+
+export default PackageHiddenLabel;
```
BogdanDenis: Please use named exports instead of default to enforce consistent naming of components where they are imported

src/components/package-hidden-label/package-hidden-label.js
```diff
@@ -0,0 +1,64 @@
+import { FormattedMessage } from 'react-intl';
+
+import { Icon, Tooltip } from '@folio/stripes/components';
+import PropTypes from 'prop-types';
```
BogdanDenis: Please group imports together in this order:

1. Third party imports
empty line
2. Imports from stripes modules
empty line
3. Imports from current module

src/components/package-hidden-label/package-hidden-label.js
```diff
@@ -0,0 +1,64 @@
...
+
+const tooltipMessageIds = [null, 'ui-eholdings.hiddenSingleCategory', 'ui-eholdings.hiddenDualCategories', 'ui-eholdings.hiddenTripleCategories'];
+const categoryLabelMessageId = (category) => {
```
BogdanDenis: This function seems redundant as this string interpolation can be passed directly to `<FormattedMessage>`

src/components/package-hidden-label/package-hidden-label.js
```diff
@@ -0,0 +1,64 @@
...
+  id,
+  visibility
+}) => {
+  const hasHiddenCategories = visibility?.some(category => category.hidden);
+
+  if (!hasHiddenCategories) {
+    return null;
+  }
+
+  const labelId = `label-${id}`;
```
BogdanDenis, line 29: `<Tooltip>` component renders it's own elements with labels, we just need to use ids that it gives us in our markup like so:

```
<span
  ref={ref}
  aria-labelledby={ariaIds.text}
  aria-describedby={ariaIds.sub}
>

```

With this, we can remove an additional span around `<FormattedMessage>`

src/components/package-hidden-label/package-hidden-label.test.js
```diff
@@ -0,0 +1,70 @@
...
+  const ftfVisible = makeCategory(CATEGORIES.FTF, false);
+  const ftfHidden = makeCategory(CATEGORIES.FTF, true);
+  const marcVisible = makeCategory(CATEGORIES.MARC, false);
+  const marcHidden = makeCategory(CATEGORIES.MARC, true);
+
+  const allVisible = [pfVisible, ftfVisible, marcVisible];
+  const allHidden = [pfHidden, ftfHidden, marcHidden];
+
+  describe('basic rendering', () => {
+    it('does not render when an empty argument is passed in', () => {
```
BogdanDenis: Please update the test structure to follow this pattern:

```
describe('when condition`, () => {
  it('should ...', ...)
});
```

We can group up several checks in one group, and with even one check per condition such strucure makes it easier to understand what conditions are broken when reviewing failed tests

src/components/search-package-list-item/search-package-list-item.js
```diff
@@ -58,6 +58,9 @@ const SearchPackageListItem = ({
           tag={headingLevel || 'h3'}
         >
           {packageName || item.name}
+          {item.customDisplayName &&
+            <em> ({item.customDisplayName})</em>
```
BogdanDenis: One heading space here

BogdanDenis (changes requested):

src/components/package-hidden-label/package-hidden-label.js
zepheiryan: This `Tooltip` doesn't use the `sub` prop, so it seemed like the main `aria-labelledby` should be the `FormattedMessage` while the `aria-describedby` would come from the `Tooltip` `text` prop. The top `span` wrapper is necessary because the `Icon` has a `presentation` role, which implies any aria properties are ignored. I don't think anything should be altered here.

BogdanDenis (approved):

